PR #9, opened 2020-01-30
meta tags, images, and home page updates

drewbanin:
- cleans up the repo (removes some unused files)
- adds meta tags/images
- adds google tag manager

Docusaurus automatically uses the first line in a markdown file as the `meta:description` for each page, so I shuffled some things around in the guide to make actual content come before random react code (otherwise this happens:)

Screen Shot 2020-01-30 at 3 17 41 PM

The home page is still kind of weak but.... better than what we had before IMO.


clrcrl:
The dbt logo now has the black logo. Was that intentional?

drewbanin:
Ooof - I was trying to set the meta image that would render in social links. I'll revert this

PR #28, opened 2020-02-28
Feature/port over docs

drewbanin:
@clrcrl requesting review. Lots of changes, but the big idea:
 - fixup some components
 - fix all the links
    - the site will now error at build-time if internal links point to a `slug` that doesn't exist
    - in dev, invalid links are underlined in red
- added a [component reference](https://5e66683817f01100081755b5--docs-getdbt-com.netlify.com/styles/)
- fixed all the headers to make in-page nav sensible
- added redirects to map the existing readme.io url scheme onto the new docs
- download images into this repo from readme.io
- fix some markdown... viewed every page in this repo at least twice :)

TODO (in a future PR):
 - redirects for older versions?
 - what do we do about versioning these docs??


drewbanin:
@clrcrl requesting review. Lots of changes, but the big idea:
 - fixup some components
 - fix all the links
    - the site will now error at build-time if internal links point to a `slug` that doesn't exist
    - in dev, invalid links are underlined in red
- added a [component reference](https://5e66683817f01100081755b5--docs-getdbt-com.netlify.com/styles/)
- fixed all the headers to make in-page nav sensible
- added redirects to map the existing readme.io url scheme onto the new docs

TODO (in a future PR):
 - redirects for older versions?
 - what do we do about versioning these docs??

PR #34, opened 2020-03-13
Redo seeds documentation

clrcrl:
## Description & motivation
* Redo seeds docs to leverage FAQ structure + a new reference section
* Include placeholder docs for other reference sections to be written

## To-do before merge
- [x] Fact checking, esp. around hooks and quote_columns advice
- [x] Proof reading — I need to step away and come back to it to do this :) 
- [ ] Get the hyperlinked yaml working in the `dbt_project.yml` docs
- [ ] Component-ify the configs sections so they render on the page
- [ ] Ensure each config will show up as a sensible search result
- [x] Ensure we don't break existing redirects, esp. around the `reference` path

## Notes
- There's still a few tiny inconsistencies in the docs — some say "usage", others just have "examples". I'm not too fussed right now as I think we'll figure out what we want to do as we build more of these reference pages.

### DIFF
--- a/website/docs/docs/building-a-dbt-project/seeds.md
+++ b/website/docs/docs/building-a-dbt-project/seeds.md
@@ -3,22 +3,30 @@ title: "Seeds"
 id: "seeds"
 ---
 
-## Overview
-`dbt seed` loads data from csv files into your data warehouse. Because these csv files are located in your dbt repository, they are version controlled and code reviewable. Thus, `dbt seed` is appropriate for loading static data which changes infrequently.
+## Getting started
+Seeds are CSV files in your dbt project (typically in your `data` directory),
+that dbt can load into your data warehouse using the `dbt seed` command.
 
-The `dbt seed` command will load `csv` files located in the `data-paths` directory of your dbt project into your data warehouse. You can configure the `data-paths` directory by adding the following line to your `dbt_project.yml` file:
+Seeds can be referenced in downstream models the same way as referencing models
+— by using the `ref` [function](docs/writing-code-in-dbt/jinja-context/ref.md).
 
-<File name='dbt_project.yml'>
+Because these CSV files are located in your dbt repository, they are version
+controlled and code reviewable. Seeds are best suited to static data which
+changes infrequently.
 
-```yaml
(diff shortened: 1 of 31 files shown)


clrcrl:
@drewbanin -- heads up that this build is failing locally because of the `reference` redirects :/

drewbanin:
@clrcrl when you say "locally" - do you mean on your computer? These redirects shouldn't have any effect on a local build!

clrcrl:
@drewbanin -- I had another thought (which might have been one of your earlier thoughts). As well as hyperlinked yaml, it would be cool if these hyperlinks had pop-ups when you hover which displays the front matter. It might require a revision to some of the frontmatter/structure of the markdown files though. A little thing like:
```
data-paths: [directorypath]
___/\__________________________________________
| definition: Optionally specify a custom list |
| of directories where seed files are located. |
| type: [directorypath]                        |
| default: [data]                              |
| example: `data-paths: [data]`                |
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```

Also, you  may have noticed that I've been defining different data types here, and I've been trying to use them pretty consistently. It might be worth writing them down somewhere.

drewbanin:
@clrcrl i totally buy this and we can totally do a popover! As long as the front-matter is consistent and comprehensive, we should be able to do this relatively soon. Nice work on that ascii art haha

I <3 the idea of having good and well-defined types. There are other situations where we have funny "types" like:
 - either a string, or a list of strings
 - either `null` or a string
 - a well-defined dictionary, like the partitioning config for BigQuery tables in v0.16.0. Some fields are required and other fields are optional

I think enumerating the types that we use in a version controlled text file is a good idea - we'll thank ourselves for it later :)

clrcrl:
What if (and stick with me here), we put all the configs in a yaml file, and then for things where we need long form markdown, we link do out to a markdown file with Jinja. Like some sort of call like 
```yml
- name: my_config
   examples: "{{ docs('my_config_examples' }}"
```
🤯

PR #37, opened 2020-03-17
add missing docs, version redirects

drewbanin:
## Description & motivation
Migrated over some new docs, added redirects. I think after merging this, we can cut over the docs!


drewbanin:
@clrcrl fyi, merging this!

PR #45, opened 2020-03-18
Update snowflake-configs.md

djayatillake:

### DIFF
--- a/website/docs/docs/building-a-dbt-project/building-models/snowflake-configs.md
+++ b/website/docs/docs/building-a-dbt-project/building-models/snowflake-configs.md
@@ -139,6 +139,26 @@ create or replace table my_database.my_schema.my_table as (
  alter table my_database.my_schema.my_table cluster by (session_start);"
 ```
 
+It is possible to use functions of fields to cluster by but this must be enclosed in square brackets as below, even if only for a function on one field:
+
+```sql
+
+{{
+  config(
+    materialized='table',
+    cluster_by=['to_date(session_start)', 'left(session_id,1)']
+  )
+}}
+
+select
+  session_id,
+  min(event_time) as session_start,
+  max(event_time) as session_end,
+  count(*) as count_pageviews
+
+from {{ source('snowplow', 'event') }}
+group by 1
+```
 ### Automatic clustering
 
 Automatic clustering is a preview feature in Snowflake (at the time of this writing) and as such, some accounts may have it turned on while others may not. You can use the `automatic_clustering` config to control whether or not automatic clustering is enabled for dbt models. When `automatic_clustering` is set to `true`, dbt will run an `alter table <table name> resume recluster` query after building the target table. This configuration is only required for Snowflake accounts which do not have automatic clustering enabled. For more information, consult the [Snowflake documentation on Manual Reclustering](https://docs.snowflake.net/manuals/user-guide/tables-clustering-manual.html#switching-from-manual-reclustering-to-automatic-clustering).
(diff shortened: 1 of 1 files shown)


clrcrl:
@drewbanin — assigning you to this one as I don't know this part of the codebase that well

drewbanin:
hey @djayatillake - thanks for opening this PR! I did some more testing here to confirm -- I don't think this updated documentation is necessarily correct, but there's definitely some valuable information in here.

Here's the test I ran:

```
-- models/clustered.sql
{{ config(materialized='table', cluster_by="left(uuid, 3)") }}

select 'abc123' as uuid
```

When I run
```
$ dbt run -m clustered
```

I see the following in the logs:
```
create or replace transient table "ANALYTICS".dbt_dbanin.cluster  as (
  select * from(
    select 'abc123' as uuid
   ) order by (left(uuid, 3))
);
....
alter table "ANALYTICS".dbt_dbanin.cluster cluster by (left(uuid, 3));
````

So, even without using the `list` syntax to specify a `cluster_by` config, I _am_ seeing dbt add the clustering keys as intended. Can you advise on the type of error message you saw when you tried something like this in the past?

-------

I really like the part of this PR where you call out that expressions can be used as clustering keys. I'll drop some suggestions inline in the PR here.

File: website/docs/docs/building-a-dbt-project/building-models/snowflake-configs.md [hunk 1]
```diff
@@ -139,6 +139,26 @@ create or replace table my_database.my_schema.my_table as (
  alter table my_database.my_schema.my_table cluster by (session_start);"
 ```
 
+It is possible to use functions of fields to cluster by but this must be enclosed in square brackets as below, even if only for a function on one field:
```
drewbanin: Maybe we could change this to something like:

> It is possible to use expressions in the `cluster_by` config as shown below

djayatillake:
> hey @djayatillake - thanks for opening this PR! I did some more testing here to confirm -- I don't think this updated documentation is necessarily correct, but there's definitely some valuable information in here.
> 
> Here's the test I ran:
> 
> ```
> -- models/clustered.sql
> {{ config(materialized='table', cluster_by="left(uuid, 3)") }}
> 
> select 'abc123' as uuid
> ```
> 
> When I run
> 
> ```
> $ dbt run -m clustered
> ```
> 
> I see the following in the logs:
> 
> ```
> create or replace transient table "ANALYTICS".dbt_dbanin.cluster  as (
>   select * from(
>     select 'abc123' as uuid
>    ) order by (left(uuid, 3))
> );
> ....
> alter table "ANALYTICS".dbt_dbanin.cluster cluster by (left(uuid, 3));
> ```
> 
> So, even without using the `list` syntax to specify a `cluster_by` config, I _am_ seeing dbt add the clustering keys as intended. Can you advise on the type of error message you saw when you tried something like this in the past?
> 

This is strange, I can't reproduce the error again.

website/docs/docs/building-a-dbt-project/building-models/snowflake-configs.md
```diff
@@ -139,6 +139,26 @@ create or replace table my_database.my_schema.my_table as (
  alter table my_database.my_schema.my_table cluster by (session_start);"
 ```
 
+It is possible to use functions of fields to cluster by, as below:
```
drewbanin: I think saying "expressions" might be clearer than "functions of fields" here - you buy that?

```suggestion
It is possible to cluster by expressions, as below:
```

drewbanin (commented):
One last suggestion and then i think we'll be good to merge this :)

VersusFacit:
@drewbanin Interested in having me "rescue" this PR. I'd have to do a bit of command line work to be able to see this through, but I am working through the repo to get us up to date.

Alternatively, if @djayatillake you're still with us, I'd love for you to do the honors of wrapping this up :)

drewbanin:
blast from the past! Sure thing - if @djayatillake is down to revisit, would love to get this merged!

djayatillake:
Sure let me take a look

djayatillake:
I think it looks good now but there are conflicts to resolve

PR #48, opened 2020-03-19
Update exceptions example

clrcrl:
Noticed this today when I went to use it for the first time. If you use `{{ exceptions.warn("hello") }}` it gets resolved to `None` which can mess with your SQL.

@drewbanin -- Would you consider this a dbt bug? I expected this to work similarly to `{{ log() }}`, but my guess is that `log` was hacked to not return `None`.

### DIFF
--- a/website/docs/docs/writing-code-in-dbt/jinja-context/exceptions.md
+++ b/website/docs/docs/writing-code-in-dbt/jinja-context/exceptions.md
@@ -31,7 +31,7 @@ __Example usage__:
 
 ```sql
 {% if number < 0 or number > 100 %}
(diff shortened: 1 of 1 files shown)


drewbanin:
@clrcrl I think using `{% do ... %}` for things like this is probably always a good idea, including for things like `config()`, but that's definitely not a best practice today:
```
{% do config(...) %}
```

So we've instead made methods that are called in the context like `config` and `var` return an empty string. We could totally do this for exceptions too, but in general, i'd advise document thing `do` syntax for these methods to make sure folks understand the difference between `{{ ` and `{% do ...`

PR #64, opened 2020-03-26
Remove readme syntax

clrcrl:


clrcrl:
@drewbanin -- just as an FYI

PR #72, opened 2020-03-31
Remove trailing quote from oauth command

ryantuck:
## Description & motivation

Copying the `gcloud` oauth command didn't work as the command had a trailing quote `"`. 

Removing it allows the command to work as displayed.

## To-do before merge

¯\\_(ツ)_/¯


clrcrl:
@drewbanin -- we've had a couple of these come in. I tried to do a regex search through the repo for <code>"\n```</code>, but nothing is coming back (idk if I just don't know regex), can you also take a squiz?

And thanks @ryantuck!

PR #73, opened 2020-04-01
remove stray quotes

drewbanin:
## Description & motivation
Opening via this comment: https://github.com/fishtown-analytics/docs.getdbt.com/pull/72#issuecomment-606965711

FWIW you can't use `grep` to find these things because grep only matches a single line at a time. I use [the silver searcher](https://github.com/ggreer/the_silver_searcher) and found these instances with:

```
ag -l '"\n```'
```

I also removed the converter scripts since we don't really need them anymore! The relevant and extraneous quote char was [here](https://github.com/fishtown-analytics/docs.getdbt.com/blob/47c6e1de6d1896de7aec07b1dbb34e1c87b67f97/website/converter/fix_blocks.py#L96) -- looks like it only impacted unnamed code blocks, which i guess we didn't have too many of!


clrcrl:
Merging this now as the sooner it gets in, the better!

PR #76, opened 2020-04-02
Add migration guides for old dbt versions

jtcohen6:
## Description & motivation

> We’re going to push to remove old/deprecated versions of dbt from Cloud. We want to remove everything < 0.13 and help users upgrade if they’re that far behind.
>
> The only thing for us to do is update the migration guides in the dbt docs to include an upgrading guide for these very old versions of dbt.

## To-do before merge

This is my first time editing the docs since they moved to Docusaurus, so I want to be sure that:
- [ ] All relative links are working


clrcrl:
You have to add these to `sidebars.json` to get them to render! The links _should_ give us back a bad Netlify build, but I'm not confident, so I'll manually check them :)

PR #79, opened 2020-04-04
links in markdown

drewbanin:
## Description & motivation
Add support for markdown links in markdown code blocks

## To-do before merge
- Verify that we're not accidentally breaking existing markdown code blocks

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -23,6 +23,7 @@ module.exports = {
 
     prism: {
       theme: require('prism-react-renderer/themes/nightOwl'),
+      additionalLanguages: ['bash', 'shell'],
     },
     navbar: {
       logo: {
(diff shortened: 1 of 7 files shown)


clrcrl:
Played around with this [here](https://www.loom.com/share/534d9a32228c47d0a9a553ee67f07f64)

Overall feedback
- This is amazing
- Would love an update to the `styles` page to show how to use this
- Noticed that it doesn't work for filepath links (e.g. `project-configs/data-paths.md`), only for slug links (`project-configs/data-paths`)
    - Filepath links are a Docusaurus feature. The benefit of them is that it makes links work when viewing in GitHub (marginal benefit)
    - I've been using them when writing new docs, but the migration script used the slug links.
    - The styles page shows examples of slug links right now
    - I'm happy to live without them here, but would like to be consistent

drewbanin:
Ok - I just pushed a change to update the `/styles` page with an example. I investigated using a relative path, but that's going to require some more work here. Are you alright with shipping this as-is and then revisiting to add support for relative links in the future? Looks super doable

PR #81, opened 2020-04-04
Extra detail on incremental models

alanmcruickshank:
## Description & motivation
I've been delving back into incremental models recently and this would have been really useful info to help me to that. I think having a link to the old discourse article also provides a route for more advanced users to find out more.


clrcrl:
Thanks for the PR @alanmcruickshank!

I'm going to take the time to think about this one. I'm not sure whether we'd just prefer to link out to the Discourse article, and whether we need to generally restructure this article.

I'm a little stretched this week so I probably won't look at it again until a little later!

alanmcruickshank:
@clrcrl no problemo. I realise that this is the slightly larger contribution, so feel free to take as much or as little as you like. If you want me to rework anything myself then drop me a line :)

Issue #63, opened 2020-03-26
Feature: docs home page


clrcrl:
Random musings:
The docs.getdbt.com gives you a very clean landing page (but with little info), and if you click the "docs" link you get pushed straight into the Introduction.

I kind of like how in the [Django docs](https://docs.djangoproject.com/en/3.0/) there's a page which is like "here's where everything is". It makes it easier to get oriented quickly.

I'm not sure if this should live on the docs.getdbt.com landing page, or on docs.getdbt.com/docs

Anyway, just a thought 🤔

PR #90, opened 2020-04-13
WIP: Add markdown for docs homepage

clrcrl:
## Description & motivation
Update the home page

## To-do before merge
- [ ] Get this markdown rendering on the home page

### DIFF
--- a/website/docs/homepage.md
+++ b/website/docs/homepage.md
@@ -1,0 +1,31 @@
+# dbt documentation
+
+## Get started
+
+If you’re new to dbt, start here. These resources will get you off to a strong start:
+
+- **[What is dbt?](docs/introduction)**: Some of the very first questions a new user has are covered in this introduction.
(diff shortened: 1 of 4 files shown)


drewbanin:
@clrcrl what do you reckon about something like this?

Screen Shot 2020-04-13 at 6 59 36 PM

clrcrl:
OK, you were right — I think this might be a little too plain (I miss the moving arrows)

But I don't want you to sink any more time into it. I think this gives me enough to write a good brief for Beth though!

drewbanin:
Ok! We can actually keep the graph-viz-looking thing in there really easily. I removed it because it didn't look _great_ when it overlapped with the text.

I agree with your assessment though -- let's get Beth in here when we can -- I'm sure she'll do a much better job than I can :)

clrcrl:
Closes #63

clrcrl:
Closing in favor of #119

PR #98, opened 2020-04-15
Fix/broken relative links

drewbanin:
## Description & motivation

TODO : Change base branch to `feature/project-configs` after testing

Summary of changes:
 - support "relative" links in Markdown links (both inside and outside of code blocks)
 - change example code block links to use `.md` file extenstions

Let's consider this a WIP / proof-of-concept for the time being.

How linking works now:
 - Markdown links (or links created manually with the `<Link>` component will rewrite provided links to return "permalinks" that should work regardless of the path that the link is being clicked from
 - These aren't proper "relative" links, though I imagine we could support those in the future if we wanted to
 - Now, you can link to a document by either using:
  - The slug: `[my link](doc-slug)`
  - The trailing paths parts of the url: `[my link](some-path/doc-slug.md)`

When using the "trailing path-parts", the link will render to the first document which ends with with specified with the given path. This has two benefits:
1. We don't need to think super hard about relative paths if we move files around
2. The link will render nicely in GitHub's markdown viewer in most cases

This will _not_ work for relative links like `../some-path/other-path/../../the-path.md`. The link rewriting isn't going to know anything about the path that the link is being created from. I think that's ok?

### DIFF
--- a/website/docs/reference/dbt_project.yml.md
+++ b/website/docs/reference/dbt_project.yml.md
@@ -10,47 +10,46 @@ The following is a list of all available configurations in the `dbt_project.yml`
 <File name='dbt_project.yml'>
 
 ```yml
-[name](project-configs/name.md): string
+[name](project-configs/name): string
 
-[version](project-configs/version.md): version
+[version](project-configs/version): version
 
-[profile](project-configs/profile.md): profilename
+[profile](project-configs/profile): profilename
 
-[source-paths](project-configs/source-paths.md): [directorypath]
-[data-paths](project-configs/data-paths.md): [directorypath]
-[test-paths](project-configs/test-paths.md): [directorypath]
-[analysis-paths](project-configs/analysis-paths.md): [directorypath]
-[macro-paths](project-configs/macro-paths.md): [directorypath]
-[snapshot-paths](project-configs/snapshot-paths.md): [directorypath]
-[docs-paths](project-configs/docs-paths.md): [directorypath]
+[source-paths](project-configs/source-paths): [directorypath]
+[data-paths](project-configs/data-paths): [directorypath]
+[test-paths](project-configs/test-paths): [directorypath]
+[analysis-paths](project-configs/analysis-paths): [directorypath]
+[macro-paths](project-configs/macro-paths): [directorypath]
+[snapshot-paths](project-configs/snapshot-paths): [directorypath]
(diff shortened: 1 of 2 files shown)


drewbanin:
@clrcrl check this out, and feel free to change the base branch and merge into your branch if you'd like :)

clrcrl:
@drewbanin — I just rebased this and pushed it.

I ended up working around this by just using the slug references, and it seems to have worked. Do you still think we should merge this PR?

drewbanin:
@clrcrl it's up you you! I'm happy to close it and reopen in the future if we decide we want more flexibility around internal links. Also happy to merge as-is. Your call

clrcrl:
Let's merge it. It doesn't create any regressions (as far as I can tell) and it's a good thing to have up our sleeve!

website/docs/reference/dbt_project.yml.md
```diff
@@ -10,32 +10,31 @@ The following is a list of all available configurations in the `dbt_project.yml`
...
+[snapshot-paths](project-configs/snapshot-paths): [directorypath]
+[docs-paths](project-configs/docs-paths): [directorypath]
 
-[modules-path](modules-path): directorypath
-[target-path](target-path): directorypath
-[log-path](log-path): directorypath
-[modules-path](modules-path): directorypath
+[target-path](project-configs/target-path): directorypath
+[log-path](project-configs/log-path): directorypath
+[modules-path](project-configs/modules-path): directorypath
```
clrcrl, line 29: Lol this actually fixed a duplicate error that I had! Magic!

website/docs/reference/dbt_project.yml.md
```diff
@@ -10,32 +10,31 @@ The following is a list of all available configurations in the `dbt_project.yml`
...
 
-[clean-targets](clean-targets): [directorypath]
+[clean-targets](project-configs/clean-targets): [directorypath]
 
-[query-comment](query-comment): string
+[query-comment](project-configs/query-comment): string
 
-[require-dbt-version](require-dbt-version): version-range | [version-range]
+[require-dbt-version](project-configs/require-dbt-version): version-range
```
clrcrl, line 35:
```suggestion
[require-dbt-version](project-configs/require-dbt-version): version-range | [version-range]
```

clrcrl: Lost this in the rebase

PR #111, opened 2020-04-20
Snapshot configs

clrcrl:
## Description & motivation
Snapshots article — my goal was to get this to be a little shorter and more to the point.
- Replaced a lot of hand-waving with more concrete steps in an example.
- Moved some of the questions out to FAQs

Snapshots reference:
- Added the different configs

## To-do before merge
- [x] Finish reference docs
- [x] See if I can get the main article to be even shorter
- [x] Add articles to sidebar (blocked by open PRs)


clrcrl:
I'm now at the point where I don't know if I'm making the docs better or worse.

Send help.

drewbanin:
fyi these are incredible
Screen Shot 2020-04-27 at 11 03 40 PM

My only real feedback here is that I think it could be easier to navigate these docs if we tuck resource-specific configs into a resource-specific folder in the sidebar. In this case, that would involve moving `target_database`, `target_schema`, `check_cols`, `updated_at`, and `strategy` into a folder called `snapshot configs`, or similar. I think this will become really helpful when we add docs for other resource types, like models and tests later.

You buy that?

Overall these are really great, and I'm excited to merge them :)

PR #116, opened 2020-04-23
Fix on-run-end context docs

tjengel:
## Description & motivation

I wanted to use the grant_usage_to_schemas macro in my local project. After copying and pasting the code from the docs, I noticed there were a couple things missing that I needed to change in order for the macro to work. These changes include adding `schema` in front of the schema variable in the macro. I then put single quotes around the user in the grant block in the dbt_project.yml

I built the docs locally and I see my changes on the on run end context page.


clrcrl:
Thank you!

PR #128, opened 2020-04-30
Move schema.yml docs into reference section

clrcrl:
## Description & motivation
This PR moves the existing schema.yml docs into the reference section, and renames them "resource properties".

I'm proposing to merge these work-in-progress docs since I think they are already useful. Here's what I didn't do in this PR:
- Not all properties are documented (e.g. the source properties aren't fully documented) — this is because I wanted to document model properties, and ended up at least filling in most of the blanks for the other resource types.
- Some properties are not fully documented, e.g. `tests`, since they would require much more work
- I should add more examples, but again, would prefer to merge this now and make a separate PR for examples :)


clrcrl:
So, uh, we have a problem:

<pre>
### Use some markdown in a description

You can use markdown in your descriptions, but you may need to quote your description to ensure the YAML parser doesn't get confused by special characters!

<File name='models/schema.yml'>

```yml
version: 2

models:
  - name: dim_customers
    description: "**[Read more](www.google.com)**"

```
</pre>

I actually _want_ the `description:  "**[Read more](www.google.com)**"` to show as code:
```yml
version: 2

models:
  - name: dim_customers
    description: "**[Read more](www.google.com)**"

```

Because we made our code-blocks linky, it gets rendered as a hyperlink:

Screen Shot 2020-05-04 at 1 10 46 PM


How can I escape this?
I tried: 
```
    description: "**[Read more]\(www.google.com\)**"
```

But it rendered as:
Screen Shot 2020-05-04 at 1 10 08 PM

 _Or_, I've also been thinking that maybe we should make our linky-code-blocks a separate component? idk.

clrcrl:
Hey @aescay — I've assigned you as a reviewer.

My recent PRs are pretty big, because I'm changing the structure of our docs. The general idea of this restructure is:
1. write proper reference sections
2. condense articles in the "docs/" section into one-pagers, that introduce a concept
3. use FAQs as a way to link from the "docs/" section, into the reference section, rather than going into insane detail in the "docs/" section ([example](https://docs.getdbt.com/docs/building-a-dbt-project/using-sources/#overriding-quoting) — like, why would someone who has never heard of sources before need to know about overriding quoting straight away? That should be an FAQ)

☝️ Definitely happy to talk more about this, and how I think about documentation in general.

In this PR, I prepared the "schema.yml" reference docs (well, like 80% of them), but didn't so much of (2) and (3). I'm now working on the docs for "sources", so this structure might make more sense once we have that.

In terms of feedback, I'm more just looking for someone to glance through my work — mistakes in docs are very fixable, so I'm not too concerned about shipping mistakes. 

As you get more comfortable with docs.getdbt.com, I'm very excited to have someone that I can bounce ideas off!

aescay:
@clrcrl sorry for the late response on this! getting on to this now!

clrcrl:
Ooh I wasn't sure if I wanted to keep them, so left them out! The next PR deals with them a little better!

PR #130, opened 2020-05-01
Home page fixes

clrcrl:
## Description & motivation
- Add `h1` title, change other headings to `h2`
- Small copy changes for `Sentence case`

## Questions
- Does the `h1` look okay here?
- Should I revert the heading changes?

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -16,12 +16,12 @@ function Home() {
   const context = useDocusaurusContext();
 
   return (
-	  
+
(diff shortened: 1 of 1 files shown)


janessa-lantz:
I don't mind the h1. I think it makes the information architecture a little more clear even if it junks up the design a bit. So I'm think pro h1. 

Button text should go back to proper case

clrcrl:
> even if it junks up the design a bit
Should we get Beth to quickly add some style to this?

PR #134, opened 2020-05-05
Update modules with helpful reference.

mgenereu:
This reference would have helped me locate the correct datetime retrieval for my use case.


mgenereu:
Thank you for your help!

PR #136, opened 2020-05-05
Update getting-started-with-jinja.md

clrcrl:


clrcrl:
@aescay — FYI
[Here](https://deploy-preview-136--docs-getdbt-com.netlify.app/docs/writing-code-in-dbt/getting-started-with-jinja/#what-parts-of-jinja-are-dbt-specific) is the deploy preview.

This happened because we have some duplicate file names. We get a warning on our build about this, but we have sort of let it slide, since I am restructuring docs anyway.

The way to get around this is to provide some more of the URL path, and it will resolve correctly

PR #142, opened 2020-05-07
fix/update github setup with restart the app pod instructions

amychen1776:
This updates the on-prem setup with instructions to restart the pod.

### DIFF
--- a/website/docs/docs/dbt-cloud/on-premises/setup.md
+++ b/website/docs/docs/dbt-cloud/on-premises/setup.md
@@ -67,13 +67,13 @@ A new version of dbt Cloud will appear on the Version History page in your Confi
 
 You can apply or roll back these changes at any time by clicking the Deploy and Rollback buttons on the right side of this screen.
 
-### Github
+### Github Setup
 
 This setup is only required for Github App usage. 
 
-Log into the dbt Cloud configuration console, navigate to Settings, and check Enable under Github Integration.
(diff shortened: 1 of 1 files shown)


website/docs/docs/dbt-cloud/on-premises/setup.md
```diff
@@ -84,4 +84,16 @@ Enter the following:
...
+4. Restart the "app" pod to update the configuration values.
+##### To do this on kubernetes 1.15+:
+```
+kubectl rollout restart deployment app
+```
+##### To do this on kubernetes <1.15:
+```
+kubectl patch deployment app -p \
+  "{\"spec\":{\"template\":{\"metadata\":{\"annotations\":{\"date\":\"`date +'%s'`\"}}}}}"
+```
```
cmcarthur, line 97: this is strictly an improvement so I'm going to approve this. But, in the long run, it feels like this should be part of a "maintenance" section in the on-premises guide.

amychen1776: I agree. I also do think that Github setup (and maybe a guide to setting up a Slack app) should probably live in its own page. It feels a bit random on this one.

amychen1776: I think what needs to be done is more information on useful things to know about maintenance.

PR #144, opened 2020-05-08
Temp_fix of Github setup

amychen1776:
This will be the temporary fix for the disconnect between Github setup. We should probably create a more synced page about Github setup so it's all on one page.

### DIFF
--- a/website/docs/docs/dbt-cloud/on-premises/setup.md
+++ b/website/docs/docs/dbt-cloud/on-premises/setup.md
@@ -69,7 +69,7 @@ You can apply or roll back these changes at any time by clicking the Deploy and 
 
 ### Github Setup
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/dbt-cloud/on-premises/setup.md
```diff
@@ -69,7 +69,7 @@ You can apply or roll back these changes at any time by clicking the Deploy and
 
 ### Github Setup
 
-This setup is only required for Github App usage. 
+This setup is only required for Github App usage. Please reference the [Prerequisites](https://docs.getdbt.com/docs/dbt-cloud/on-premises/prerequisites/#github) for necessary information.
```
drewbanin, line 72: you should be able to use `on-premises/prerequisites#github` as the URL here. This will make the link work in local dev, and will turn on link validation :)

amychen1776: Ah epic, thank you!

PR #153, opened 2020-05-14
Update sample test logic

xarasco:
I think the logic in the test is accidentally flipped.

## Description & motivation
Reviewing testing documentation I noticed the test logic didn't seem right.  The test should pass if no rows are returned, therefore, I think `where pct_null < 0.05` should be changed to `where pct_null >= 0.05`.

## To-do before merge
make sure I'm not crazy


clrcrl:
@drewbanin — this LGTM but also I've not had enough coffee

PR #158, opened 2020-05-18
Feature/0.17.0 docs

drewbanin:
## Description & motivation
- new docs for v0.17.0

## To-do before merge
 - Review places where we render a v1 dbt_project.yml config
 - I need to do another pass / check for typos / etc
 - Double-check the changelog for any missed features that warrant a docs change
 - Link out to more helpful documents (release notes, discourse post, etc)


clrcrl:
You'll also need to update `using-variables.md`. I'm currently doing some work to restructure the model configs though, so I'd advise making as few changes as possible to avoid conflicts.

PR #159, opened 2020-05-19
Deprecate outdate freshness syntax

aescay:
## Description & motivation

Looks like the old syntax of `freshness: {}` does not work anymore and will still run the freshness test against the table that has an emptied out config. The `null` flag still works and does also look like the cleaner implementation. I opted to just drop the old syntax altogether.


clrcrl:
Assigning to @drewbanin since this was counter to my understanding, and he is closer to the product!

aescay:
No problem at all @drewbanin. I know things can get hectic and thanks!!! 🙏

PR #164, opened 2020-05-22
fix broken tutorial link

xarasco:
Getting Started Tutorial link in the FAQ was broken


clrcrl:
Thank you! That was quick, I only deployed that yesterday!

PR #168, opened 2020-05-26
Add example of snapshot ref

olhmr:
## Description & motivation

This adds an example of how to use `ref` to select from a snapshot in a downstream model.


clrcrl:
Thanks @olhmr!

PR #169, opened 2020-05-26
Add source example: bq sharded table

jtcohen6:
As seen in [this Slack thread](https://getdbt.slack.com/archives/C0VLZPLAE/p1590459213011300). I changed slightly vs. what I posted there after checking the compiled quoting.

### DIFF
--- a/website/docs/reference/resource-properties/identifier.md
+++ b/website/docs/reference/resource-properties/identifier.md
@@ -55,3 +55,37 @@ Will get compiled to:
 ```sql
 select * from postgres_backend_public_schema.orders
 ```
+
+### Reference sharded tables in BigQuery
+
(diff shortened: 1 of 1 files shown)


website/docs/reference/resource-properties/identifier.md
```diff
@@ -55,3 +55,37 @@ Will get compiled to:
 ```sql
 select * from postgres_backend_public_schema.orders
 ```
+
+### Reference sharded tables in BigQuery
```
clrcrl, line 59: Just for SEO purposes, I can imagine someone googling "sharded table source bigquery"
```suggestion
### Reference sharded tables as a source in BigQuery
```

jtcohen6: Good call!

PR #177, opened 2020-05-28
Add transient property to General Configs

xarasco:
I'm sure there's others that apply as well, however, the transient property seems important since: 
1. it's likely you would want to have recovery coverage in Snowflake
2. by default snapshots would be made as transient tables


clrcrl:
Hiya! Thanks for this PR! I actually have an open to-do to add "database-specific-configs" for models and snapshots — since `transient` is snowflake specific, we will address this as part of that. This is on me for not having a clear set of issues open — for now I'm going to close this issue

PR #179, opened 2020-05-28
update execute.md

stuartjash:
added payment_method_query instead of relation_query which is not referring to any local query variable


clrcrl:
Closing as this was merged in #180 
Thank you though!!!

PR #184, opened 2020-06-01
Re-write hooks and operations

clrcrl:
#### Group pre- and post-hooks. And on-run-start and -end hooks.:
The logic in these is pretty similar. No need for two separate articles

#### Make hooks and operations one article
I think hooks and operations serve a similar purpose — to provide a way to execute additional SQL. It makes more sense to me to put them into one article.
 
## To-do before merge
- [ ] I wrote these docs a few weeks back, before opening this PR, so it will be worth going through them to make sure I didn't leave any hanging sentences


website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -0,0 +1,127 @@
...
+id: "hooks-operations"
+---
+
+## Related documentation
+* [pre-hook & post-hook](pre-hook-post-hook)
+* [on-run-start & on-run-end](on-run-start-on-run-end)
+* [`run-operation` command](run-operation)
+
+## Assumed knowledge:
+* [Project configurations](dbt_project.yml.md)
```
drewbanin, line 12:
```suggestion
* [Project configurations](dbt_project.yml)
```

This is actually a broken link! I guess there's a change we should make in the link checker

PR #190, opened 2020-06-08
update command line interface

stuartjash:
added target to list of CLI commands


clrcrl:
Hello! `--target` is a command line option, rather than a separate command! We're working on these docs at the moment, so hopefully some of the newer edits will make things clearer

PR #202, opened 2020-06-11
Share another option for using dbt with Airflow

d-swift:
## Description & motivation

- We often see questions in the slack about using Airflow + dbt.. here's another method!

Plus @clrcrl suggested we add a reference :-)

## To-do before merge

I've checked the markdown renders correctly in vscode. 
- Do I need to check this renders correctly on the docs site?


clrcrl:
Thank you!

PR #203, opened 2020-06-11
Update using-operations.md

ramgopalan:
The given example of run-operation would result in error, as the argument should be given in ""
existing: $ dbt run-operation say_hi --args '{name: world}'
revised: $ dbt run-operation say_hi --args "{name: world}"


ramgopalan:
Hi @drewbanin  Yeah i am using a windows machine and this is the error I get
image

PR #219, opened 2020-06-18
Use a script to update dbt_project.yml to version 2 syntax

clrcrl:
This doesn't work, but is close to working. Feel like finishing it off @drewbanin?

Here's my current [regex test](https://regex101.com/r/AlZ8zs/1) — I can't figure out how to _just_ pluck out "schema".

Once this is done, we'll also likely need to go through and check if we should add `config-version: 2` to some of the full-length `dbt_project.yml` samples


clrcrl:
Closing in favor of #221

PR #230, opened 2020-06-22
change model description name

eogilvy12:
## Description & motivation

The model name is `payments` in the file description, but in the DAG and description below, the model is called `orders`. this change is for clarification

image


clrcrl:
Thanks Erin!

PR #251, opened 2020-07-02
Adding alternate view of Azure app registration

nave91:
## Description & motivation
Azure AD has multiple views depending on who is viewing the screen. This PR adds an alternate view and also fixes a permission error. 

https://fishtownanalytics.slack.com/archives/C015K0HD8BB/p1593635727020600


drewbanin:
hey @nave91 - thanks for fixing this! I think we can just replace the old image with the one you've included here. That might require a slight reworking of the contents of these sections - let me know if I can help out with it!

nave91:
@drewbanin If you are seeing that screen, there might be other people seeing that screen too. Maybe we can leave 2 images up there? I can re-word the image title to make that more clear if it isn't already.

drewbanin:
ok, fair enough. I do think the **Configuring the redirect URI** section might need to be updated though if anyone is seeing the _Alternate View_ screenshot. We'd probably want to collapse that section into the **Creating an application** section IMO

nave91:
@drewbanin That makes sense. I just collapsed redirect uri section into creating an application. Also updated docs to reflect alternative view. Let me know if you need me to change anything else or feel free to add commits and merge.

drewbanin:
made some updates based on our conversations - shipping this!

PR #259, opened 2020-07-06
WIP: reference overview

clrcrl:
I think something like this _could_ be useful, but at the moment I think it ends up being more confusing than less. So, idk. I'm opening it up as a draft to get thoughts on it.

### DIFF
--- a/website/docs/reference/overview.md
+++ b/website/docs/reference/overview.md
@@ -1,0 +1,22 @@
+---
+title: Reference section overview
+---
+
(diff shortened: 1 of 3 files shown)


drewbanin:
I like this concept, but I think the table leaves a little something to be desired. Maybe just replacing the table with markdown / an index could work here instead? What do you thin?

clrcrl:
What do you mean by an index?

drewbanin:
ah like a Table of Contents kind of thing

PR #264, opened 2020-07-07
Add article on debugging

clrcrl:
This is very overdue.

Was on my mind while answering Slack questions today — just like, how can we solve 90% of these super easily

### DIFF
--- a/website/docs/docs/guides/debugging-errors.md
+++ b/website/docs/docs/guides/debugging-errors.md
@@ -1,0 +1,364 @@
+---
+title: Debugging errors
+---
+## General process of debugging
+Learning how to debug is a skill, and one that will make you great at your role!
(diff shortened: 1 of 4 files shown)


website/docs/docs/guides/debugging-errors.md
```diff
@@ -0,0 +1,378 @@
+---
+title: Debugging errors
+---
+## General process of debugging
+Learning how to debug is a skill, and one that will make you great at your role!
+1. Read the error message — we try our best to write as much information as possible. It will normally contain the type of error, and the file where the error occurred. There's more details about different types of errors below.
```
drewbanin, line 6: should we reframe this from "we" to "dbt"?

clrcrl: Changed it slightly

PR #265, opened 2020-07-07
0.17.1: opt-in native env rendering

jtcohen6:
## Description & motivation

* Add reference docs for `is_bool`, `is_number`, `is_native`
* Update `is_text` and 0.17.0 migration guide with changelogs, callouts
* Reflects changes in https://github.com/fishtown-analytics/dbt/pull/2618

## To do before merge

Confirm that the pseudo-code I wrote as examples _actually_ works

**Update:** All the example code is working when I've checked out https://github.com/fishtown-analytics/dbt/pull/2618. One catch: the `freshness` example for `is_native` works _without_ the `is_native` filter. I'm guessing that the coercion-to-dict is already happening because of hologram. Can you think of a better example?

### DIFF
--- a/website/docs/reference/dbt-jinja-functions/as_bool.md
+++ b/website/docs/reference/dbt-jinja-functions/as_bool.md
@@ -7,12 +7,6 @@ The `as_number` Jinja filter will coerce Jinja-compiled output into a boolean
 value (`True` or `False`), or return an error if it cannot be represented
 as a bool.
 
-:::caution Heads up!
-In dbt v0.17.1, native rendering is not enabled by default. It is possible to
-natively render specific values using the [`as_bool`](as_bool), 
-[`as_number`](as_number), and [`as_native`](as_native) filters.
-:::
-
 ### Usage:
 
 In the example below, the `as_bool` filter is used to coerce a Jinja 
--- a/website/docs/reference/dbt-jinja-functions/as_bool.md
+++ b/website/docs/reference/dbt-jinja-functions/as_bool.md
@@ -4,7 +4,7 @@ id: "as_bool"
 ---
 
 The `as_number` Jinja filter will coerce Jinja-compiled output into a boolean
-value (`true` or `false`), or return an error if it cannot be represented
+value (`True` or `False`), or return an error if it cannot be represented
 as a bool.
 
 :::caution Heads up!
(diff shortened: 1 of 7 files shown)


clrcrl:
Can you also add these to `dbt-jinja-functions` ([page](https://docs.getdbt.com/reference/dbt-jinja-functions/)).
In an ideal world they would not need to be updated in two places, but, alas!

website/docs/reference/dbt-jinja-functions/as_bool.md
```diff
@@ -0,0 +1,37 @@
+---
+title: "as_bool"
+id: "as_bool"
+---
+
+The `as_number` Jinja filter will coerce Jinja-compiled output into a boolean
+value (`true` or `false`), or return an error if it cannot be represented
```
drewbanin, line 7:
```suggestion
value (`True` or `False`), or return an error if it cannot be represented
```

website/docs/reference/dbt-jinja-functions/as_text.md
```diff
@@ -3,26 +3,62 @@ title: "as_text"
...
+It is still possible to natively render specific values using the [`as_bool`](as_bool), 
+[`as_number`](as_number), and [`as_native`](as_native) filters. 
+
+:::
+
+### Usage
+
+In the example below, the `as_text` filter is used to assert that `''` is an
+empty string. In a native rendering, `''` would be coerced to the Python 
+keyword `None`. This specification is necessary in `v0.17.0`.
```
drewbanin, line 23:
```suggestion
keyword `None`. This specification is necessary in `v0.17.0`, but is not useful or necessary in later releases of dbt.
```

Or something like that...

jtcohen6:
@drewbanin Do you have any ideas for a better `is_native` example? A case where it's actually/uniquely needed

drewbanin:
I can't think of a good use case where we'd want to recommend `as_native` over `as_number` or `as_bool`. Maybe the one use case is around returning tuple/list/dict/set values from a jinja expression? I think that's pretty uncommon....

jtcohen6:
I couldn't come up with a sane example where `is_native` is strictly necessary, so I just removed the Usage section and added a note about recommending `as_bool` or `as_number` instead.

PR #267, opened 2020-07-08
Add note about html in dbt docs

tayloramurphy:
## Description & motivation

PR Adds:

* details about embedding HTML in descriptions
* how to use a hidden ref to force model execution order


clrcrl:
I think the place we really want to update this is the [`reference/resource-properties/description` page](https://github.com/fishtown-analytics/docs.getdbt.com/blob/master/website/docs/reference/resource-properties/description.md) ([rendered](https://docs.getdbt.com/reference/resource-properties/description/)).

Basically, the guides are more introductory concepts ("what is this thing? how do I use it?), while the advanced examples should live in the reference section
```

### Use html in a description

(and then an example!)

```

website/docs/reference/resource-properties/description.md
```diff
@@ -328,3 +328,26 @@ models:
 </File>
 
 If mixing images and text together, consider using a docs block instead.
+
+
+
```
tayloramurphy, line 333:
```suggestion
```

tayloramurphy:
@clrcrl submitting for your review. Not sure on the exact language though 🙂

website/docs/reference/dbt-jinja-functions/ref.md
```diff
@@ -36,8 +36,24 @@ The `{{ ref }}` function returns a `Relation` object that has the same `table`,
...
+
+### Two-argument variant
+
 There is also a two-argument variant of the `ref` function. With this variant, you can pass both a package name and model name to `ref` to avoid ambiguity. This functionality is not commonly required for typical dbt usage.
 
 ```sql
 select * from {{ ref('package_name', 'model_name') }}
 ```
+
+### Hidden ref
```
clrcrl, line 48: I'd call this "forcing dependencies"

website/docs/reference/dbt-jinja-functions/ref.md
```diff
@@ -36,8 +36,24 @@ The `{{ ref }}` function returns a `Relation` object that has the same `table`,
...
+
+In normal usage, dbt knows the proper order to run all models based on the usage of the `ref` function. There are cases though where dbt doesn't know when a model should be run. An example of this is when a model only references a macro. In that case, dbt thinks the model can run first because no explicit references are made at compilation time. To address this, a comment can be added in the file, after the configuration setting, to indicate which model it depends on:
```
clrcrl, line 50:
```suggestion
In normal usage, dbt knows the proper order to run all models based on the usage of the `ref` function. There are cases though where dbt doesn't know when a model should be run. An example of this is when a model only references a macro. In that case, dbt thinks the model can run first because no explicit references are made at compilation time. To address this, you can use a SQL comment along with the `ref` function — dbt will understand the dependency, and the compiled query will still be valid:
```

clrcrl (approved):
I made a couple of suggestions, but broadly I'm happy for this to go in as-is.

(Also sorry that I let this one slip!)

website/docs/reference/dbt-jinja-functions/ref.md
```diff
@@ -36,8 +36,23 @@ The `{{ ref }}` function returns a `Relation` object that has the same `table`,
...
+
+### Two-argument variant
+
 There is also a two-argument variant of the `ref` function. With this variant, you can pass both a package name and model name to `ref` to avoid ambiguity. This functionality is not commonly required for typical dbt usage.
 
 ```sql
 select * from {{ ref('package_name', 'model_name') }}
 ```
+
+### Hidden ref
```
tayloramurphy, line 48:
```suggestion
### Forcing Dependencies
```

tayloramurphy:
No problem! Thanks 🙂

tayloramurphy:
@clrcrl I think this is ready for your review.

clrcrl:
Thanks for the nudge :)

PR #272, opened 2020-07-09
Update tests.md

bashyroger:
I found the description of schema VS data tests in he beginning not that clear, think it is better now.


clrcrl:
Hi @bashyroger — I think we can improve these docs but I think this language might make it a little more confusing. Definitely see where you're coming from though! 

I've jotted it down as a to do though!

PR #284, opened 2020-07-14
Update preview img

amy-byrum:


amy-byrum:
So @clrcrl , this is the only place I could find reference to that dbt-logo-full (attached). I'm not sure if we actually even needed to update this as I couldn't find a lightbox actually using the logo anywhere but I updated it just in case. If you know and could enlighten me, that would be awesome! Or, if you'd rather play it safe, I can undo that commit.
Screen Shot 2020-07-14 at 11 32 06 AM

PR #292, opened 2020-07-20
Update profile.md

kylelundstedt:
Change kebab_case (not correct; kebab case would be "kebab-case") to snake_case


clrcrl:
Oh man, that's embarrassing 😀

PR #295, opened 2020-07-21
windows.md

gordonhwongmh:
Would be great to add a reference to 'dbt init' to this and other installation pages


clrcrl:
Hi Gordon! I agree our "setting up" docs could do with some improvement, but I'm hesitant to add this step to the installation docs — for uses joining an existing team, running "dbt init" won't make sense (they should instead clone a repo). So I'm going to close this issue, but I've definitely heard the feedback

PR #300, opened 2020-07-23
Update using-custom-database.md

gonzalodiaz:
Removing the `+` sign while declaring the database value on the .yml file.


clrcrl:
Same as #301

PR #301, opened 2020-07-23
Update using-custom-schemas.md

gonzalodiaz:
Remove the `+` sign when declaring a custom schema in the .yml file.


clrcrl:
Hi @gonzalodiaz — this is actually intentional! See [here](https://docs.getdbt.com/reference/advanced-config-usage#using-the--prefix)

PR #302, opened 2020-07-24
[v0.18.0] New selectors

jtcohen6:
Document new node selector functionality ahead of 0.18.0b2:
* methods: `config`, `test_type`, `test_name`, `package`
* intersections
* nth-parent/child
* version-controlled YML `selectors`

## todos
- [x] Merge https://github.com/fishtown-analytics/dbt/pull/2640
- ~Release 0.18.0b2~ I've decided I don't need to wait for this
- [x] Ensure example code is legit, especially selectors yml

PR #306, opened 2020-07-28
Prerelease docs site

jtcohen6:
I added Netlify config (`netlify.toml`) for branch-based deployment of the long-lived `next` branch, which will serve prerelease docs.

Here's the site: https://next--docs-getdbt-com.netlify.app/
For prettiness' sake, we'll want to CNAME this to `docs-next.getdbt.com`, but that's not _actually_ a blocker.

I defined two new environment variables:
* `ALGOLIA_INDEX_NAME`: `dbt` normally, `dbt-next` for prerelease docs
* `DISCLAIMER`: I just added the word "Prerelease " in some prominent places. We could/should add a bright warning banner to the top of the homepage / every page as well.

Also:
* Added `X-robots-tag = "noindex"` as a header for the `next` deployment
* Netlify already/automatically adds this header for all preview (PR-based) deployments
* In Netlify, I turned off all branch-based deployments that _aren't_ `master` or `next`. So the only deployments we'll have are `master` (production), `next` (prerelease), and previews of PRs that target either of those two branches.

My thinking is that we can merge this commit into `master` with absolutely zero effect, so that the two branches can stay even. Then, I'll merge #302 into the `next` branch.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,3 +11,10 @@ Include any notes about things that need to happen before this PR is merged, e.g
 - [ ] Change the base branch
 - [ ] Ensure PR #56 is merged
 -->
+
+## Pre-release docs
+Is this change related to an unreleased version of dbt?
+- [ ] Yes
+- [ ] No (if you're not sure, it's probably "No")
+
+If yes, please change the base branch of this PR to `next`
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,21 @@
+# Releasing docs
+
+There are two long-lived branches of this repo:
+  - `current` is the version of dbt released at [docs.getdbt.com](docs.getdbt.com)
+  - `next` reflects the next version of dbt to be releases. It is accessible at [docs-next.getdbt.com](docs-next.getdbt.com), and a warning banner is displayed on the site.
+
+- Any fixes or revisions to existing docs should be merged into `current`
+- Any pre-release docs should be merged into `next`
+- The `next` branch should be rebased from `current` somewhat regularly
+
+When new versions of dbt are released:
+1. Rebase `next` on top of `current`. The workflow might look something like this:
+```
+(current) $ git fetch
+(current) $ git pull
+(current) $ git checkout next
+(next) $ git pull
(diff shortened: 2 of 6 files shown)


clrcrl:
Could we consider adding `pre-release` as a yaml config to the page headers, and then rendering a warning based on that?

jtcohen6:
> Could we consider adding `pre-release` as a yaml config to the page headers, and then rendering a warning based on that?

@clrcrl I decided to add a [warning banner](https://v2.docusaurus.io/docs/theme-classic/#announcement-bar) in lieu of interpolating the word "Prerelease" in arbitrary places. Let me know if you had something quite different in mind!

clrcrl:
> @clrcrl I decided to add a warning banner

Very nice! We'll have to remember to remove this when we merge into master :)

Also I noticed that this PR is against `master`. We probably want to re-jig that. We should name this branch `add-pre-release-docs` and have it against the `next` branch instead.

I think we should also have a `RELEASE.md` file / a note in the pr template about which branch to make your PR against. I'm happy to do those as the current "maintainer of the docs" .

Let me fuss around with it and see how far I get with things.

jtcohen6:
> Very nice! We'll have to remember to remove this when we merge into master :)

The banner dynamically populates, using Netlify context variables, based on whether the branch being deployed is named `next`. So merging this code into `master` still wouldn't cause it to show up. I made one test to prove via one-time deployment from a branch that _isn't_ next: https://test-next--docs-getdbt-com.netlify.app/

> Also I noticed that this PR is against master. We probably want to re-jig that. We should name this branch add-pre-release-docs and have it against the next branch instead.

My goal in merging this into `master` now is to prove that the env-aware stuff works and have the branches at the same level before I merge any substantive content to the `next` branch. I want to avoid accidentally merging anything that _doesn't_ work at release time, when we'll merge commits from `next --> master`.

> I think we should also have a RELEASE.md file / a note in the pr template about which branch to make your PR against. I'm happy to do those as the current "maintainer of the docs" .

Great idea! I think _most_ community-contributed docs should still be against `master`, since any fixes or clarifications of existing features belong there. We should lock down the `next` branch, to some extent.

PR #308, opened 2020-07-28
Fix docs for test-paths and analysis-paths configs

clrcrl:
## Description & motivation
dbt has some non-intuitive behavior here. In the effort to be correct, I've documented it


website/docs/docs/building-a-dbt-project/tests.md
```diff
@@ -206,9 +206,9 @@ where {{ column_name }} is null
 ## Data tests
 A data test is a `select` statement that returns `0` records when the test is successful. Note: this differs from schema tests that return the number `0`.
 
-Data tests are defined in `.sql` files, typically in your `tests` directory. Each `.sql` file contains one data test / one `select` statement:
+Data tests are defined in `.sql` files, typically in your `tests` directory (as defined by your [`test-paths` config](test-paths)). Each `.sql` file contains one data test / one `select` statement:
 
-<File name='test/assert_total_payment_amount_is_positive.sql'>
+<File name='tests/assert_total_payment_amount_is_positive.sql'>
```
clrcrl, line 211: @beckjake — I changed this back since _most_ people won't change the defaults generated dbt `dbt init`, so we should cater for the masses her

clrcrl:
@jtcohen6 — can you sense check this one?

jtcohen6 (approved):
There remains a high potential for confusion, insofar as the underlying behavior is confusing and will remain so until we rationalize it. I think you've taken the right approach, which is to lay out all the interlaced pieces as clearly as possible.

PR #311, opened 2020-07-30
updates run command hyperlink

alexbethea22:
Currently goes here: https://docs.getdbt.com/faqs/failed-prod-run


clrcrl:
ooh we merged something else that fixed this!

PR #312, opened 2020-07-30
Update license.md to add clarity

jthandy:
We may not have touched this in a long time; the existing version is not reflective of the current way that "dbt" is used--"dbt" applies both to Core and Cloud, and as such we need to be more specific about our licenses.


clrcrl (approved):
I'll take your word on this one

PR #327, opened 2020-08-04
Update setting-up-sso-with-okta.md

alexbethea22:
Adds clarification around who/where enterprise users will get the deployment ID from.

### DIFF
--- a/website/docs/docs/dbt-cloud/dbt-cloud-enterprise/setting-up-sso-with-okta.md
+++ b/website/docs/docs/dbt-cloud/dbt-cloud-enterprise/setting-up-sso-with-okta.md
@@ -53,7 +53,7 @@ On the **SAML Settings** page, enter the following values:
 * **Single sign on URL**: `https://auth.getdbt.com/login/callback?connection=<your-deployment-id>`
 * **Audience URI (SP Entity ID)**: `urn:auth0:dbt-cloud:<your-deployment-id>`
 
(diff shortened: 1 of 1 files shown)


alexbethea22:
@eogilvy12 after speaking with Amy it sounds like wwe is the last integration that we will need to do this for. This is why I just updated that one line here!

website/docs/docs/dbt-cloud/dbt-cloud-enterprise/setting-up-sso-with-okta.md
```diff
@@ -53,7 +53,7 @@ On the **SAML Settings** page, enter the following values:
 * **Single sign on URL**: `https://auth.getdbt.com/login/callback?connection=<your-deployment-id>`
 * **Audience URI (SP Entity ID)**: `urn:auth0:dbt-cloud:<your-deployment-id>`
 
-Replace `<your-deployment-id>` with your dbt Cloud deployment ID. If you aren't sure what value you should use, please contact support (support@getdbt.com).
+Replace `<your-deployment-id>` with your dbt Cloud deployment ID. Your deployment ID will be provided to you by your Customer Success Manager during implementation.
```
eogilvy12, line 56: @amychen1776 when is this provided? I am wondering if this wording clarifies enough for people. I read "id" as a number which is I think what is most confusing here but interested if you have had clients get confused on this before?

eogilvy12:
Ahh well if we won't have to do this again in the future we may not have to even change it?

drewbanin:
FYI this is going away - see https://github.com/fishtown-analytics/docs.getdbt.com/pull/368

PR #328, opened 2020-08-04
Add docs for adapter.dispatch, adapter_macro

jtcohen6:
## Description & motivation

* Add documentation for `adapter.dispatch` (new in v0.18.0) and `adapter_macro` (marked for deprecation in v0.18.0, previously undocumented)
* I added `dispatch` to the existing `adapter` reference article. While it's much more complex than all other `adapter` methods, I think the reference section feels right; this information is relevant to developers, maintainers, contributors, or curious users of packages and plugins.

@clrcrl would you mind giving this a sense check all the same?

## Pre-release docs
- [x] Yes
- [ ] No (if you're not sure, it's probably "No")


clrcrl:
Sorry that I missed this :(

jtcohen6:
you're good! I'm just being selfish bc of office hours in an hour

PR #329, opened 2020-08-05
Feature: deferred runs

jtcohen6:
## Description & motivation

* Add docs for `--defer` and `--state`, flagged as a Beta Feature in v0.18.0
* Added to `run` reference section for now. Once the full Slim CI play is functional, we should add a Guide for the entire recommended workflow

@drewbanin I'd appreciate your feedback on the language around this being a "Beta Feature" (YMMV, some behavior liable to change), along the lines of our discussion yesterday @ product office hours

## Pre-release docs
- [x] Yes
- [ ] No (if you're not sure, it's probably "No")

### DIFF
--- a/website/docs/docs/guides/migration-guide/upgrading-to-0-18-0.md
+++ b/website/docs/docs/guides/migration-guide/upgrading-to-0-18-0.md
@@ -21,9 +21,11 @@ For more details, see [new and changed documentation](#new-and-changed-documenta
 - intersections
 - nth-parent/child
 - version-controlled YML selectors
+- [Beta Feature] Defer unselected node references to state defined by a previous run's artifacts
 
 ### Docs
 - Include static assets (such as images) in auto-generated docs site
+- Improved resource search
 
 ### Database-specific
 - Specify IAM profile when connecting to Redshift
@@ -51,3 +53,4 @@ Please be aware of the following changes in v0.18.0. While breaking, we do not e
 - [Redshift profile](redshift-profile#specifying-an-iam-profile)
 - [`asset-paths` config](asset-paths) (also updated [dbt_project.yml](dbt_project.yml.md) and the [description](description) docs to match)
 - [`impersonate_service_account` in the BigQuery profile configuration](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile#service-account-impersonation)
+- [Deferring to previous run state](run#deferring-to-previous-run-state)
--- a/website/docs/reference/commands/run.md
+++ b/website/docs/reference/commands/run.md
@@ -127,8 +127,8 @@ as it obviates the need to create all of those models' parents.
 
 Deferral and state can be set by environment variables as well as CLI flags:
 
-- `--defer` or `DEFER_MODE`
-- `--state` or `ARTIFACT_STATE_PATH`
+- `--defer` or `DBT_DEFER_TO_STATE`
+- `--state` or `DBT_ARTIFACT_STATE_PATH`
 
 If both the flag and env var are provided, the flag takes precedence.
 
--- a/website/docs/reference/commands/run.md
+++ b/website/docs/reference/commands/run.md
@@ -116,8 +116,13 @@ This is net-new functionality in v0.18.0, with iterative improvements to come.
 If you encounter unexpected behavior, please post in Slack or open an issue.
 :::
 
-dbt will resolve `ref` calls differently depending on whether a node is selected
-for a given run. If a node is unselected, references to it will resolve to the namespace 
+```
+dbt run --models [...] --defer --state path/to/artifacts
+```
+
+When the `--defer` flag is provided, dbt will resolve `ref` calls differently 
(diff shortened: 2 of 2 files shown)


website/docs/reference/commands/run.md
```diff
@@ -102,3 +102,96 @@ Encountered an error:
...
+
+This is especially useful when running a subset of models in a new test or development schema,
+as it obviates the need to create all of those models' parents.
+
+**Environment variables**
+
+Deferral and state can be set by environment variables as well as CLI flags:
+
+- `--defer` or `DEFER_MODE`
+- `--state` or `ARTIFACT_STATE_PATH`
```
beckjake:
```suggestion
- `--defer` or `DBT_DEFER_TO_STATE`
- `--state` or `DBT_ARTIFACT_STATE_PATH`
```

jtcohen6: ty!

website/docs/docs/guides/migration-guide/upgrading-to-0-18-0.md
```diff
@@ -51,3 +53,4 @@ Please be aware of the following changes in v0.18.0. While breaking, we do not e
 - [Redshift profile](redshift-profile#specifying-an-iam-profile)
 - [`asset-paths` config](asset-paths) (also updated [dbt_project.yml](dbt_project.yml.md) and the [description](description) docs to match)
 - [`impersonate_service_account` in the BigQuery profile configuration](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile#service-account-impersonation)
+- [Deferring to previous run state](run#deferring-to-previous-run-state)
```
drewbanin, line 56: this links to the wrong place because of the logic (read: bug) in the link checker component. Can you make sure this branch incorporates the latest `current` commit? Would like to make sure that https://github.com/fishtown-analytics/docs.getdbt.com/pull/334 works correctly now :)

website/docs/reference/commands/run.md
```diff
@@ -102,3 +102,96 @@ Encountered an error:
...
+ - The `--defer` and `--state` flags are new in dbt v0.18.0
+
+</Changelog>
+
+:::info Beta Feature
+This is net-new functionality in v0.18.0, with iterative improvements to come.
+If you encounter unexpected behavior, please post in Slack or open an issue.
+:::
+
+dbt will resolve `ref` calls differently depending on whether a node is selected
```
drewbanin: Might be clearer here to start with an example of a `dbt run` command that includes `--defer` and `--state`. I think it will help frame the first couple of paragraphs for new readers? Even something like "When the --defer flag is provided..." would be welcomed IMO

website/docs/reference/commands/run.md
```diff
@@ -102,3 +102,96 @@ Encountered an error:
...
+- `dbt run` (with no `--models` or `--exclude` specification) includes _all_
+nodes, even seeds and sources. In this way, `dbt run` and `dbt run --defer --state ...` always yield
+identical results, with no deferred nodes.
+- Ephemeral models are never deferred, since they serve as "passthroughs" for other `ref` calls.
+Including or excluding an ephemeral model makes no difference; including or excluding
+the non-ephemeral parent of an ephemeral model does.
+
+**Example**
+
+In my local development environment, I create all models in a schema named
```
drewbanin, line 153: This is a great example. It's not _super_ critical here, but I think we might want to give the terminology "unselected" vs. "not included" vs. "not selected" some thought. dbt's CLI flags (namely `--models`) make this a little unwieldy to talk about.

I ultimately don't have strong preferences on how we describe selected/unselected/not selected nodes, but I think it will be important for us to be consistent throughout the documentation and other resources.

website/docs/reference/commands/run.md
```diff
@@ -102,3 +102,96 @@ Encountered an error:
...
+`dbt_me`. In production, the same models are created in a schema named
+`prod`. Both are located in a database called `analytics`.
+
+Given a model `model_2`:
+
+<File name='models/model_2.sql'>
```
drewbanin, line 170: just flagging that this could be a neat use case for tabs, eg:
 - source code
 - without deferral
 - with deferral

take it or leave it

drewbanin (approved):
Approved but with caveats, lmk if you'd like to discuss at all!

website/docs/docs/guides/migration-guide/upgrading-to-0-18-0.md
jtcohen6: Confirmed that this works locally when I pulled in the change from `current`

website/docs/reference/commands/run.md
jtcohen6: Really good thought. I've opted for the verbose (but hopefully clear) "included in the model selection criteria." I think include/exclude is the right terminology since users are already familiar with `--exclude`.

website/docs/reference/commands/run.md
jtcohen6: I'll never say no to a tab

PR #330, opened 2020-08-05
Update as_bool.md to reference as_bool instead of as_number

nickwu241:
## Description & motivation

Update `as_bool.md` to reference `as_bool` instead of `as_number`

## Pre-release docs
- [ ] Yes


clrcrl:
Thank you!

PR #334, opened 2020-08-05
Scope links with leading slashes

drewbanin:
## Description & motivation
- Scope links with leading slashes
- Fail to build for ambiguous links

**Example:** You have three pages with the slugs:
 - `docs/guides/how-do-i-make-it-run`
 - `docs/using-dbt/run`
 - `reference/commands/run`

Previously, the Link component would match the first slug with the suffix specified in the link. So the markdown:
```
[run](run)
```

Would have linked to `docs/guides/how-do-i-make-it-run`. That was incorrect. Now, the Link component will add a leading slash before looking up a slug. This will make dbt resolve to either of:
 - `docs/using-dbt/run`
 - `reference/commands/run`

Unfortunately, a link like `[run](run)` is ambiguous, and we can't know which link was intended. With this PR, Docusaurus will fail to build if an ambiguous link is detected. There is more to do to make this easy to use, but it will hopefully be a good enough starting point for now.

## To-do before merge
I expect this build to fail with an ambiguous link error, and we should fix those errors before merging this PR.

## Pre-release docs
- [ ] Yes

### DIFF
--- a/website/docs/dbt-cli/installation-guides/macos.md
+++ b/website/docs/dbt-cli/installation-guides/macos.md
@@ -64,7 +64,7 @@ brew link dbt-development
 ```
 
 ## pip
-You can also use pip to install dbt – check out the [main installation documentation](installation) for more information.
+You can also use pip to install dbt – check out the [main installation documentation](dbt-cli/installation) for more information.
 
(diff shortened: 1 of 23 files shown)


clrcrl (approved):
LGTM!!! 🌈

drewbanin:
woop woop! I was worried these would be hard to track down from the error logs in Netlify. Nice work! shipping :)

PR #337, opened 2020-08-10
Clarify --strict flag usage

jtcohen6:
## Description & motivation

I've seen this question crop up a few times in dbt Slack. Based on the current language in the docs, it sounds like `--strict` does strict field validation (https://github.com/fishtown-analytics/dbt/issues/1570), rather than what it really does, which is nothing special since v0.15.0.

Let me know if I'm missing a technical piece of this.

## Pre-release docs
- [ ] Yes

### DIFF
--- a/website/docs/reference/global-cli-flags.md
+++ b/website/docs/reference/global-cli-flags.md
@@ -87,7 +87,9 @@ $ dbt --no-write-json run
 
 ## Strict
 
-The `-S` or `--strict` flag runs schema validations on dbt objects at runtime. This flag may incur a performance penalty, but it is useful for catching logic errors in development of the dbt project.
+The `-S` or `--strict` flag runs schema validations on dbt python objects at runtime. This flag may incur a performance penalty, but it is useful for catching logic errors in development of the dbt project.
+
+**N.B.** In versions >=0.15.0, dbt uses [hologram](https://github.com/fishtown-analytics/hologram) and [mypy](http://mypy-lang.org/) for object type declaration, validation, and testing. The `--strict` flag has no functional use except as an alias for `--warn-error`. We may choose to someday repurpose it.
 
 <File name='Usage'>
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/global-cli-flags.md
```diff
@@ -87,7 +87,9 @@ $ dbt --no-write-json run
 
 ## Strict
 
-The `-S` or `--strict` flag runs schema validations on dbt objects at runtime. This flag may incur a performance penalty, but it is useful for catching logic errors in development of the dbt project.
+The `-S` or `--strict` flag runs schema validations on dbt python objects at runtime. This flag may incur a performance penalty, but it is useful for catching logic errors in development of the dbt project.
+
+**N.B.** In versions >=0.15.0, dbt uses [hologram](https://github.com/fishtown-analytics/hologram) and [mypy](http://mypy-lang.org/) for object type declaration, validation, and testing. The `--strict` flag has no functional use except as an alias for `--warn-error`. We may choose to someday repurpose it.
```
clrcrl: I think if it truly is functionally the same as `--warn-error` we should use that as the _entire_ description, ya know?

> The `-S` or `--strict` flag is equivalent to the `--warn-error` flag. We may choose to someday repurpose it.

beckjake: There are a couple things strict does:
 - we do an extra `validate` call in some paths when you set `--strict`. This can be really slow as it requires serializing the object to json to validate it.
 - Connections type checks
 - some internal consistency checks during compilation

There's still no reason users should use --strict over --warn-error. We should rename the flags so `--strict` is just an alias for `--warn-error` and `--strict` becomes `--strict-development-checks` or something similarly verbose and obnoxious. You should only run it automatically via integration tests anyway.

jtcohen6:
Thanks for the comments, both!

@beckjake I think that re-aliasing makes sense.

For now, I made this a bit more concise, with the basic message of "don't use this."

clrcrl (approved):

PR #340, opened 2020-08-13
Suggest dbt run before docs generate

beckjake:
## Description & motivation
I'm not sure this is the best way to describe this, but we should note somehow that if you don't use `dbt run`, `dbt docs generate` will only know about columns described in your configuration files.

## Pre-release docs
- [ ] Yes
- [X] No (if you're not sure, it's probably "No")


jtcohen6 (approved):
Good addition, worth getting in there

clrcrl (approved):
LGTM!

PR #341, opened 2020-08-14
Delete FAQ schema-yml.md because it's an empty page

nickwu241:
## Description & motivation
This page is currently empty. Not sure if this page supposed to have content?
I've checked usage of reference `schema-yml` but doesn't seem like anywhere references it -- https://github.com/fishtown-analytics/docs.getdbt.com/search?q=schema-yml&type=Code

https://docs.getdbt.com/faqs/schema-yml/
image

On the FAQ page, it also shows up as empty
https://docs.getdbt.com/faqs/all
image

context: I was searching the schema of `schema.yml` for fields we can write

## Pre-release docs


clrcrl:
This is very valid haha!

PR #347, opened 2020-08-18
Update quoting.md

ghost:
Comment says unquote identifier, but it was quoted in source()

## Pre-release docs
- [ ] Yes


clrcrl:
Hmm, I think the docs could be clearer here, but the code snippet is correct. We currently have:
```
-- here, the identifier should be unquoted
left join {{ source('jaffle_shop', 'customers') }} using (order_id)
```

I think maybe we want it to read:

```
-- here, the identifier should be unquoted when it gets compiled to a table name
left join {{ source('jaffle_shop', 'customers') }} using (order_id)
```

You definitely need to quote `'customers'` — it's the compiled code that changes :)

PR #349, opened 2020-08-19
Breaking commit: upgrade docusuarus versions

clrcrl:
## Description & motivation
[Docusaurus](https://v2.docusaurus.io/docs/) has a cool `cmd + k` shortcut for search. I wanted to bump our version to get it on our docs too. But something broke....


annafil:
Going to close this out, not because we don't need to upgrade but because this is very far behind in git history and it probably makes sense to start a new PR off a more recent tree.

PR #350, opened 2020-08-24
[DRAFT] Add docs on load_result macro

clrcrl:
## Description & motivation
We don't do a great job of documenting this, and it comes up enough in Slack / dbt Learn that I think we should have something better.

I think this needs a bit of refinement:
- I want to change the docs to be "how to operate on the `data` object" and "how to operate on the `table` object". That way, we can also use the same docs for the results of `run_query` (since that just returns the `table` object, not the other two parts).
- The examples feel clunky at the moment, trying to figure out a good way to render these. I kind of want "raw" / "compiled" side by side, yanno? Or like a python notebook interface.

ANYWAY, this will do for now

## Pre-release docs

## Checklist

### DIFF
--- a/website/docs/reference/dbt-classes.md
+++ b/website/docs/reference/dbt-classes.md
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
     'required': ['node'],
 }
 ```
+
+
+## SQL Results
+:::warning These docs are a work in progress
+:::
+
+Note for the discerning reader: this is not a true python Class — it's a dictionary.
+
+The [`load_result` macro](load_result) macro returns a SQL Results object. The schema of a SQL Results Object is as follows ([source code](https://github.com/fishtown-analytics/dbt/blob/dev/marian-anderson/core/dbt/context/providers.py#L667)):
+
+```python
+{
+    'status': 'SUCCESS 1',
+    'data': [(Decimal('1'),)],
+    'table': <agate.table.Table object at 0x11466ad90>
+}
+```
+
+Note that this contract may be subject to change in the future until dbt's API has reached a "stable" state.
+
+Keys:
+- `status`: The status code returned by the database. Note: this varies across databases and query types.
+- `data`: An array of tuples represent the results. The first item represents the first row. For example, for a table with results:
+
+| my_col_1 | my_col_2 |
+|----------|----------|
+| 1        | 2        |
+| 3        | 4        |
+
+```python
+> results['data']
+
+[
+    (Decimal('1'), Decimal('2')),
+    (Decimal('3'), Decimal('4'))
+]
+
+```
+
+- `table`: An [Agate Table](https://agate.readthedocs.io/en/1.6.1/api/table.html) containing results. This is a powerful API for interacting with query results, but you may need to use the Agate docs to navigate it.
+
+### Examples
+
+All examples use this setup SQL:
+
+```sql
+{%- call statement('test_query', fetch_result=True) -%}
+
+    select 1 as my_col_1, 2 as my_col_2
+    union all
+    select 3 as my_col_1, 4 as my_col_2
+
+{%- endcall -%}
+
+{%- set test_query = load_result(name='test_query') -%}
+```
+
+We've included examples of how to use the [`log` macro](log) to print results to the command line in order to "see" what the object is.
+
+#### Returning the first row
+
+You can use both the `data` and the `table` properties to return this.
+
+Using the `data` property, both of these are equivalent:
+
+```sql
+
+{% set first_row=test_query['data'][0] %}
+
+{{ log(first_row, info=True) }}
+
+{% set first_row=test_query.data[0] %}
+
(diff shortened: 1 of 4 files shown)


clrcrl:
@jtcohen6 — I'm keen to get your initial feedback on these docs. I know they aren't quite right yet (see above), but figuring out how we can do this in a better way!

website/docs/reference/dbt-classes.md
```diff
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
...
+    'data': [(Decimal('1'),)],
+    'table': <agate.table.Table object at 0x11466ad90>
+}
+```
+
+Note that this contract may be subject to change in the future until dbt's API has reached a "stable" state.
+
+Keys:
+- `status`: The status code returned by the database. Note: this varies across databases and query types.
+- `data`: An array of tuples represent the results. The first item represents the first row. For example, for a table with results:
```
jtcohen6:
```suggestion
- `data`: An array of tuples representing the results. The first item represents the first row. For example, for a table with results:
```

website/docs/reference/dbt-classes.md
```diff
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
...
+```
+
+
+#### Returning the column names of a query result
+Included for illustrative purposes only.
+
+Consider using the built-in [`get_columns_in_query` macro](https://github.com/fishtown-analytics/dbt/blob/dev/marian-anderson/core/dbt/include/global_project/macros/adapters/common.sql#L1) instead.
+
+```
+{% set column_names=test_query.table.columns | map(attribute='name') | list }
```
jtcohen6:
```suggestion
{% set column_names=test_query.table.columns | map(attribute='name') | list %}
```

website/docs/reference/dbt-classes.md
```diff
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
...
 ```
+
+
+## SQL Results
+:::warning These docs are a work in progress
+:::
+
+Note for the discerning reader: this is not a true python Class — it's a dictionary.
+
+The [`load_result` macro](load_result) macro returns a SQL Results object. The schema of a SQL Results Object is as follows ([source code](https://github.com/fishtown-analytics/dbt/blob/dev/marian-anderson/core/dbt/context/providers.py#L667)):
```
jtcohen6:
```suggestion
The [`load_result` macro](load_result) returns a SQL Results object. The schema of a SQL Results Object is as follows ([source code](https://github.com/fishtown-analytics/dbt/blob/dev/marian-anderson/core/dbt/context/providers.py#L667)):
```

website/docs/reference/dbt-classes.md
```diff
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
...
+    (Decimal('3'), Decimal('4'))
+]
+
+```
+
+- `table`: An [Agate Table](https://agate.readthedocs.io/en/1.6.1/api/table.html) containing results. This is a powerful API for interacting with query results, but you may need to use the Agate docs to navigate it.
+
+### Examples
+
+All examples use this setup SQL:
```
jtcohen6:
```suggestion
The examples below use this setup SQL:
```

website/docs/reference/dbt-classes.md
```diff
@@ -167,3 +167,201 @@ The execution of a resource in dbt generates a Result object. This object contai
...
+{%- set test_query = load_result(name='test_query') -%}
+```
+
+We've included examples of how to use the [`log` macro](log) to print results to the command line in order to "see" what the object is.
+
+#### Returning the first row
+
+You can use both the `data` and the `table` properties to return this.
+
+Using the `data` property, both of these are equivalent:
```
jtcohen6, line 233: They're equivalent for [Jinja reasons](https://jinja.palletsprojects.com/en/2.11.x/templates/#variables), not because of anything special about `data` / `SQL Results`. (That's the implication I read here)

jtcohen6 (commented):
This looks good! It's definitely better to add these docs than to go on not having it. I like the framing of SQL Results as a special dbt "class." I left a few pedantic/proofread comments.

It's a pretty long vertical scroll to get through the whole section. What do you think about using tabs to alternate between `data` and `table`?

annafil:
I would love to see this merged in -- it looks like a lot of effort went into this one!

@jtcohen6 is this still relevant/useful now that core is a few versions ahead? 

If yes, thoughts on making this into sub-pages? It's a loooong scroll even without this addition and might make things easier to parse.

runleonarun:
Closing this PR so we can plan the page appropriately while addressing the challenges stated above. Capturing this in an issue to be addressed when possible. https://github.com/dbt-labs/docs.getdbt.com/issues/2001 

All comments should be captured in the issue going forward. Thank you for your work on this, `@clrcrl`!

PR #353, opened 2020-08-25
fix codeowners precedence

drewbanin:
I had these inverted - lower entries on the list take precedence over higher entries on the list! Via https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

If there are any other codeowners adjustments we should make, let's do them here :)


clrcrl (approved):
That truly is not intuitive...

PR #357, opened 2020-08-25
minor partial parsing correction

beckjake:
## Description & motivation

Since we invalidate the results on any vars change, part of this warning isn't really true anymore, unless there's some third way to set vars that I'm forgetting about.

We could also expand on the new bit I added. The current list of things that invalidate the partial parse result:
 - `dbt_project.yml` changing at all
 - `profiles.yml` changing at all
 - the `--profile` argument changing
 - the `--target` argument changing
 - the cli `--vars` arguments changing at all (even being the same dicts in different order would probably trigger this)

But I thought maybe we should just keep it simple.

## Pre-release docs


clrcrl (approved):
Simple looks good to me

PR #358, opened 2020-08-26
Update snapshots.md

mikekaminsky:
## Description & motivation

I ran into this opaque error message when trying to set up snapshotting in my project the first time:

```
Compilation Error in model scorecard_history (models/snapshots/scorecard_history.sql)
  Encountered unknown tag 'snapshot'.
```

It turns out I needed to enable snapshots in my `dbt_project.yml` but this was not at all obvious to me when reading these docs! I imagine someone less familiar with dbt would find this even more bewildering. I wanted to note at least somewhere on this page that that configuration is required and include the error message so that it's searchable on the page if someone else runs into it.

## Pre-release docs

### DIFF
--- a/website/docs/docs/building-a-dbt-project/snapshots.md
+++ b/website/docs/docs/building-a-dbt-project/snapshots.md
@@ -65,6 +65,14 @@ When you run the [`dbt snapshot` command](snapshot):
 
 Snapshots can be referenced in downstream models the same way as referencing models — by using the [ref](ref) function.
 
+*Note*: You'll need to have snapshots configured in your `dbt_project.yml` file for this to work. You're looking for a line like this:
+
+```yaml
+snapshot-paths: ["snapshots"]
+```
+
+If that configuration isn't set, when you run `dbt snapshot` you might seen an error that says something like `Encountered unknown tag 'snapshot'.`
+
 ## Example
 To add a snapshot to your project:
 


clrcrl:
huh, is this so?

My reading of [this line](https://github.com/fishtown-analytics/dbt/blob/dev/marian-anderson/core/dbt/config/project.py#L408) made me think that there was a default value here!

I'll test it out tomorrow – if you're right we'll need to update [this doc](https://github.com/fishtown-analytics/docs.getdbt.com/blob/current/website/docs/reference/project-configs/snapshot-paths.md) too

mikekaminsky:
I might have deleted that line when I was first setting up the project 😬 

Regardless, I think it's useful to have this part of the docs call out that that's where the path is defined in case someone ends up in this same state or wants to adjust that path.

clrcrl:
Oh wait, I see what has happened. So the way to fix this error it to move your file from:
```
models/snapshots/scorecard_history.sql
```
to:
```
snapshots/scorecard_history.sql
```

Then it will work regardless of whether you have the `snapshot-paths` config or not — in this example it's commented out
```
$ cat dbt_project.yml # note snapshot-paths is commented out

name: 'jaffle_shop'
version: '1.0.0'
config-version: 2

profile: 'jaffle_shop'

source-paths: ["models"]
analysis-paths: ["analysis"]
test-paths: ["tests"]
data-paths: ["data"]
macro-paths: ["macros"]
# snapshot-paths: ["snapshots"]

target-path: "target"  # directory which will store compiled SQL files
clean-targets:         # directories to be removed by `dbt clean`
    - "target"
    - "dbt_modules"

$ dbt snapshot
Running with dbt=0.18.0-b2
Found 5 models, 24 tests, 1 snapshot, 0 analyses, 140 macros, 0 operations, 0 seed files, 3 sources

11:30:02 | Concurrency: 1 threads (target='learn')
11:30:02 |
11:30:02 | 1 of 1 START snapshot dbt_snapshots.test_snapshot.................... [RUN]
11:30:06 | 1 of 1 OK snapshotted dbt_snapshots.test_snapshot.................... [SUCCESS 0 in 4.21s]
11:30:07 |
11:30:07 | Finished running 1 snapshot in 10.74s.

Completed successfully

Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

mikekaminsky:
🤦

mikekaminsky:
Ouch. Okay that makes sense! We could change the line about "generally found in the `snapshots/` directory" to "generally found in the `snapshots/` directory (*not* the `models/snapshots/` directory)"

But we could also just close this because I'm an idiot.

clrcrl:
hehe yes we absolutely can!

clrcrl:
I'm going to close this PR in the meantime

PR #407, opened 2020-10-01
Update sql_header docs

clrcrl:
## Description & motivation
We clarified this ourselves recently. Good to clarify it for the community

## To do
My read of the source code is that this is a config that overrides rather than accumulates (pre-hooks accumulate). Do we want to call this out explicitly?

## Pre-release docs
- [ ] Yes: please

### DIFF
--- a/website/docs/reference/resource-configs/sql_header.md
+++ b/website/docs/reference/resource-configs/sql_header.md
@@ -2,25 +2,130 @@
 resource_types: [models]
 datatype: "string"
 ---
-:::caution Heads up!
-These docs are a work in progress.
 
-:::
+<Tabs
+  defaultValue="models"
+  values={[
+    { label: 'Models', value: 'models', },
+    { label: 'Seeds', value: 'seeds', },
+    { label: 'Snapshots', value: 'snapshots', },
+  ]
+}>
+<TabItem value="models">
 
-<!----
-To-do:
-- use the reference doc structure for this article
-- check if SQL headers apply to models only
---->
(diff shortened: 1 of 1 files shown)


website/docs/reference/resource-configs/sql_header.md
```diff
@@ -2,25 +2,130 @@
...
+
+models:
+  +sql_header: "alter session set timezone = 'Australia/Sydney'"
+```
+
+</File>
+
+### Create a BigQuery Temporary UDF
+
+This example calls the `set_sql_header` macro. This macro is a convenience wrapper which you may choose to use if you have a multi-line SQL statement to inject. You do not need to use the `sql_header` configuration key in this case.
```
clrcrl, line 128: it's still not clear to me why we provide this method. Why not just do:
```
{% set my_bq_udf %}
-- this is sql
{% endset %}

{{ config(sql_header = my_bq_udf) }}

```

jtcohen6: No real reason, other than that's how the original contributor wanted to use it (https://github.com/fishtown-analytics/dbt/pull/1967)

jtcohen6 (approved):
Looks good!

What do you think of adding "when to use `pre-hook` and when to use `sql_header`" as an FAQ?

clrcrl:
For now, I've got a "Comparison to pre-hooks" header which I think solves most of it. We can revisit later if we see the question popping up again :)

PR #416, opened 2020-10-14
Update setting-up-snowflake-sso.md to include multiple projects

amychen1776:
## Description & motivation
We have been getting the question of how many security integrations do I need if I'm connected to the same Snowflake account, only to multiple dbt Cloud project. Just wanted to clarify this in the docs.


drewbanin (approved):
LGTM

PR #430, opened 2020-10-21
Add faq on having more than one database connection per project.

### LINKED ISSUE #25: FAQ: Can I connect to two databases?

VersusFacit:
## Description & motivation
Closes #25 . 

I tried to have something other than just a frank no, but if there's too much, I'm happy to trim it down. Was trying to provide context but may have given too much? Do let me know.

## Pre-release docs
- [ ] Yes: please

### DIFF
--- a/website/docs/faqs/connecting-to-two-dbs-not-allowed.md
+++ b/website/docs/faqs/connecting-to-two-dbs-not-allowed.md
@@ -1,0 +1,12 @@
+---
+title: Can I connect my dbt project to two databases?
+---
+
+No. dbt talks to one database at a time, namely your data warehouse.
+
+The modern data stack reflects ELT thinking (i.e. extract -> load -> transform). The three components of ELT are all modular services--building blocks--that can be assembled into robust pipelines.
+
+dbt enters the picture after your extractors and loaders have funneled data into a warehouse. Hence, instead of thinking "how do I connect my dbt project to two databases", ask "what loader services will best prepare our warehouse for dbt transformations."
+
+For more on the modern data stack, see the "dbt and the modern BI stack" section of this [dbt blog post](https://blog.getdbt.com/what--exactly--is-dbt-/).
+
--- a/website/docs/faqs/connecting-to-two-dbs-not-allowed.md
+++ b/website/docs/faqs/connecting-to-two-dbs-not-allowed.md
@@ -2,11 +2,13 @@
 title: Can I connect my dbt project to two databases?
 ---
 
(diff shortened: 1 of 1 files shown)


website/docs/faqs/connecting-to-two-dbs-not-allowed.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Can I connect my dbt project to two databases?
+---
+
+No. dbt talks to one database at a time, namely your data warehouse.
```
clrcrl: Hmmmmm so there is a caveat here.

If someone means two different databases in the same Snowflake account (or projects in a BigQuery account), then the answer could be "yes", as long as the data is queryable from one set of credentials. So you might need to reword here.

clrcrl (changes requested):

website/docs/faqs/connecting-to-two-dbs-not-allowed.md
amychen1776: To add to this ^

I think it would be useful to clarify that while we can't connect to two Snowflake accounts, we can connect to one Snowflake account with multiple databases. I think a key confusion that is creating this is because some people use databases and data warehouses interchangeably, especially because you can connect dbt to Postgres.

website/docs/faqs/connecting-to-two-dbs-not-allowed.md
```diff
@@ -0,0 +1,12 @@
...
+
+The modern data stack reflects ELT thinking (i.e. extract -> load -> transform). The three components of ELT are all modular services--building blocks--that can be assembled into robust pipelines.
+
+dbt enters the picture after your extractors and loaders have funneled data into a warehouse. Hence, instead of thinking "how do I connect my dbt project to two databases", ask "what loader services will best prepare our warehouse for dbt transformations."
```
amychen1776: I would love to widen this context by diving into what moves with dbt. We don't move data from your source to your warehouse but we can `move` data in your data warehouse by transforming it.

website/docs/faqs/connecting-to-two-dbs-not-allowed.md
VersusFacit: this is the problem with database being used interchangeable with database management system. The two are certainly not the same, but end up being in conversation. I'll make edits shortly. Thanks

clrcrl (approved):
LGTM! Thanks for making those changes and picking this back up!

PR #439, opened 2020-10-27
Update/add an example of database switching onto the target reference page. 

amychen1776:
## Description & motivation
It's useful to note that you can change your database depending on the target. A lot of enterprise folks need to do this because of how their environments are configured (In order to maintain the multiple silo-ed environments). 
This code snippet that I'm including is one that I've had to share with multiple clients.

### DIFF
--- a/website/docs/reference/dbt-jinja-functions/target.md
+++ b/website/docs/reference/dbt-jinja-functions/target.md
@@ -63,3 +63,22 @@ from source('web_events', 'page_views')
 where created_at >= dateadd('day', -3, current_date)
 {% endif %}
 ```
+
+### Use `target.name` to change your source database
+If you have specific Snowflake databases configured for your dev/qa/prod environments,
+you can set up your sources to compile to different databases depending on your 
+environment. 
+
+```yml
+version: 2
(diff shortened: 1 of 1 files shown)


website/docs/reference/dbt-jinja-functions/target.md
```diff
@@ -63,3 +63,22 @@ from source('web_events', 'page_views')
...
+
+```yml
+version: 2
+ 
+sources:
+  - name: source_name 
+    database: |
+      {%- if  target.name == "dev" -%} raw_dev
+      {%- elif target.name == "qa"  -%} raw_qa
+      {%- elif target.name == "prd"  -%} raw_prod
```
clrcrl:
```suggestion
      {%- elif target.name == "prod"  -%} raw_prod
```

clrcrl (approved):
One small potential typo that I noticed! I haven't tested this out so I assume the code works ;)

Overall this is really cool, I wonder whether we should also include a note like:

Alternatively, this can be written as:
```
- name: source_name
  database: "raw_{{ target.name }}"
```

PR #465, opened 2020-11-19
Quote wildcard

max-sixty:
## Description & motivation

At least on my shell — the `*` wildcard needs to be quoted to avoid the shell attempting to expand it. Or are others' shells different? (very possible!)

## Pre-release docs
- [ ] Yes: please

### DIFF
--- a/website/docs/reference/node-selection/syntax.md
+++ b/website/docs/reference/node-selection/syntax.md
@@ -56,7 +56,7 @@ dbt supports a shorthand language for defining subsets of nodes. This language u
 $ dbt run --models my_first_model my_second_model
 
 # these arguments can be projects, models, directory paths, tags, or sources
-$ dbt run --models tag:nightly my_model finance.base.*
+$ dbt run --models tag:nightly my_model 'finance.base.*'
 
 # use methods and intersections for more complex selectors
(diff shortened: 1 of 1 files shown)


clrcrl:
Hmm I think this is a `zsh` vs `bash` difference, I'm not sure which is more popular @jtcohen6, do you have a sense?

max-sixty:
It does work on bash! I'm using fish, which is niche enough that I could understand closing this — your call. 

The single quoting is never functionally worse, though maybe aesthetically so.

annafil:
Hey there @max-sixty! Thanks for putting thought into this one and sorry we dropped the ball on this thread. I suspect we have more than this one example in the docs, and I'd rather not ask you to go find them and change them all :) 

What do you think about creating a section on the [FAQ instead](https://docs.getdbt.com/faqs/all) to help others discover this if they run into a similar issue?  You can check out [this folder](https://github.com/fishtown-analytics/docs.getdbt.com/tree/current/website/docs/faqs) for examples. I think you just need to check in a .md file in the right format, and it will show up. ✨ 

In the meantime, I'll close out this PR but I'll be here to 👍 the next one :) And if you don't hear from anyone after a few days, feel free to nudge me in the Community slack 🙇‍♀️

PR #468, opened 2020-11-23
Remove 'coming soon' note re docs in dbt cloud

jaypeedevlin:
## Description & motivation
This seems like it is an old note that can be removed.


clrcrl (approved):
Thank you!!

drewbanin (approved):
Nice one - thanks!

PR #492, opened 2020-12-22
Update adapter.md

paweljmachnik:
Remove excessive bracket in the create_schema usage example.

## Description & motivation

I'm removing a typo in the create_schema usage example, to save the effort of finding this issue for the next person that will use it.


clrcrl (approved):
Ah, thank you so much!

Issue #27, opened 2020-02-27
Feature: FAQ: How do I run tests on sources only?


clrcrl:

runleonarun:
I am closing this issue and putting it in the backlog. Labeled with "icebox" so we can revisit at a later date.

PR #512, opened 2021-01-16
Update notes on test selection and invocation.

VersusFacit:
## Description & motivation
This started with making sure I could close issue #27. The "test all sources" syntax is actually misleading for unix command lines. So, I fixed the broken syntax bits.

But, in the course of attacking this, I discovered my own understanding of test selectors was generally flawed and the documentation was a bit obscure at places. After learning what was going on, I tried to "decouple" what `--model` `--select` and `--selector` with a few more sentences of context and slightly different organization of ideas. When streamlining the language and adding more examples, I also found it useful to explain plainly what a node is.

In addition to several additions, this PR closes issue #27 .

## Pre-release docs
- [ ] Yes: please

### DIFF
--- a/website/docs/faqs/test-sources.md
+++ b/website/docs/faqs/test-sources.md
@@ -2,10 +2,12 @@
 title: How do I run tests on sources only?
 ---
 
-Because the test selection syntax grew out of the model selection syntax (and pre-dates sources), the syntax here is a little unintuitive, but it is possible!
-
+Test selection syntax predates sources. It is based on model selection syntax:
 ```
-$ dbt test --models source:*
+$ dbt test --models source:<source_name>.<table_name>  # run tests on a single source table
+$ dbt test --models source:<source_name>               # run tests on a source
+$ dbt test --models source:\*                          # run all tests for all sources
 ```
+Take special note of the slash, or escape character. This is necessary for command prompts to parse * as a character and not what is sometimes called a [wildcard](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm).
 
-Check out the [model selection syntax documentation](node-selection/test-selection-examples) for more operators and examples.
+See our [test selection examples](node-selection/test-selection-examples) for more operators and examples.
--- a/website/docs/faqs/test-sources.md
+++ b/website/docs/faqs/test-sources.md
@@ -6,8 +6,8 @@ Test selection syntax predates sources. It is based on model selection syntax:
 ```
 $ dbt test --models source:<source_name>.<table_name>  # run tests on a single source table
 $ dbt test --models source:<source_name>               # run tests on a source
-$ dbt test --models source:\*                          # run all tests for all sources
+$ dbt test --models 'source:*'                         # run all tests for all sources
 ```
-Take special note of the slash, or escape character. This is necessary for command prompts to parse * as a character and not what is sometimes called a [wildcard](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm).
+On some systems, the slash, or escape character may be necessary. Some command prompts parse the * not as a character but as a [wildcard](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm). Sometimes single quoting disables the wildcard feature. Sometimes, \* may be necessary.
 
 See our [test selection examples](node-selection/test-selection-examples) for more operators and examples.
--- a/website/docs/reference/node-selection/syntax.md
+++ b/website/docs/reference/node-selection/syntax.md
@@ -2,7 +2,14 @@
 title: "Syntax overview"
 ---
 
-dbt's node selection syntax makes it possible to run only specific resources in a given invocation of dbt. This selection syntax is used for the following subcommands:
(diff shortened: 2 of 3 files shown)


website/docs/faqs/test-sources.md
```diff
@@ -2,10 +2,12 @@
...
 ---
 
-Because the test selection syntax grew out of the model selection syntax (and pre-dates sources), the syntax here is a little unintuitive, but it is possible!
-
+Test selection syntax predates sources. It is based on model selection syntax:
 ```
-$ dbt test --models source:*
+$ dbt test --models source:<source_name>.<table_name>  # run tests on a single source table
+$ dbt test --models source:<source_name>               # run tests on a source
+$ dbt test --models source:\*                          # run all tests for all sources
```
jtcohen6: IMO it's less confusing to single-quote this expression, rather than adding backslashes:
```suggestion
$ dbt test --models 'source:*'                          # run all tests for all sources
```

Then our note can be that, on some systems/shells, it's necessary to single-quote expressions containing wildcard characters.

website/docs/reference/node-selection/syntax.md
```diff
@@ -13,14 +20,9 @@ dbt's node selection syntax makes it possible to run only specific resources in
...
 
-:::info Nodes and resources
-
-We use the terms <a href="https://en.wikipedia.org/wiki/Vertex_(graph_theory)">"nodes"</a> and "resources" interchangeably. These  encompass all the models, tests, sources, seeds, snapshots, and analyses in your project. They are the objects that make up dbt's DAG (directed acyclic graph).
-:::
-
 ## Specifying resources
 
-By default, `dbt run` executes _all_ of the models in the dependency graph; `dbt seed` creates all seeds, `dbt snapshot` performs every snapshot. The `--models` and `--select` flags are used to specify a subset of nodes to execute.
+By default, `dbt run` executes _all_ of the models in the dependency graph; `dbt seed` creates all seeds, `dbt snapshot` performs every snapshot. The `--models` and `--select` flags are used to specify a subset of nodes to execute. The `--selector`` flag specifies which [user-defined yaml selector](http://localhost:3000/reference/node-selection/yaml-selectors) to use.
```
jtcohen6:
```suggestion
By default, `dbt run` executes _all_ of the models in the dependency graph; `dbt seed` creates all seeds, `dbt snapshot` performs every snapshot. The `--models` and `--select` flags are used to specify a subset of nodes to execute. The `--selector`` flag specifies which [user-defined yaml selector](yaml-selectors) to use.
```

website/docs/reference/node-selection/test-selection-examples.md
```diff
@@ -2,16 +2,18 @@
...
+dbt test -m <path.to.models.folder>  # run tests on models in a sub directory
+dbt test -m +<model_name>+           # run tests on models upstream/downstream of a model
+```
 
-Tests have their own properties _and_ inherit the properties of the nodes they select from. This means you:
-* select tests based on the file path of the models being tested, rather than the file paths of the `.yml` files that configure the tests
-* can use selector methods that check config properties of the resources being tested
+Tests have their own properties (set in yaml files or data test `config` blocks). Tests _also_ inherit the properties of the nodes they test (i.e. select from). This enables you to select which nodes to test using:
+* the file paths of the models themselves, as opposed to paths of `.yml` files where schema tests are defined
+* the properties of a source or model
```
jtcohen6: I'm hesitant about the change from `This means you` to `This enables you to`.

It's not currently possible to select tests based on the `.yml` files in which they are defined. If you want to select them by file path, it must be the path of the model `.sql` files. The phrase `This enables you to` makes it sound a bit more like a choice, when it only really works one way. (I'm definitely not saying the way it works is ideal.)

Similarly, we should lay out exactly how test selection differs from model selection.

A test is selected if the `--models` criteria includes any property of:
- the test itself (e.g. `tag:` or `test_name:` selection method)
- a resource (model/source/snapshot/seed) the test directly depends on

By the same token, a test is _not_ selected if the `--exclude` criteria includes any property of:
- the test itself
- a resource the test directly depends on

website/docs/reference/node-selection/test-selection-examples.md
```diff
@@ -2,16 +2,18 @@
...
 
-Things start to get a little unfamiliar when you want to test things other than models, so we've included lots of examples below. In the future, we plan to make this syntax more intuitive.
+Selecting nodes other than models, such as sources, deviates from the usual model selection syntax, so we've listed several examples below. In the future, we plan to make test selection syntax more intuitive.
```
jtcohen6: We're selecting tests that directly depend on sources captured by the `source:` selection method. It's a subtle point. But I don't think it `deviates` from the way other test selection works. My read of the previous language here is to take "unfamiliar" in the sense of less-intuitive, not in the sense of being functionally different.

jtcohen6 (commented):
I left a few comments! In general I find `''` easier to read than `\` escaping, if the goal is to present cross-shell-compatible arguments.

I wrote up a much longer thought regarding the rephrased bits about test selection, and I ended up opening it as a new issue: https://github.com/fishtown-analytics/docs.getdbt.com/issues/514

alexiswo:
This PR is from Jan. 2021. I checked all of the changes suggested and they are all valid and great - however the testing docs have already been updated in the meantime to reflect all of this helpful stuff. Closing this PR out.

PR #513, opened 2021-01-17
Update dbt-project-yml-context.md

mmeasic:
## Description & motivation
The `project_name` is not available in the `dbt_project.yml` context. 
'project_name' is available in: query headers, hooks, models, and macro execution

## To-do before merge
See: https://github.com/fishtown-analytics/dbt/pull/2085

## Pre-release docs
- [ ] Yes: please


jtcohen6 (approved):
Good call! You're right, `project_name` is available in the hook context, but not the general `dbt_project.yml` context.

Surprisingly, `project_name` renders to `''` (instead of `None`), so using it in, say, conditional statements within project-level config does not raise a compilation error, it just doesn't work the way one might expect.

mmeasic:
> Good call! You're right, `project_name` is available in the hook context, but not the general `dbt_project.yml` context.
> 
> Surprisingly, `project_name` renders to `''` (instead of `None`), so using it in, say, conditional statements within project-level config does not raise a compilation error, it just doesn't work the way one might expect.

Yes! Cheers.

PR #520, opened 2021-01-19
Minor doc code change

anderseriksson:
The first approach's column name should be the same in the model definition (SQL) as well as in the schema, right?

## Description & motivation

Minor doc

## Pre-release docs
Is this change related to an unreleased version of dbt?No


clrcrl (approved):
Thank you!

PR #528, opened 2021-01-21
Update redshift-profile.md

MBarendregt:
## Pre-release docs
- [ ] Yes: please
- [x ] No: please ensure the base branch is `current`

### DIFF
--- a/website/docs/reference/warehouse-profiles/redshift-profile.md
+++ b/website/docs/reference/warehouse-profiles/redshift-profile.md
@@ -23,8 +23,7 @@ company-name:
       threads: 4
       keepalives_idle: 0 # default 0, indicating the system default
       # search_path: public # optional, not recommended
-      sslmode: [optional, set the sslmode used to connect to the database]
-      sslrootcert: [optional, This parameter specifies the name of a file containing SSL certificate authority (CA) certificate(s). If the file exists, the server's certificate will be verified to be signed by one of these authorities. The default is ~/.postgresql/root.crt.]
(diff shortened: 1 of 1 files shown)


jtcohen6 (commented):
Hey @MBarendregt, thanks for the PR! I'm curious to hear more about this.

`sslmode` is definitely a supported config for postgres ([source](https://github.com/fishtown-analytics/dbt/blob/cb460a797c5cf73567fb42f17fba165e12facd80/plugins/postgres/dbt/adapters/postgres/connections.py#L25), [docs](https://docs.getdbt.com/reference/warehouse-profiles/postgres-profile/#sslmode)).

Since `dbt-redshift` also uses `psycopg2` and builds off the connection logic in `dbt-postgres`—crucially, it inherits `PostgresConnectionManager.open()`—it makes sense that `sslmode` is supported implicitly, and that we could document it in the Redshift profile as well as the Postgres one.

How exactly does `sslrootcert` figure in, though? I imagine that, in order to take effect, it would need to be included as a keyword argument to `psycopg2.connect()`, which isn't true for `dbt-postgres` today ([source](https://github.com/fishtown-analytics/dbt/blob/cb460a797c5cf73567fb42f17fba165e12facd80/plugins/postgres/dbt/adapters/postgres/connections.py#L78-L104)).

MBarendregt:
Yep you are right, I just found that out over the weekend as well :). Seemed to work because I baked it in /home/user/.postgresql/root.crt which will be the default path it will look (or ~/.postgresql/root.crt to be precise).

I updated the pr

jtcohen6 (approved):
Thanks for this @MBarendregt!

PR #535, opened 2021-01-26
Use relative links

clrcrl:
## Description & motivation
These are vestiges of either:
- a time when docs.getdbt.com pointed to the readme.io version of the site
- a time when our link checker worked a little differently for duplicate doc IDs, so the full version of the URL was required
- small mistakes

## Validation
I checked the links as I went and the all work (FAQ links work when both embedded, and on the `/faqs` page)

## Pre-release docs


drewbanin (approved):
Looks great - thanks for doing this! 🚢

PR #547, opened 2021-02-02
replace 'status' with 'response' where needed

### LINKED ISSUE #546: Rename class methods to implement in adapter plugins
The documentation (https://docs.getdbt.com/docs/contributing/building-a-new-adapter) mentions 4 methods to implement in a new adapter plugin:
- `open`
- `get_status`
- `cancel`
- `exception_handler`

Actually, the inline documentation in https://github.com/fishtown-analytics/dbt/blob/dev/margaret-mead/core/dbt/adapters/sql/connections.py lists the 4 following methods, existing as abstract in `SQLConnectionManager`:
- `exception_handler`
- `cancel`
- `get_response`
- `open`

From the code, it looks like `get_status` has just been renamed as `get_response` and the fix is mainly consisting in a s/status/response in the docs.

I can provide a PR linked to this issue if you agree with this conclusion.

gansanay:
Replace `status` with `response` where needed in the how to page for building a new adapter, following the change in abstract methods for `SQLConnectionManager`

Fixes #546

### DIFF
--- a/website/docs/docs/contributing/building-a-new-adapter.md
+++ b/website/docs/docs/contributing/building-a-new-adapter.md
@@ -155,7 +155,7 @@ For example:
 
 #### get_response(cls, cursor)
 
-get_response is a classmethod that gets a cursor object and returns the status of the last executed command. The return value is one of the values returned from `execute()` and is logged. You can just return 'OK' or similar if your database does not provide status.
+get_response is a classmethod that gets a cursor object and returns adapter-specific information about the last executed command. Ideally, the return value is an `AdapterResponse` object that includes items such as `code`, `rows_affected`, `bytes_processed`, and a summary `_message` for logging to stdout. Or, get_response can just return a string `'OK'` if your connection cursor does not provide richer metadata.
 
 <File name='connections.py'>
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/contributing/building-a-new-adapter.md
```diff
@@ -153,15 +153,15 @@ For example:
 
 </File>
 
-#### get_status(cls, cursor)
+#### get_response(cls, cursor)
 
-get_status is a classmethod that gets a cursor object and returns the status of the last executed command. The return value is one of the values returned from `execute()` and is logged. You can just return 'OK' or similar if your database does not provide status.
+get_response is a classmethod that gets a cursor object and returns the status of the last executed command. The return value is one of the values returned from `execute()` and is logged. You can just return 'OK' or similar if your database does not provide status.
```
jtcohen6: I realized there is a subtle distinction to make here:
```suggestion
get_response is a classmethod that gets a cursor object and returns adapter-specific information about the last executed command. Ideally, the return value is an `AdapterResponse` object that includes items such as `code`, `rows_affected`, `bytes_processed`, and a summary `_message` for logging to stdout. Or, get_response can just return a string `'OK'` if your connection cursor does not provide richer metadata.
```
The basic implementation below still works, so just a simple rename there is fine

jtcohen6 (commented):
Thanks for this @gansanay! One small recommendation, now that I've remembered that `get_status` &rarr;`get_response` was a simple rename _and_ a very slight functional expansion.

jtcohen6 (approved):
Thanks @gansanay!

PR #555, opened 2021-02-08
Use 'develop' branch for dbt repo links

jtcohen6:
## Description & motivation

We're trying a new branch management strategy in the dbt repo that includes a long-lived `develop` branch, which will persist beyond specific minor versions.

This PR updates all dbt repo links to `develop` branch and, as necessary, the linked line. Of course, there is a good chance that code in the `develop` branch will change, and the relevant line will move around. I'd rather folks be linked to the wrong line than be looking at outdated code. Barring significant refactoring work, the linked function or macro should still be searchable in the same file.

### DIFF
--- a/website/docs/docs/contributing/building-a-new-adapter.md
+++ b/website/docs/docs/contributing/building-a-new-adapter.md
@@ -15,7 +15,7 @@ This guide will walk you through the first two steps, and provide some resources
 
 ## Scaffolding a new adapter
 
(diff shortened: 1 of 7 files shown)


clrcrl:
I know there's other ones floating around, [example](https://github.com/fishtown-analytics/docs.getdbt.com/blame/current/website/docs/docs/guides/debugging-schema-names.md#L19). Can you quickly double check what you used as your search string? Other than that LGTM and i'm fully in support of this

jtcohen6:
Nice catch, looks like I missed `debugging-schema-names` and one other page. Just updated. I searched for `github.com/fishtown-analytics/dbt/blob/` in `website/docs/`. I didn't update links that reference changelogs of specific past versions.

clrcrl (approved):
TY!

PR #556, opened 2021-02-08
Add a package FAQ for SEO purposes

clrcrl:
## Description & motivation
Every so often we get people who ask this question, and haven't realized that packages are the answer. Hoping that adding this question as an FAQ means they can search their way to the right answer in the future!

## Pre-release docs


annafil (approved):
this shall be merged! 💪

PR #566, opened 2021-02-18
Add trailing slash to canonical links

lfracalossiFA:
## Description & motivation
Per [Research - Update canonical tags on docs.getdbt.com](https://www.notion.so/fishtownanalytics/Research-Update-canonical-tags-on-docs-getdbt-com-a69bd4cd99854e869045abe640868eaa), we wanted to make sure the canonical link tag is being written with a trailing slash.

I accomplished this by running the `swizzle` command, which creates a copy of a certain component which you can then modify. The modified version will take precedence over the original component. All I changed was appending a "/" to the end of the code that wrote the &lt;link rel="canonical" /&gt; tag.

## Pre-release docs
[x ] Not a pre-release - to be merged into `current`.

### DIFF
--- a/website/src/theme/DocItem/index.js
+++ b/website/src/theme/DocItem/index.js
@@ -1,0 +1,239 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+import Head from '@docusaurus/Head';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import DocPaginator from '@theme/DocPaginator';
+import useTOCHighlight from '@theme/hooks/useTOCHighlight';
+import Link from '@docusaurus/Link';
+
+import clsx from 'clsx';
+import styles from './styles.module.css';
+
+const LINK_CLASS_NAME = 'table-of-contents__link';
+const ACTIVE_LINK_CLASS_NAME = 'table-of-contents__link--active';
+const TOP_OFFSET = 100;
+
+function DocTOC({headings}) {
+  useTOCHighlight(LINK_CLASS_NAME, ACTIVE_LINK_CLASS_NAME, TOP_OFFSET);
+  return (
+    <div className="col col--3">
+      <div className={styles.tableOfContents}>
+        <Headings headings={headings} />
+      </div>
+    </div>
+  );
+}
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
+function Headings({headings, isChild}) {
+  if (!headings.length) {
+    return null;
+  }
+  return (
+    <ul
+      className={
+        isChild ? '' : 'table-of-contents table-of-contents__left-border'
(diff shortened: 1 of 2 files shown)


lfracalossiFA:
@krevitt I would love your feedback on the best ways to test this. I'm pretty sure that this is doing what we want, but you have a better understanding of the SEO implications than I do. 

@drewbanin I mostly just want to make sure I didn't royally mess anything up with editing the swizzled DocItem component!

clrcrl:
removing myself as reviewer as I don't know enough on this topic!

website/src/theme/DocItem/index.js
```diff
@@ -0,0 +1,239 @@
...
+        )}
+        {keywords && keywords.length && (
+          <meta name="keywords" content={keywords.join(',')} />
+        )}
+        {metaImage && <meta property="og:image" content={metaImageUrl} />}
+        {metaImage && <meta property="twitter:image" content={metaImageUrl} />}
+        {metaImage && (
+          <meta name="twitter:image:alt" content={`Image for ${title}`} />
+        )}
+        {permalink && <meta property="og:url" content={siteUrl + permalink} />}
```
krevitt, line 103: Should og:url match canonical here?  Only used by FB but may as well.

krevitt (commented):
[Recorded thoughts here](https://www.loom.com/share/13a60762d92748919b04ee88c2252349) - main question is, do we know what's causing the underlying redirect from https://docs.getdbt.com/dbt-cli/installation --> https://docs.getdbt.com/dbt-cli/installation/?

lfracalossiFA:
> [Recorded thoughts here](https://www.loom.com/share/13a60762d92748919b04ee88c2252349) - main question is, do we know what's causing the underlying redirect from https://docs.getdbt.com/dbt-cli/installation --> https://docs.getdbt.com/dbt-cli/installation/?

To summarize, for anyone joining the conversation late or who can't access the video -- the canonical URL may not be the right problem to fix here.

Clicking around on links throughout docs.getdbt.com, all the links seem to be written without the trailing slash. Which is correct.

However, you do see a redirect from "" to "/" under the following circumstances:
- When you refresh the page
- When you enter the URL in directly/manually
- If you're clicking in from search

... which is when the URL that's written in the address bar gets out of sync with the canonical URL.

@krevitt , you make a really good point -- we don't want our canonical URLs to be out of sync with what the links actually are on the site itself. So if there's anything to be fixed here, it's the redirect, not the canonical URL.

Do you feel like this redirect, as-is, is causing the problem David Khim laid our in the original ticket? i.e. is this even something we need to fix?

krevitt:
> @krevitt , you make a really good point -- we don't want our canonical URLs to be out of sync with what the links actually are on the site itself. So if there's anything to be fixed here, it's the redirect, not the canonical URL.
> 
> Do you feel like this redirect, as-is, is causing the problem David Khim laid our in the original ticket? i.e. is this even something we need to fix?

Yes, fair to say that the redirect is causing the problem that David Khim saw in the crawl - issue is we could be sending googlebot in circles (from non-trailing-slash -> redirect to / -> canonicalized to non-trailing-slash).  

Not a huge issue, but can be a negative search signal.

lfracalossiFA:
Closing in favor of #576.

PR #568, opened 2021-02-19
Add macros in macros faq.

### LINKED ISSUE #166: FAQ: "Can you call a macro within another macro"
I mean, technically not a dbt question, but we get it often enough that it might be worth writing it up

VersusFacit:
## Description & motivation
closes #166. Totally agree this should be documented in some way.


website/docs/faqs/calling-macros-in-macros.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+One warning, do not introduce an infinite cycle (i.e. two macros are defined in terms of each other).
+
+```sql
+
+-- this will cause a runtime exception
+{% macro macro_a(column_name) %}
+    macro_b('{{ column_name }}')
+{% endmacro %}
```
clrcrl, line 31: I'm a little confused by this — this doesn't look like valid jinja (regardless of the cyclic dependency)

Did you mean something more like this:

```suggestion
{% macro macro_a(column_name) %}
    {{ macro_b(column_name) }}
{% endmacro %}
```

website/docs/faqs/calling-macros-in-macros.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Can I call a macro inside a macro?
+---
+
+Yes. Macros can call other macros. Jinja will automatically seek out all macros defined in the project before attempting to deference a call to a macro. In other words, a macro that invokes another macro does not need to be defined in the same file or directory.
```
clrcrl, line 5:
```suggestion
Yes. Macros can call other macros, check out this example:
```

I'd add this after the first example:
```
Note: the referenced macro can be defined:
- In the same file
- In another file in the same project
- Or in an installed [package](package-management) (in which case, you'll need to prefix the macro's name with the package name)

clrcrl (changes requested):
I think I see people ask this for two reasons:
1. The example described here (where they just haven't tried it!)
2. They are trying to call a macro as an argument to another macro, and are getting hit by the ol' [nesting curlies](https://docs.getdbt.com/docs/building-a-dbt-project/dont-nest-your-curlies) problem

Let's try to cover both cases! That's on me though — I didn't scope the issue quickly, as I probably wrote it back in the one-person-team times

In general, I also encourage you to make the language less formal — as per our brand voice, we are happy using conjunctions, and I'd avoid using words that aren't commonly used ("deference" stuck out to me)

website/docs/faqs/calling-macros-in-macros.md
NicolasPA: I also had this question, I can confirm this suggestion works. @VersusFacit would you like to finish this PR, or should I create a new one?

runleonarun:
I am closing this PR and putting it in the backlog. Labeled with "icebox" so we can revisit at a later date.

PR #570, opened 2021-02-19
Spell out member / owner default behavior

matt-winkler:


website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +16,20 @@ dbt Cloud supports two different permission sets to manage permissions for self-
...
+We recommend the following steps be taken after onboarding Administrative users and before adding less priveleged users to a dbt Cloud Enterprise account.
+
+1) Click the Hamburger icon at the top left of the Cloud Console, then Account Settings >> Groups >> <Owner/Member>  
+2) Uncheck the "ADD USERS BY DEFAULT" option:  
+
+<Lightbox src="static/img/docs/dbt-cloud/dbt-cloud-enterprise/access-control/uncheck-add-users-by-default.png" title="Adjusting default group assignments"/>
+
+With this setting disabled, permission sets will apply transparently to additional groups configured in dbt Cloud.
+
+## Read Only vs. Developer License Types
```
amychen1776: I personally think this section should come up higher, before the enterprise section.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +16,20 @@ dbt Cloud supports two different permission sets to manage permissions for self-
 | Manage billing | ❌ | ✅ |
 | Manage team permissions | ❌ | ✅ |
 | Invite Owners to the account | ❌ | ✅ |
+
+## Member and Owner Groups in dbt Cloud Enterprise
+
+By default, new users are added to the Member and Owner groups when they onboard to dbt Cloud. This is necessary for account onboarding but can create confusion when initially setting up SSO and RBAC for dbt Cloud Enterprise accounts as described in the [Enterprise Permissions](https://docs.getdbt.com/docs/dbt-cloud/access-control/enterprise-permissions/) guide. Users may be added with more permissions than intended without adjusting this setting.
```
amychen1776: Could this link to the actual object (don't know what that's called but I think it should be `dbt-cloud/access-control/enterprise-permissions`)

amychen1776: >Users may be added with more permissions than intended without adjusting this setting.

The key issue with having these groups is that these groups override all the other permission sets because they're both account level permissions sets. I would emphasize that this needs to be removed because of the potential creation of conflicting permission sets.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +16,20 @@ dbt Cloud supports two different permission sets to manage permissions for self-
...
+
+We recommend the following steps be taken after onboarding Administrative users and before adding less priveleged users to a dbt Cloud Enterprise account.
```
amychen1776: Why not just have them delete the groups after SSO is setup?

matt-winkler: Some smaller teams might have an "everyone is an Owner" approach, which probably still isn't ideal. I'll add some additional text noting that they can delete the groups as well.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: Unclear which object you're referring to and how that's different from the link. Do you mean the diagram within the linked page?

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: Cool - swapped these so ##Read Only comes first.

matt-winkler: Moved.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: @amychen1776 I'm not following what you mean by the actual object and how that's different from providing a link to the relevant page. Can you please elaborate?

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: @amychen1776 Would bet that some smaller groups will have an "everyone's an Owner" approach, so was thinking they may want to disable instead of remove outright. Definitely worth noting they can remove the groups as well so will add that to the text.

amychen1776: Ah sorry Matt. I should have been more clear. They need to actually remove the groups outright after they have successfully created the Enterprise RBAC groups. They can't have the Owner and Member groups existing because it can cause weird bugs. The Owner permission set and the Account Admin permissions are completely the same so they need to delete the owner group and then create a new one with the Account Admin permission set. If they have a `everyone is an owner` mindset --> they still need to move everyone to the Account Admin permission set.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
amychen1776: Yeah, let me see if I can try to explain better with my limited knowledge of markdown. The better practice is to use a relative link rather than one that links directly to the production page. This way, when you host the website locally, your links still work rather than redirecting you to the production page. You want to link to the file in the repository.

amychen1776 (dismissed):

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: Got it! Thanks for adding that detail.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: Ah ok I see what you mean now! Thanks for describing that more. Changed this as well.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: @amychen1776 would it be possible to include this information in the Multi-Tenant Deployment Plan? 

image

Looks like I'm read only on this one. If you want to grant me edit access I can add as well.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +16,21 @@ dbt Cloud supports two different permission sets to manage permissions for self-
...
 | Manage team permissions | ❌ | ✅ |
 | Invite Owners to the account | ❌ | ✅ |
+
+## Read Only vs. Developer License Types
+
+Users configured with Read Only license types will experience a restricted set of permissions in dbt Cloud. If a user is associated with an member permission set and a Read Only seat license, then they will only have access to what a Read-Only seat allows. See [Seats and Users](cloud-seats-and-users) for more information on the impact of licenses on these permissions.
+
+## Owner and Member Groups in dbt Cloud Enterprise
+
+By default, new users are added to the Member and Owner groups when they onboard to dbt Cloud. This is necessary for account onboarding but can create confusion when initially setting up SSO and RBAC for dbt Cloud Enterprise accounts as described in the [Enterprise Permissions](enterprise-permissions) guide. Owner and Member groups are **account level** groups, so their permissions override any project-level permissions you wish to apply.
```
amychen1776: Could we mention that these are the default groups that come with a dbt cloud account? I've had folks ask me why these exist and they are literally created when you create a new account.

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +16,21 @@ dbt Cloud supports two different permission sets to manage permissions for self-
...
+
+We recommend the following steps be taken after onboarding Administrative users and before adding less priveleged users to a dbt Cloud Enterprise account.
```
amychen1776: Sorry to be nitpicky here but can we say:

> After onboarding Adimstrative users, complete the following steps after SSO and RBAC has been successfully created


We really don't want them to delete the groups before RBAC has been setup or else...they will have no permissions at all to set up RBAC. 😓

amychen1776: I should also note that we do mention deleting the Owner and Member group in the RBAC loom video. Perhaps we should also link to this section on the Enterprise permissions page.

matt-winkler: @amychen1776 Good thinking on spelling out that part of the process. Thoughts on adding another step to "add an Admins group for your privileged users" to make sure those folks can do the other group management things described below?

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: For sure! Added this verbiage

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
amychen1776: I like the update; do you think it would be better if we just spell out that they need to create an administrator group with the Account Admin permission set? I think there should be no account with no account admins on it so it makes to just say, this is what you need to do .

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
```diff
@@ -18,3 +18,36 @@ dbt Cloud supports two different permission sets to manage permissions for self-
 | Manage billing | ❌ | ✅ |
 | Manage team permissions | ❌ | ✅ |
 | Invite Owners to the account | ❌ | ✅ |
+
+## Read Only vs. Developer License Types
+
+Users configured with Read Only license types will experience a restricted set of permissions in dbt Cloud. If a user is associated with an member permission set and a Read Only seat license, then they will only have access to what a Read-Only seat allows. See [Seats and Users](cloud-seats-and-users) for more information on the impact of licenses on these permissions.
```
drewbanin:
```suggestion
Users configured with Read Only license types will experience a restricted set of permissions in dbt Cloud. If a user is associated with a _Member_ permission set and a Read Only seat license, then they will only have access to what a Read-Only seat allows. See [Seats and Users](cloud-seats-and-users) for more information on the impact of licenses on these permissions.
```

drewbanin (commented):
This is great @matt-winkler ! I dropped a couple minor copy-editing suggestions inline, but overall, this is great content :)

One thing that jumps out at me: this info is written in the "Self-Service Permissions" document. Is that the right home for this? Some of the steps enumerated in the new docs are enterprise-only, so wondering if there could be a better home for that information!

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: did this, good call!

website/docs/docs/dbt-cloud/access-control/self-service-permissions.md
matt-winkler: Thanks @drewbanin and nice catch on the indefinite article!

matt-winkler: @drewbanin @amychen1776 On whether this page is the right home for this, I could see this moving to either Access Control or Enterprise Permissions. But, wondering if the setup steps would fit better in a new "Enterprise SSO Account Onboarding" page? Don't have a super-strong opinion here.

drewbanin: ok - i similarly do not feel strongly about the home for this content. I think an Enterprise SSO Account Onboarding page might be a _great_ idea, but we can save that one for later :)

Call it how you see it, and if you do make a change, please feel free to tag me for re-review! If not, this LGTM as-is :)

drewbanin (approved):

matt-winkler (commented):
addressed these

PR #575, opened 2021-02-23
Add # to url_path example

one-data-cookie:
## Description & motivation
The `url_path` example was missing `#` at the beginning, even though it was correct in the `yml` example. It was clearly just a typo.


clrcrl (approved):
Thank you so much!

PR #577, opened 2021-02-23
Update broken link for schema test source code

ghost:
Fixing a link that was out of date due to `core` being in the directory tree now.


clrcrl (approved):
TY TY!

PR #578, opened 2021-02-24
Static insert_overwrite: partitions must match partition_by.data_type

jtcohen6:
Motivated by [this Slack thread](https://getdbt.slack.com/archives/C2JRRQDTL/p1614074541079900).

Now that BigQuery supports `partition by timestamp_trunc(...)`, dbt no longer wraps all `timestamp`-type partition columns in `date()`. (This is necessary to support hour-partitioned models.) Where previously users would always produce `date`-type `partitions` to filter on, `partitions` expressions must now return the same data type as `partition_by.data_type`.

IMO this makes _more_ sense than before, it's just a subtle detail that's easy to slip under the radar. The "static" `insert_overwrite` incremental strategy is also "full control" mode, where the user is expected to be thoughtful about what _exactly_ is being overwritten, because it promises to save the most $$$.

## Pre-release docs
- [ ] Yes: please

### DIFF
--- a/website/docs/reference/resource-configs/bigquery-configs.md
+++ b/website/docs/reference/resource-configs/bigquery-configs.md
@@ -443,14 +443,14 @@ will reduce costs by eliminating multiple queries in the model build script.
 
 #### Static partitions
 
-To supply a static list of partitions to overwrite, use the `partitions` configuration:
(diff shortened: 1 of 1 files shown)


clrcrl (approved):
LGTM!

PR #579, opened 2021-02-24
Adjustment for unique test for two columns.

jaypeedevlin:
## Description & motivation
The example provided generates the following line when compiled into the test:

```sql
 where country_code || '-' || order_id is not null
```

In Redshift (and possibly others), the `order_id is not null` gets evaluated first, and then the line errors out as it tries to concatenate a bool with text.  Adding parens solves this.


clrcrl (approved):
Ahh good catch!! (Sorry I missed this the other week)

PR #582, opened 2021-02-26
Update 2b-create-a-project-dbt-cli.md

natekupp:
`git init` still defaults to creating a `master` branch unless the user has previously run `git config --global init.defaultBranch main`; renaming the `master` branch to `main` is required, otherwise the `git push` will fail


clrcrl (approved):
TY!

PR #589, opened 2021-03-02
Add 'openid' scope to gcloud command

heisencoder:
When I run with the `gcloud auth` command as shown, I get this error:

```
ERROR: gcloud crashed (Warning): Scope has changed from "https://www.googleapis.com/auth/drive.readonly https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/accounts.reauth" to "https://www.googleapis.com/auth/accounts.reauth https://www.googleapis.com/auth/cloud-platform openid https://www.googleapis.com/auth/drive.readonly https://www.googleapis.com/auth/userinfo.email".
```

This error message shows that the `openid` scope needs to be added.  After adding this scope, the error message goes away.

See https://github.com/fishtown-analytics/dbt/issues/3040 for more context.

### DIFF
--- a/website/docs/reference/warehouse-profiles/bigquery-profile.md
+++ b/website/docs/reference/warehouse-profiles/bigquery-profile.md
@@ -314,10 +314,8 @@ To connect to BigQuery using the `oauth` method, follow these steps:
 
 ```shell
 gcloud auth application-default login \
-  --scopes=https://www.googleapis.com/auth/userinfo.email,\
-https://www.googleapis.com/auth/cloud-platform,\
-https://www.googleapis.com/auth/drive.readonly,\
-openid
+  --scopes=https://www.googleapis.com/auth/bigquery,\
+https://www.googleapis.com/auth/drive.readonly
 ```
 
 A browser window should open, and you should be promoted to log into your Google account. Once you've done that, dbt will use your oauth'd credentials to connect to BigQuery!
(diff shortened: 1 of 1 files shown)


jtcohen6 (commented):
Thanks for this @heisencoder!

Both commands (with and without `openid`) work for me locally. I'm guessing this has something to do with organizational settings for [OpenID Connect](https://developers.google.com/identity/protocols/oauth2/openid-connect)?

heisencoder:
Unfortunately, I don't know the reason why we need to add the `openid` scope, but do know that adding it (at least for me) fixed my error.

heisencoder:
I'm happy either adding this change or leaving it out if it only affects my particular setup.

jtcohen6:
@heisencoder Thanks for your patience as I try to figure out what's what!

Per the ongoing conversation in fishtown-analytics/dbt#2953, after some more testing on my end, I think it's possible that we could trim `userinfo.email` and `cloud-platform` from the required scopes in `application-default`. I wonder if you see the same error when you revoke all and instead run:
```
gcloud auth application-default login \
  --scopes=https://www.googleapis.com/auth/bigquery,\
https://www.googleapis.com/auth/drive.readonly
```
If you still see an error, fixed by the inclusion of `openid`, then we'll have confirmed that this is required for some users' gcloud authentication and we can definitely add it in.

heisencoder:
@jtcohen6 This sounds great!  I'm a big fan of narrowing the scopes!

I've run the gcloud command, and can confirm that by removing the userinfo.email scope, I no longer see the error with needing `openid`.  I was also able to successfully read from an external table backed by a Google Sheet using the Python client!

Thanks!
-Matt

heisencoder:
Based on the feedback, I updated the text and tested the changes locally to verify correct syntax.

jtcohen6 (approved):
Sweet! Let's give this a go here for now. If we're proven wrong, it's much easier to update the docs site than dbt-bigquery code.

HeddeCrisp:
Hi all, the above solutions did not work for me unfortunately. I still get the same error message. I am using dbt=0.20.1.
Any ideas on how to solve? I revoked my access and tried the recommended commands, but it did not work.

PR #595, opened 2021-03-09
Removing Beta banner for Selectors

saouderkirk:
Jerco sez it's not in beta, therefore, it's not in beta!

## Description & motivation

This doc says Selectors are in beta, and that is not true. 

## Pre-release docs

[x] Unsure: we'll let you know!


jtcohen6 (approved):
:shipit: 

The only thing that's not "fully there" about YAML selectors is that there still isn't a way to use them in the dbt-docs DAG viz (https://github.com/fishtown-analytics/dbt-docs/issues/119). We just haven't gotten a chance to do it yet.

PR #602, opened 2021-03-15
Fix a typo in the Jinja documentation

michaelecolley:
Update case statement in an example to remove ' from the jinja example because it's incorrect per how dbt operates and actually generates an error

## Pre-release docs
- [ ] Yes: please

## Checklist

### DIFF
--- a/website/docs/docs/building-a-dbt-project/jinja-macros.md
+++ b/website/docs/docs/building-a-dbt-project/jinja-macros.md
@@ -43,7 +43,7 @@ Here's an example of a dbt model that leverages Jinja:
 select
     order_id,
     {% for payment_method in payment_methods %}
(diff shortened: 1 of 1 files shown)


clrcrl:
Hi Michael! I appreciate the edit, however this is definitely _not_ my experience with this code sample (having taught the lesson that uses this sample many times over!). We need the quotes here, since the compiled SQL  needs to look like this:
```sql
sum(case when payment_method = 'bank_transfer' then amount end) as bank_transfer_amount
```
Removing the quotes would result in:
```sql
sum(case when payment_method = bank_transfer then amount end) as bank_transfer_amount
```
^ i.e. the `case when` statement would try to compare the value to another column `payment_method = bank_transfer`

PR #606, opened 2021-03-16
Improve pre-installation instructions

brylie:
Move the pre-installation instructions to before the installation instructions. Expand on requirements for Debian/Ubuntu.

### DIFF
--- a/website/docs/dbt-cli/installation.md
+++ b/website/docs/dbt-cli/installation.md
@@ -3,11 +3,19 @@ title: "Installation"
 id: "installation"
 ---
 
-We recommend you use install dbt using one of three tried and tested methods:
+We recommend you use install dbt using [homebrew](#homebrew) (recommended for MacOS), [pip](#pip), or by [installing from source](#install-from-source).
 
-- [homebrew](#homebrew) (recommended for MacOS)
+- [Pre-installation requirements](#pre-installation-requirements)
+  - [CentOS](#centos)
+  - [Debian/Ubuntu](#debianubuntu)
+  - [Windows](#windows)
+- [Homebrew](#homebrew)
+  - [Upgrading](#upgrading)
+  - [Installing different versions with Homebrew](#installing-different-versions-with-homebrew)
 - [pip](#pip)
-- [from source](#install-from-source)
+  - [Installation](#installation)
+  - [Upgrading](#upgrading-1)
+- [Install from source](#install-from-source)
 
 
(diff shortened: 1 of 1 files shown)


clrcrl:
@jtcohen6 — is anyone on your team well-placed to review this? It's all latin to me!

jtcohen6:
Hey @brylie, since each of these pre-installation requirements applies to a much smaller subset of users, I'd rather keep them further down on the page, and keep the more generally applicable information about brew, pypi, and building from source up top. That said, I think it would make a lot of sense to include a sentence near the top of the page to the effect of:
```
If you use CentOS, Debian/Ubuntu, or Windows, some [pre-installation setup](#pre-installation-requirements) is required.
```

brylie:
I understand that not all operating systems have the same amount of set-up. 

However, the document flows more naturally when the *prerequisites* are described first followed by the installation. Providing a table of contents makes it easier for the reader to skip to the relevant sections, if the prerequisites are not relevant.

clrcrl:
Since there's a table of contents in the right hand pane, I don't think we would want one at the top of the page as well. I agree with Jeremy here — even though the flow isn't linear, I'd prefer if this information is not at the top of the page. I think a hyperlink in the first paragraph is the right move

brylie:
I spent some amount of time (15-30 minutes) troubleshooting the dbt installation on a fresh version of Ubuntu 20.04, whereas a clearly organized installation document could have cut that time and *confusion* down significantly.

Every person who installs dbt will have some prerequisites that don't come pre-installed on the OS:
- MacOS: Homebrew and Python 3
- Windows: Git and Python 3
- Debian/Ubuntu: utilities related to detecting OS version and pip

It is conventional to put prerequisites before installation in technical documentation.

The left-side table of contents along with a friendly, intro message like "Hey friend, there are some prerequisites you need on your OS before installing dbt. If you think you've got the prerequisites covered, feel free to skip to the relevant installation section." would both assist readers who want to get straight to installation.

I have closed this pull-request as we seem to have artistic differences, but urge the dbt documentation team to reconsider the structure of this document.

brylie:
In addition to Git and Python 3, both Windows and Linux will fail to install dbt if `pkg-config` is not installed. 

Please test these installation instructions by observing someone who is not familiar with dbt follow the instructions on a fresh operating system. I believe there are some tacit steps and prerequisites that are missing from the documentation and have been inadvertently suppressed by core developers who are more familiar with dbt than new users.

PR #628, opened 2021-04-11
add Relation.include snippet

dataders:
## Description & motivation

AFAIK, [`BaseRelation.include()`](https://github.com/fishtown-analytics/dbt/blob/749f87397ec1e0a270b2e09bd8dbeb71862fdb81/core/dbt/adapters/base/relation.py#L121-L134) isn't currently documented, but it is very helpful!

I'm not sold on the example I used as the others are quite pithy. But `.include()` has more functionality that's worth addressing IMHO.


jtcohen6 (approved):
This is great! Thanks @swanderz

PR #641, opened 2021-04-21
Adding Google site verification

lfracalossiFA:
## Description & motivation
We are trying to add Google Search Console to docs, but first we have to verify that we own the site. This PR adds an HTML file to the root of /website/ that Google will use to verify our site.

(This worked locally, not entirely sure it will work in preview or production, so this PR is still draft).

## Pre-release docs
No

### DIFF
--- a/website/google18df7e5e81c2dcae.html
+++ b/website/google18df7e5e81c2dcae.html
@@ -1,0 +1,1 @@
+google-site-verification: google18df7e5e81c2dcae.html
--- a/website/google18df7e5e81c2dcae.html
+++ b/website/google18df7e5e81c2dcae.html
@@ -1,1 +1,0 @@
-google-site-verification: google18df7e5e81c2dcae.html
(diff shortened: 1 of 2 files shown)


lfracalossiFA:
Unfortunately if I try to go to https://deploy-preview-641--docs-getdbt-com.netlify.app/google18df7e5e81c2dcae.html, I get a 404 error. Don't think this approach is going to work...

lfracalossiFA:
Not 100% sure this method (of adding the verification text as a component) will work for Google's purposes, but there's at least something with the right text in the right place. The risk of releasing this is minimal, so we may just want to merge this and see if it works. @drewbanin ?

drewbanin (approved):
Let's try it!

lfracalossiFA:
I'm gonna revert this work, because I expect we don't want this stuff in prod since it's non-functional.

PR #650, opened 2021-04-27
Update tests.md

alieus:
fix for using > instead of < for negative amounts


clrcrl:
I think this is right, though perhaps not super clearly worded — it's saying that a $10 payment, plus a $5 refund (expressed as -5) should sum to an amount >= 0
```
10 + (-5) = 5
```

alieus:
Ahh, that makes sense! Thanks, @clrcrl

PR #661, opened 2021-05-06
Update require-dbt-version.md

mateuszklimek:
Correct name of option used in cli

## Description & motivation

Docs are missing `-` in name of option used in cli, PR is fixing that.

## Pre-release docs


clrcrl (approved):
Ooh good catch! Thanks for this :)

PR #689, opened 2021-06-11
Update style guide URL

mateuszklimek:
Update URL to point to new style guide

## Description & motivation

Current URL points to page which was moved.


annafil (approved):
Sharp 👀 s! Thanks so much for taking time to fix this!

PR #695, opened 2021-06-14
Add developer mode and long paths for Windows 10

barberscott:
## Description & motivation
Our requirements for Windows 10 development are a bit sparse.

- dbt deps by default will leverage symlinks and Windows 10 will report them as being supported, but unless the user is in Developer Mode or an Administrator, they do not have the privileges to create symlinks.  
- It possible to have files with a pathname+filename that exceeds 260 characters which is the default limit without enabling Long Paths -- this most often would affect tests where the compiled SQL filename could be quite long AND deep in the folder hierarchy.  This requires a configuration change on Windows 10 to support.
- We reference Python 3.5 but dbt itself requires Python 3.6 -- and Python 3.6 is ALSO the first version that supports long pathnames on Windows 10.
- We reference "Windows" generically when Windows 10 would be the only supported platform for these reasons.

## Pre-release docs


annafil:
Hello! :) @barberscott looks like we have a merge conflict -- someone else had [checked in some changes](https://github.com/fishtown-analytics/docs.getdbt.com/pull/695/conflicts) that overlap with yours re: Python 3.9. 

Would you mind taking a look and resolving?

barberscott:
Resolved merge changes -- lmk if we need to change anything else!

annafil (approved):
Sweet thank you!!🙏

PR #697, opened 2021-06-15
Modify explicit page title and description

annafil:
## Description & motivation

An astute [Community member has pointed out ](https://getdbt.slack.com/archives/C0VLNUUTZ/p1623688726351500) that we have inconsistent capitalizations for dbt on the docs site. We think this is because of Google Search taking liberties with auto-generating page structure. This PR attempts to address this by setting an explicit title that is not a URL. 

## To-do before merge
- [x] Test locally to make sure nothing breaks 

Screen Shot 2021-06-15 at 2 33 40 PM

## Pre-release docs
- [ ] Yes: please

## Checklist
no page additions/removal

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -55,8 +55,8 @@ console.log("DEBUG: ALGOLIA_INDEX_NAME = ", ALGOLIA_INDEX_NAME);
 module.exports = {
   baseUrl: '/',
   favicon: '/img/favicon.ico',
-  tagline: 'Your entire analytics engineering workflow',
-  title: 'docs.getdbt.com',
+  tagline: 'End user documentation, guides and technical reference for dbt (data build tool)',
+  title: 'dbt Docs',
   url: SITE_URL,
 
   themeConfig: {


website/docusaurus.config.js
```diff
@@ -55,8 +55,8 @@ console.log("DEBUG: ALGOLIA_INDEX_NAME = ", ALGOLIA_INDEX_NAME);
 module.exports = {
   baseUrl: '/',
   favicon: '/img/favicon.ico',
-  tagline: 'Your entire analytics engineering workflow',
-  title: 'docs.getdbt.com',
+  tagline: 'End user documentation, guides and technical reference for dbt (data build tool)',
```
jtcohen6, line 58: Do we even want `(data build tool)` here? We've stopped treating dbt as an initialism for this. (It still hangs out in a lot of old places, though)

jtcohen6 (approved):
Nice one! Just one comment from me

website/docusaurus.config.js
annafil: Good question. I think this tagline only shows up for SEO purposes and is a _suggestion_ for Google. I figured having the word data in there would be helpful but no strong opinion one way or the other.

jtcohen6: Ah that makes sense. SEO benefit probably outweighs pedantry

krevitt:
FYI - doesn't look from the source of [docs.getdbt.com](https://docs.getdbt.com/) like `tagline` compiles to a `meta:description` or `og:description` (seeing no description fields atm).  

On individual docs pages in the frontmatter it looks like it's the `description` field populates those.

annafil:
💯  that pages use the description field to populate the tag (although we aren't super good at setting these right now 🙃  ) 

According to the v1 Docusaurus docs (! I couldn't find the equivalent v2 docs page) tagline _should_ provide the default that then gets superseded by the page description if it's available. But we can also set [something at the React level](https://v1.docusaurus.io/docs/en/api-pages#description-for-pages) if we really want to be sure :)

clrcrl:
heads up that the reason the page title was set to `docs.getdbt.com` was to differentiate between tabs that are the website generated by "dbt docs generate" (left), and this page (right). It wasn't an intentional choice! (As was the old teal dbt icon for docs.getdbt.com

Can someone (@jasnonaz ?) think through how to differentiate these two tabs?

Screen Shot 2021-10-16 at 1 12 06 PM
m)

jasnonaz:
Thanks @clrcrl! @grantwinship-consulting had a [great idea](https://github.com/dbt-labs/dbt-docs/issues/219) that could help us distinguish here - interested to hear your thoughts.

PR #702, opened 2021-06-21
Update cloud-enabling-continuous-integration-with-github.md

alexbethea22:
Adding Github verbiage into the Troubleshooting section so folks know that this method works for **both** Gitlab and Github!


eogilvy12 (approved):
LGTM!

Alex I can't remember where we landing with linking to this with userflow during the onboarding process or from the job set up? may be worth some time to further investigate!

PR #705, opened 2021-06-23
Release v0.20

jtcohen6:
Opening now so that we can start reviewing the many docs changes associated with v0.20.0, currently living in `next.docs.getdbt.com`. All changes should be called out in / linked from the "Upgrading to 0.20.0" Migration Guide. Current plan is to have v0.20.0-final end of next week or early the following week.

@annafil @drewbanin I know it's a good chunk of changes. What makes most sense here in terms of review?

## To-do before merge
- Wait for v0.20.0 final release


drewbanin:
@jtcohen6 FYI that I can give this a look later today!

jtcohen6:
I'm going to merge these in now to coordinate with the v0.20.0 (final) release earlier this afternoon.

@annafil @drewbanin If you get a chance to check out the diffs, we can always change after the fact.

PR #710, opened 2021-06-28
Select Source Freshness

barryaron:
## Description & motivation
Changes include:
- making it clear that you need to ALWAYS add a freshness block to your source and loaded_at_field
- talking about new source freshness checkbox functionality 
- talking about how to run a job that only optimizes for source freshness

## To-do before merge
Should be fine, base is current.  I only pushed the files that have changes.

## Pre-release docs

## Checklist
Did not add new pages (no need for website/sidebars.js update), but did add new image with unique filename


ashnm (approved):
LGTM 🚀

PR #730, opened 2021-07-19
Set correct end for test block

davidwitk:
## Description & motivation
Corrected typo


jtcohen6 (approved):
Thanks for this @chodera!

PR #796, opened 2021-08-31
Rebrand

JKarlavige:
## Pre-release docs

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -205,5 +205,7 @@ module.exports = {
     '/css/search.css',
     '/css/api.css',
     'https://cdn.jsdelivr.net/npm/featherlight@1.7.14/release/featherlight.min.css',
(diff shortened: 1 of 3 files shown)


lfracalossiFA (changes requested):
One small thing Jason and I discussed -- it seems like the viewport meta is only one part of the problem with the styling of the inner pages. The other is the lightbox images, which are surrounded by a div with the style of `width: 600px` on them. This will cause those images to break out of the container at mobile device widths, causing the page to look like the header and body are shrunk (relative to the image).

We decided to add a `max-width: 100%` to `.docImage_2V3X` to fix this.

JKarlavige:
Noticed some images with the .docImage_src-components-lightbox- container class. So I added that along with .docImage_2V3X for the max-width

lfracalossiFA (approved):
lgtm!

PR #804, opened 2021-09-07
adds hsts support for docs subdomain

JKarlavige:
## Description & motivation
Adds HSTS header for docs subdomain to redirect all requests to HTTPs

## Pre-release docs

### DIFF
--- a/_headers
+++ b/_headers
@@ -1,2 +1,3 @@
 /*
   X-robots-tag: noindex
+  Strict-Transport-Security: max-age=63072000; includeSubDomains; preload


lfracalossiFA (changes requested):
Hi @JKarlavige -- correct me if I'm wrong, but I believe we don't want/can't use the includeSubDomains and preload parameters, since all our subdomains are different Netlify sites/hosts. I did not implement them when I did this for www, for example. 

(ETA: though I may be wildly misinterpreting [the docs](https://infosec.mozilla.org/guidelines/web_security#http-strict-transport-security) -- let me know what you think

lfracalossiFA:
(also please go ahead and add @emilyemcqueendbt as reviewer)

JKarlavige:
> Hi @JKarlavige -- correct me if I'm wrong, but I believe we don't want/can't use the includeSubDomains and preload parameters, since all our subdomains are different Netlify sites/hosts. I did not implement them when I did this for www, for example.
> 
> (ETA: though I may be wildly misinterpreting [the docs](https://infosec.mozilla.org/guidelines/web_security#http-strict-transport-security) -- let me know what you think

From what I've read, adding includeSubDomains and preload to a specific subdomain will only affect that subdomain, as this is only added to the docs site header in this case. This is based off this q&a [here](https://serverfault.com/questions/482350/can-i-turn-on-hsts-for-1-subdomain).

For example, if there was ever a subdomain added to docs, such as example.docs.getdbt.com, that example subdomain would receive the HSTS header. 

If we feel it is best to include the HSTS header strictly for docs, and not any other subdomains within docs, I can remove that no problem!

lfracalossiFA:
Ah, okay, that makes sense -- so we're kind of talking about sub-subdomains, lol. I do think we have one of those, next.docs.getdbt.com, so if we add this header, it will apply to that. But I don't think that will be an issue.

lfracalossiFA (approved):

emilyemcqueendbt (commented):
looks good to me!

PR #814, opened 2021-09-13
Setting up auto add of issues

annafil:
## Description & motivation
Take two of the actions script, removing the gh cli requirement


VersusFacit (approved):

jasnonaz (approved):
This looks great to me! I'm not super familiar with some of the complexities of the GH API invoked here but it all checks out on a conceptual level - very, very cool thank you for putting this together.

PR #818, opened 2021-09-14
Add script for backfilling issues

annafil:
## Description & motivation
Some GraphQL and bash goodness to take all open issues in a repo and send them to a Project board Triage column automagically.


runleonarun:
I'm going to close because I think we don't need this anymore. Feel free to re-open, Anna if you think we could use it!

PR #819, opened 2021-09-15
Fix Action Syntax

annafil:
## Description & motivation

Action wasn't running because uses was unexpected in that part of the script.


jasnonaz (approved):
lgtm!

PR #824, opened 2021-09-19
Update custom-generic-tests.md

clrcrl:
## Description & motivation
`%` doesn't work on BQ, but `mod()` works on all core warehouses.

I also removed a ton of placeholder SQL that I feel gets in the way of what the actual test does.

## Pre-release docs

### DIFF
--- a/website/docs/docs/guides/custom-generic-tests.md
+++ b/website/docs/docs/guides/custom-generic-tests.md
@@ -28,28 +28,11 @@ Here's an example of an `is_even` schema test that uses both arguments:
 ```sql
 {% test is_even(model, column_name) %}
 
-with validation as (
-
     select
         {{ column_name }} as even_field
 
     from {{ model }}
-
-),
-
-validation_errors as (
-
-    select
-        even_field
-
-    from validation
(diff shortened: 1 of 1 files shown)


website/docs/docs/guides/custom-generic-tests.md
```diff
@@ -28,28 +28,11 @@ Here's an example of an `is_even` schema test that uses both arguments:
...
-
-    from validation
-    -- if this is true, then even_field is actually odd!
-    where (even_field % 2) = 1
-
-)
-
-select *
-from validation_errors
+    where not(mod(even_field, 2) = 0)
```
clrcrl, line 35: I tend to prefer this `not` syntax when teaching, since it's easier to say "well I expect them all to be even, so give me the records that are `not` meeting that expectation"

clrcrl: oh wait, no never mind, this might not work in BQ since you can't use lateral column aliasing for `even_field`. I think this should be:
```suggestion
    where not(mod({{ column_name }}, 2) = 0)
```
which might be more confusing. ¯\_(ツ)_/¯ 

going to close for now

PR #825, opened 2021-09-20
[Release] v0.21.0

jtcohen6:
We released `v0.21.0-rc1` today. If all goes well, we'll be releasing v0.21.0 (final) in 2 weeks (Oct 4).

- [x] Wait for v0.21.0 final release

(Note: I very quickly enabled force pushes in the `next` branch, while I cleaned up commits for an easier review. Full branch protection is back on, including the prohibition of force pushes.)

PR #831, opened 2021-09-27
[Prerelease] v0.21 post-RC updates

jtcohen6:
Also planning to resolve PR feedback from https://github.com/dbt-labs/docs.getdbt.com/pull/825 here, so as to avoid pushing straight to `next`


jasnonaz (approved):
All looking good!

PR #838, opened 2021-10-04
updated docs for env var naming convention

schottj:
## Description & motivation
Editing documentation to account for environment variable naming conventions.


barryaron (approved):
let's go!!

PR #853, opened 2021-10-08
Result Selectors for Smarter Reruns for Failures/Errors

sungchun12:
## Description & motivation

Documentation for a brand new dbt Core feature!

## To-do before merge

- [x] Ensure PR [#4017](https://github.com/dbt-labs/dbt/pull/4017) is merged

## Pre-release docs
- [x] Yes: please


website/docs/docs/guides/best-practices.md
```diff
@@ -121,6 +121,43 @@ dbt run -s state:modified+ --defer --state path/to/prod/artifacts
...
+  - Rerun all my erroneous models AND run changes I made concurrently that may relate to the erroneous models for downstream use
+  - There's a failed test that's unrelated to modified or error nodes(think: source test that needs to refresh a data load in order to pass)
+
+```bash
+dbt test --select result:fail --exclude <example test> --defer --state path/to/prod/artifacts
+```
+  - Rerun all my failed tests and exclude tests that I know will still fail
+  - This can apply to updates in source data during the "EL" process that need to be rerun after they are refreshed
+
+> Note: `result:error` and `result:fail` flags can only be selected concurrently(in the same command) if using the `dbt build` command because `dbt test` will overwrite the `run_results.json` from `dbt run` in a previous command invocation.
```
jtcohen6: This is only true if you're using `--state target/`. Since the examples above have you referencing `path/to/prod/artifacts`, previous dbt commands won't be overwriting the state artifact.

That said, I really agree that `dbt build` is the way to go here! Makes it much easier

website/docs/docs/guides/understanding-state.md
```diff
@@ -32,3 +33,27 @@ If both the flag and env var are provided, the flag takes precedence.
...
+The following dbt commands produce `run_results.json` artifacts whose results can be referenced in subsequent dbt invocations:  
+- `dbt run`
+- `dbt test`
+- `dbt build` (new in dbt version v0.21.0)
+- `dbt seed` 
+
+After issuing one of the above commands, you can reference the results by adding a selector to a subsequent command as follows: 
+
+```bash
+$ dbt run --select result:<status>
```
jtcohen6: Note that you'll need to include either the `--state` flag here, or to have set the `DBT_ARTIFACT_STATE_PATH` environment variable

website/docs/reference/node-selection/methods.md
```diff
@@ -126,3 +126,15 @@ $ dbt run --select +exposure:weekly_kpis                # run all models that fe
 $ dbt test --select +exposure:*                         # test all resources upstream of all exposures
 $ dbt ls --select +exposure:* --resource-type source    # list all sources upstream of all exposures
 ```
+
+### The "result" method
+<Changelog>New in v1.0.0</Changelog>
+
+The `result` method is related to the `state` method described above, and can be used to select resources based on their result state from a prior run. Note that one of the dbt commands [`run`, `test`, `build`, `seed`] must have been performed in order to create the result on which a result selector operates. You can use `result` selectors in conjunction with the `+` operator.
```
jtcohen6: The words here, as everywhere, can be confusing/overlapping
```suggestion
The `result` method is related to the `state` method described above, and can be used to select resources based on their result status from a prior run. Note that one of the dbt commands [`run`, `test`, `build`, `seed`] must have been performed in order to create the result on which a result selector operates. You can use `result` selectors in conjunction with the `+` operator.
```

website/docs/reference/node-selection/methods.md
```diff
@@ -126,3 +126,15 @@ $ dbt run --select +exposure:weekly_kpis                # run all models that fe
...
+
+```bash
+$ dbt run --select result:error # run all models that generated errors on the prior invocation of dbt run
+$ dbt test --select result:fail # run all tests that failed on the prior invocation of dbt test
+$ dbt build --select result:fail # run all the models associated with failed tests from the prior invocation of dbt build
```
jtcohen6: I think this might need to be:
```suggestion
$ dbt build --select 1+result:fail # run all the models associated with failed tests from the prior invocation of dbt build
```

jtcohen6 (commented):
Thanks for the excellent contribution @sungchun12! These are looking really good.

A few small things I caught in review. I'm also thinking that we should first confirm that this works with existing deferral features in dbt Cloud. It should be possible to test now using scheduled jobs + the `development (HEAD)` version of dbt Core.

website/docs/docs/guides/understanding-state.md
sungchun12: @matt-winkler can you update this part?

website/docs/docs/guides/best-practices.md
sungchun12: Updated!

sungchun12:
@jtcohen6 @matt-winkler 

I tried testing this out in dbt Cloud to no success. git repo and branch: https://github.com/dbt-labs/snowflake_dbt_demo_project/tree/demo-sung

**Is my `dbt_project.yml` file wrong in any way?**

**Does it work in the IDE? No.** It has a compilation error that may allude to there being a wrong config in `dbt_project.yml`. I don't see any issues with it though: [here](https://github.com/dbt-labs/snowflake_dbt_demo_project/blob/demo-sung/dbt_project.yml#L7)
image

**Job settings**
image


**Does it work for scheduled jobs? No**
image

**I ran the same job on `0.21.0` and it did not return the same error**
image


I also noticed when running locally that dbt Cloud isn't pulling the latest changes from the `main` branch. The above screenshots say `1.0.0-b1` when it should be `1.0.0-b2`
image

jtcohen6:
@sungchun12 Hmm. I'm thinking this is a result of the ongoing work to get v1.0.0 supported in dbt Cloud, which is requiring a heavier lift than usual, given the packaging changes we've made for `dbt-rpc` (https://github.com/dbt-labs/dbt-cloud/pull/4392).

I'm not sure why that error is appearing as `Invalid config version: 1, expected 2`.

Once we have a v1 beta confirmed working in Cloud, we should loop back and ensure that this feature is working in Cloud as well (with the right deferral constructs). In the meantime, could we split the Cloud statements into a separate PR, and get the core docs merged in?

website/docs/docs/guides/understanding-state.md
sungchun12: Matt's on vacation, I'll take care of this!

sungchun12: Updated this!

jtcohen6 (approved):
thanks for this! lgtm :)

PR #862, opened 2021-10-14
Update Docusaurus Version

JKarlavige:
## Description & motivation
[Asana Task](https://app.asana.com/0/1200099998847559/1201136871923005/f)

This is an effort to upgrade Docusaurus to the latest version (v2.0.0-beta.6) in order to utilize newer features on the platform.

Notes on a couple changes and new features:
- The new version has a new Algolia search, which functions a bit differently than the current search. This is visually different on both mobile and desktop as well. 
- There is also broken link checkers, which causes the build to fail by default. Since we have a 'bad-link' example on the /styles page, I changed this check to only warn of broken links. This allows the build to succeed. Using this link check, I updated/fixed all broken links which were flagged.
- The new version has some updated styles as well. I made adjustments to mirror our current site as much as possible, but some slight differences may be noticeable, for example with element padding/breakpoints/etc.
- Now that we will be on the latest version, it may be a good time to review the docs again to see what new features are available. A couple examples include the live coding component, as well as the ability to add a title to code examples (which we currently use a custom component for.)
- A number of duplicate slug warnings occur as well, both on this version and the live site. I will let these remain as is due to not knowing which version of the url should be used.

A side note:
The new Docusaurus requires a Node Version of > 12. This site was originally built on Node v10. I had to add a NODE_VERSION env variable to update the node version for this site. Just want to note that this is when that env variable was set.

## Pre-release docs

[Link to preview](https://deploy-preview-862--docs-getdbt-com.netlify.app/)


website/src/css/custom.css
```diff
@@ -437,3 +441,58 @@ nav.navbar div.navbar__inner {
...
+  padding: 0 2.25rem 0 0;
+}
+
+.alert {
+  color: var(--ifm-alert-color);
+}
+
+.alert a {
+  color: var(--ifm-alert-color);
+  text-decoration: underline;
```
krevitt, line 466: Looks like this `underline` text-decoration on `.alert a` isn't being applied in the deploy preview ([looking at this page](https://deploy-preview-862--docs-getdbt-com.netlify.app/docs/dbt-cloud/cloud-quickstart)):

image

image

krevitt (changes requested):
Did a styling review and all looked good, outside of one link styling issue noted above.

website/src/css/custom.css
JKarlavige: Fix for this added

lfracalossiFA (approved):

krevitt (approved):

PR #865, opened 2021-10-15
remove reference to deprecated heap package

jasnonaz:
## Description & motivation
Removed a reference to the Heap package which is no longer available on Hub


runleonarun (approved):
LGTM

PR #917, opened 2021-11-17
Site optimization to improve Pagespeed Insights scores

JKarlavige:
## Description & motivation
This optimizes certain areas of the site to improve page speed scores. This includes deferring scripts, reducing css loading, and image optimization.

## Pre-release docs

A majority of the files changed are optimized images. The other main focus was to reduce render-blocking resources. The primary blockers are the jQuery script and featherlight lightbox script. 

It looks like the only reason jQuery is added is to support the jQuery-dependent featherlight package. We can move to a react-based lightbox and remove jQuery altogether, but deferring these scripts should improve the page speed scores sufficiently enough without having to overhaul this.

[Here](https://deploy-preview-917--docs-getdbt-com.netlify.app/dbt-cli/cli-overview) is a page to verify the lightbox functionality still works after deferring both scripts. Clicking the first image will trigger the lightbox.

I also moved the styling for the lightbox inline to reduce the number of separate css files being loaded.


lfracalossiFA (approved):
This looks good to me!

You optimized a bunch of images, which is great! I'm curious if there's anything we can do moving forward to make sure that images we add to the site in the future are also optimized. I believe they're basically just added by contributors via the standard git workflow, correct? So we have no way to enforce that optimization?

Not a show-stopper in the least; it's just something I was curious about.

JKarlavige:
> This looks good to me!
> 
> You optimized a bunch of images, which is great! I'm curious if there's anything we can do moving forward to make sure that images we add to the site in the future are also optimized. I believe they're basically just added by contributors via the standard git workflow, correct? So we have no way to enforce that optimization?
> 
> Not a show-stopper in the least; it's just something I was curious about.

I like that idea. It looks like there's actually a [GitHub Action](https://css-tricks.com/optimize-images-with-a-github-action/) which makes that possible 👀 

Not sure if there is any risk in running a type of action like this, if it would always be accurate, etc. But maybe this would be worth testing on a demo repo to see how it performs!

lfracalossiFA:
I like this idea; let's do it!

PR #944, opened 2021-12-06
add project maturity blog

jasnonaz:


krevitt (approved):
Added a blockquote for the opening quote, and turned the mid-section H3s into bold headers so as not to clog up the right ToC sidebar with them.

PR #957, opened 2021-12-10
Update Coalesce CTA to 'watch the talks'

krevitt:
## Description & motivation
Just updating CTA copy on the blog, now that Coalesce is wrapped.


runleonarun (approved):
Looks good!

PR #967, opened 2021-12-15
Update DocSearch api key and add appId, move keys to env

JKarlavige:
## Description & motivation

DocSearch is migrating our account from a config repo to a new Algolia account. This adds the appId and moves new keys to an env file, rather than having them hard-coded.

## Pre-release docs

## Preview Link
https://deploy-preview-967--docs-getdbt-com.netlify.app/

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -1,5 +1,6 @@
 
 const path = require('path');
+require('dotenv').config()
 
 /* Debugging */
 var SITE_URL;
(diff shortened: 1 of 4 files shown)


website/docusaurus.config.js
```diff
@@ -31,19 +32,7 @@ if (!PRERELEASE) {
...
-  ALGOLIA_API_KEY = process.env.ALGOLIA_API_KEY;
-}
-
-var ALGOLIA_INDEX_NAME;
-if (!process.env.ALGOLIA_INDEX_NAME) {
-  ALGOLIA_INDEX_NAME = 'dbt';
-} else {
-  ALGOLIA_INDEX_NAME = process.env.ALGOLIA_INDEX_NAME;
-}
+let { ALGOLIA_APP_ID, ALGOLIA_API_KEY, ALGOLIA_INDEX_NAME } = process.env;
```
drewbanin, line 35: this looks great! Just wondering - do these env vars get passed in from Netlify in the staging/prod envs?

drewbanin (approved):
LGTM! Thanks for making this change!

website/docusaurus.config.js
JKarlavige: > this looks great! Just wondering - do these env vars get passed in from Netlify in the staging/prod envs?

Yep! It pulls them in with the same `process.env`. I kept the debug console.log in place for the ALGOLIA_INDEX_NAME to verify this within the Netlify build logs. On our last preview build log it has `9:59:53 AM: DEBUG: ALGOLIA_INDEX_NAME =  dbt`. This would be undefined if it failed to grab the Netlify env variable.

kwigley:
Was about to write up an issue, but I think this will resolve it! 👍 
image (2)

website/docusaurus.config.js
JKarlavige: > Was about to write up an issue, but I think this will resolve it! 👍

@kwigley The build just completed, looks like search is now working on my end. Can you verify you're no longer seeing the errors on your end?

kwigley: @JKarlavige no errors, looks great! 🎉

PR #977, opened 2021-12-19
Clarify upgrade instructions

jasnonaz:
## Description & motivation
Currently if you run dbt --version on the cli, you are directed to a url [docs.getdbt.com/docs/installation](docs.getdbt.com/docs/installation) which redirects to [https://docs.getdbt.com/dbt-cli/install/overview](https://docs.getdbt.com/dbt-cli/install/overview) and tells you that you can receive information about upgrading at that page. The issue is right now, this page is purely dedicated to installing dbt and doesn't give any information on how to upgrade. This PR adds links to some of the upgrading resources we have so that we can direct people who come here expecting upgrade information to the right place.
image

### DIFF
--- a/website/docs/dbt-cli/install/overview.md
+++ b/website/docs/dbt-cli/install/overview.md
@@ -15,3 +15,7 @@ You can install dbt Core on the command line by using one of these recommended m
 ## About dbt adapters
 
 dbt works with a number of different databases, query engines, and other SQL-speaking technologies. It does this by using a dedicated _adapter_ for each. When you install dbt, you'll also want to install the specific adapter for your database. For more details, see the [list of available adapters](available-adapters).
+
+## Updating dbt
+
+To upgrade via `pip`, see the [pip upgrading instructions](pip#upgrading). dbt Labs provides a number of resources for understanding general best practices while upgrading your dbt project as well as detailed [migration guides](https://docs.getdbt.com/docs/guides/migration-guide/upgrading-to-1-0-0) highlighting the changes required for each minor and major release.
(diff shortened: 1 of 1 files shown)


website/docs/dbt-cli/install/overview.md
```diff
@@ -15,3 +15,7 @@ You can install dbt Core on the command line by using one of these recommended m
 ## About dbt adapters
 
 dbt works with a number of different databases, query engines, and other SQL-speaking technologies. It does this by using a dedicated _adapter_ for each. When you install dbt, you'll also want to install the specific adapter for your database. For more details, see the [list of available adapters](available-adapters).
+
+## Upgrading dbt
```
runleonarun, line 19: @jasnonaz I love the idea of adding an upgrade section. I think it's worthy of being its own page even if we only have pip instructions right now.

runleonarun (changes requested):

nghi-ly:
@jasnonaz : so sorry for the delay! to address our PR backlog better, the Docs team is going to dedicate time on a regular cadence to work on it. in the spirit of this, i opened this PR to see what I can do to help it along. 

looks like you created this PR in Dec 2021. is this change still needed? if it's no longer needed, we can close it.

website/docs/dbt-cli/install/overview.md
runleonarun: Actually, I this is valuable information and we should get it added.

runleonarun:
Moved changes over to https://github.com/dbt-labs/docs.getdbt.com/pull/2759 instead of fixing merge conflicts.

PR #984, opened 2021-12-22
Clarify description of --store-failures config

Elliot2718:
## Description & motivation
This clarifies the documentation for the --store-failures configuration. My understanding is that the `--store-failures` option means that the test results **will** be stored, but the sentence as currently written implies the opposite.

## Pre-release docs
No, based on current.


runleonarun:
This was fixed in another PR! Thank you so much for contributing to the dbt product docs!

PR #1009, opened 2022-01-12
snowflake role - from optional to mandatory

mirnawong1:
changed the 'optional' to 'mandatory' in teh snowflake role field. this is because we always get users reaching out with an access error (see https://dbtcloud.zendesk.com/agent/tickets/21160 and https://getdbt.slack.com/archives/CBSQTAPLG/p1631852227241700) so i think we should make this clearer

## Pre-release docs
- [ ] Yes: please

## Checklist


amychen1776:
If we change this to mandatory, we will need to change the connection setup form to say so as well. It is technically optional but I think the issue is when a user has multiple roles that are associated with their Snowflake user

dataders (approved):
better late than never!

mirnawong1:
woop woop thank you @dataders !

PR #1018, opened 2022-01-13
load time on task blog

jasnonaz:
Load Time on Task Blog


krevitt (changes requested):
Love this post!  

Content + graphics are great, couple things I'd challenge us to improve here:

1) Scannability - some paragraphs are long / dense and there aren't many headers to break up the piece / make it navigable.

In particular, we jump in the intro right into how we're solving the problem (nested macros / calculation via subquery).  

Could we (in a similarly scannable way, with a header + bullets) introduce the problem itself (ignoring weekends + holidays)?  A quick google search for these topics (ignore weekends SQL, ignore holidays SQL, etc) will help us find how people are looking for answers on these problems, and thus how we'd want to frame them in headers / bullets.

2) Is "time on task" the right title framing here?  A google search returns education-focused content ("Increasing Time-on-Task and Student Engagement") - is there something else that people call this at work?

jasnonaz:
@krevitt ready for another look. Added and reworked the intro to make it clearer what problem we're solving and added more scannability. I agree that measuring business hours is a very common way that people phrase this _problem,_ but I do think Time on Task is an elegant way to refer to the _solution_, Added more discussion and context around the problem which should hopefully help clear this up.

krevitt (approved):

krevitt:
@jasnonaz pushed a commit with some refinement to headers + the intro, otherwise looks great!

there also were some unintentional code formatting on paragraphs in the middle, removed that.

PR #1027, opened 2022-01-18
Added Teradata adapter documentation

adamtwo:
## Description & motivation

Added Teradata adapter documentation

## Pre-release docs
- [ ] Yes: please

## Checklist

### DIFF
--- a/website/docs/reference/resource-configs/teradata-configs.md
+++ b/website/docs/reference/resource-configs/teradata-configs.md
@@ -161,6 +161,10 @@ id: "teradata-configs"
 
   Examples:
 
+  :::info Separators between statements
+  Note, unlike with `table_option` statements, there are no commas between statements in `index` config.
+  :::
+
   ```yaml
   {{
     config(
@@ -169,10 +173,6 @@ id: "teradata-configs"
     )
   }}
   ```
-
-  :::info Separators between statements
-  Note, unlike with `table_option` statements, there are no commas between statements in `index` config.
-  :::
 
   ```yaml
   {{
@@ -186,10 +186,6 @@ id: "teradata-configs"
     )
   }}
   ```
-
-  :::info Separators between statements
-  Note, unlike with `table_option` statements, there are no commas between statements in `index` config.
-  :::
 
   ```yaml
   {{
--- a/website/docs/reference/resource-configs/teradata-configs.md
+++ b/website/docs/reference/resource-configs/teradata-configs.md
@@ -86,34 +86,37 @@ seeds:
       WITH [NO] [CONCURRENT] ISOLATED LOADING [ FOR { ALL | INSERT | NONE } ]
       ```
 
-    Examples:
-    ```yaml
-    {{
-      config(
-          materialized="table",
-          table_option="NO FALLBACK"
-      )
-    }}
-    ```
-    ```yaml
-    {{
-      config(
-          materialized="table",
-          table_option="NO FALLBACK, NO JOURNAL"
-      )
-    }}
-    ```
-    ```yaml
-    {{
-      config(
-          materialized="table",
-          table_option="NO FALLBACK, NO JOURNAL, CHECKSUM = ON,
-            NO MERGEBLOCKRATIO,
-            WITH CONCURRENT ISOLATED LOADING FOR ALL"
-      )
-    }}
-    ```
+  Examples:
+
+  ```yaml
+  {{
+    config(
+        materialized="table",
+        table_option="NO FALLBACK"
+    )
+  }}
+  ```
+  ```yaml
+  {{
+    config(
+        materialized="table",
+        table_option="NO FALLBACK, NO JOURNAL"
+    )
+  }}
+  ```
+  ```yaml
+  {{
+    config(
+        materialized="table",
+        table_option="NO FALLBACK, NO JOURNAL, CHECKSUM = ON,
+          NO MERGEBLOCKRATIO,
+          WITH CONCURRENT ISOLATED LOADING FOR ALL"
+    )
+  }}
+  ```
+
   For details, see [CREATE TABLE documentation](https://docs.teradata.com/r/76g1CuvvQlYBjb2WPIuk3g/B6Js16DRQVwPDjgJ8rz7hg).
+
 * `with_statistics` - should statistics be copied from the base table, e.g.:
     ```yaml
     {{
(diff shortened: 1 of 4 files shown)


website/docs/reference/resource-configs/teradata-configs.md
```diff
@@ -0,0 +1,212 @@
...
+      # option 551 is DisableQVCI. Setting it to false enables QVCI.
+      dbscontrol << EOF
+      M internal 551=false
+      W
+      EOF
+
+      # restart Teradata
+      tpareset -y Enable QVCI
+      ```
+  2. Instruct `dbt` to use `QVCI` mode. Include the following variable in your `dbt_project.yml`:
```
amychen1776, line 27: Unfortunately, I can't seem to get to the deploy working correctly but I think you might need a space before line 17 and before line 27.

website/docs/reference/resource-configs/teradata-configs.md
```diff
@@ -0,0 +1,212 @@
...
+    ```yaml
+    {{
+      config(
+          materialized="table",
+          table_kind="SET"
+      )
+    }}
+    ```
+  For details, see [CREATE TABLE documentation](https://docs.teradata.com/r/76g1CuvvQlYBjb2WPIuk3g/B6Js16DRQVwPDjgJ8rz7hg).
+* `table_option` - define table options. Legal values are:
```
amychen1776: Are these table options supposed to be in the model configuration?

amychen1776: With table options and indexes, I think it would be helpful to tell the reader what are the available configurations that a user can set in that field.

website/docs/reference/resource-configs/teradata-configs.md
```diff
@@ -0,0 +1,212 @@
...
+                        BETWEEN DATE '2020-01-01'
+                        AND     DATE '2021-01-01'
+                        EACH INTERVAL '1' MONTH)
+          INDEX index_attrA (attrA) WITH LOAD IDENTITY"
+      )
+    }}
+    ```
+## Seeds
+
+* `use_fastload` - use [fastload](https://github.com/Teradata/python-driver#FastLoad) when handling `dbt seed` command. The option will likely speed up loading when your seed files have hundreds of thousands of rows. You can set this seed configuration option in your `project.yml` file, e.g.:
```
amychen1776, line 215: Do you forsee folks using dbt seed to upload a lot of files? We generally do not recommend using dbt seed for anything larger than say a few thousand rows. This is because every seeded file is version controlled and this is generally not necessary for what they are loading.

amychen1776 (commented):
This is awesome! I left a few comments for clarification.

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: @amychen1776 , thank you for reviewing the PR. I've checked locally and this fragment renders OK:

image

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: Yes, these are table options applicable to table materializations. The page describes possible options for `table_kind`, `table_option`, `with_statistics`, `index`. We also give examples. I've noticed that examples were not properly indented. I've fixed that.

Please, let me know if it doesn't read well and needs further work.

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: We have to educate our customers that `dbt seed` is not a general-purpose data load tool and is reserved for data that is required for app logic and must be tracked in git as it changes. Customers still expect us to speed things up in case they want to load large csv files.

If you think I should add verbiage here that enforces the point, please let me know.

website/docs/reference/resource-configs/teradata-configs.md
amychen1776: Ah epic! Thank you for the confirmation. This is nitpicky but could you alias the link URL?

website/docs/reference/resource-configs/teradata-configs.md
amychen1776: Oh okay! I think rather than providing all the legal values, this section would make more sense listing out the potential input and maybe explaining/linking to the explanations of what they mean

website/docs/reference/resource-configs/teradata-configs.md
amychen1776: Yes please! A big call out box warning that this point would be great

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: thanks for the suggestion. Done.

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: Done :)

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: Amy, I've not done any usability tests for the docs yet, but my gut feeling is that, if we don't list legal values here, users will be lost when they try to look them up in our docs for [CREATE TABLE](https://docs.teradata.com/r/76g1CuvvQlYBjb2WPIuk3g/B6Js16DRQVwPDjgJ8rz7hg). `CREATE TABLE` docs are generic. They are not dbt-aware. What users need to know is, ho to map options to config keys in dbt. I maintained the same names between dbt and `CREATE TABLE` docs, but the naming convention alone may not be enough. If you are OK with it, my preference is to keep the legal values in dbt docs. As we do usability tests and get more feedback from users, we can tweak.

amychen1776: Would you be able to clarify how a user will map those options to the configure keys in the docs? My concern is that folks will not know which legal values map to which field in the configuration. 

It's not that I don't think the legal values are helpful  but more that since a user will not be writing out the create statement in the dbt model, I think it will lead to more confusion on how what they will write in the dbt model file. I want to make sure we are gearing the user to knowing exactly how to create a model file and then filling it out with that they need to materialize their object.

website/docs/reference/resource-configs/teradata-configs.md
amychen1776: thank you!

website/docs/reference/resource-configs/teradata-configs.md
amychen1776: Let me know if a call would be easier to discuss this :)

website/docs/reference/resource-configs/teradata-configs.md
```diff
@@ -44,9 +44,9 @@ seeds:
     }}
     ```
   For details, see [CREATE TABLE documentation](https://docs.teradata.com/r/76g1CuvvQlYBjb2WPIuk3g/B6Js16DRQVwPDjgJ8rz7hg).
-* `table_option` - define table options. Legal values are:
+* `table_option` - defines table options. The config supports multiple statements. The definition below uses the Teradata syntax definition to explain what statements are allowed. Square brackets `[]` denote optional parameters. The pipe symbol `|` separates statements. Use commas to combine multiple statements as shown in the examples below:
```
amychen1776, line 47: Perfect!

website/docs/reference/resource-configs/teradata-configs.md
adamtwo: @amychen1776 , thanks for the meeting today. I've added a commit that tries to implement what we talked about. Please let me know if this version is easier to follow.

website/docs/reference/resource-configs/teradata-configs.md
```diff
@@ -179,7 +186,10 @@ seeds:
     )
   }}
   ```
-  :information_source: Note, unlike in `table_option`, there are no commas between index statements!
+
+  :::info Separators between statements
```
amychen1776: Is this a duplication over the above call out?

adamtwo: Yes. I've just sent another commit to consolidate into one.

amychen1776 (approved):

PR #1033, opened 2022-01-20
Fix broken link in dbt-api

jtcohen6:
- Fix `events` link to `events-logging`
- Add redirect for anyone who's saved the old bad link (I think this is the right thing to do here?)


jtcohen6:
Merging since it's a tiny fix, and something that came up again

PR #1052, opened 2022-01-24
Docs Versioning++

JKarlavige:
## Description & motivation
This introduces versioning to the docs site. This does not use the built-in versioning Docusaurus includes as we do not want multiple duplicates created for different versions of files. This instead uses a global context to set the version and a component to determine if the item within a docs page should show.

## Pre-release docs
- [ ] Yes: please

## Preview
https://deploy-preview-1052--docs-getdbt-com.netlify.app/reference/node-selection/defer


website/docs/reference/node-selection/defer.md
```diff
@@ -17,6 +17,24 @@ Defer requires that a manifest from a previous dbt invocation be passed to the `
...
+<VersionBlock firstVersion="0.21">
+
+```shell
+$ dbt run --select [...] --defer --state path/to/artifacts
+$ dbt test --select [...] --defer --state path/to/artifacts
+```
+
+</VersionBlock>
+
+<VersionBlock firstVersion="0.20" lastVersion="0.20">
```
joellabes: does this imply that firstVersion is a `>=` and lastVersion is a `<=`? There could be behavioural changes in patches as well, are we choosing to not support that? If so then I'd advocate for the headers being 0.21.x, 1.0.x, etc

joellabes:
Will we store a user's selected version as a cookie?

JKarlavige:
> Will we store a user's selected version as a cookie?

That's a good question @joellabes, I think we should. At the moment the version is lost once a user refreshes the page or leaves the site. I added it to the list in Asana to get this done.

website/docs/reference/node-selection/defer.md
JKarlavige: Yes that's correct, although this is an early prototype so far to test how it will work. The `lastVersion` prop will be optional, so that value should only be set if a feature no longer exists but we still want to show it for an older version in the docs. This will prevent us from having to update all lastVersion props throughout the docs each time we add a new version to the site.

You make a good point regarding the patches. If a feature was released on version 0.20.3 for example, and we only have v0.20.0 and v1.0.2 in the versions dropdown, the user would have to select v1.0.2 to view that section. I think we should discuss this with @runleonarun to start narrowing down specifics on what versions we will have on the site, whether we should support patches, etc.

runleonarun: I believe when @jtcohen6 and I talked about this, we decided to document major (1.x) versions only. The idea being that minor versions would be bug fixes only that should not normally affect docs. I think 1.x is the furthest we'd need to slice the docs, right Jeremy?

jtcohen6: @runleonarun Right on, with just the minor clarification that `1.x` = "minor" and `1.x.y` = "patch".

To that end:

> There could be behavioural changes in patches as well, are we choosing to not support that? If so then I'd advocate for the headers being 0.21.x, 1.0.x, etc

IMO any behavioral changes in patches would restore/represent the "correct" behavior for that entire minor version family. If we change something in v1.0.2, it's to _correct_ the behavior that everyone using v1.0 was meant to experience. As an operating principle, I think we should assume that everyone using v1.0 is / ought to be using the latest patch for v1.0.

> If a feature was released on version 0.20.3 for example

- We will not be releasing any new features in patch versions (such as v0.20.3), only minor versions (v0.21.0)
- We will not be releasing any new versions pre-v1.0

JKarlavige: Thanks for the details @jtcohen6, helpful to know what is expected from minor and patch releases. From both and editor and development standpoint, excluding patches from the versioning on the site will reduce the complexities of having to manage the versions overall.

It would be easier to maintain and add new blocks of versioned content, while also simplifying the logic quite a bit to check if the feature is available for the current selected version.

website/src/theme/DocPage/index.js
```diff
@@ -0,0 +1,185 @@
...
+            className={clsx(
+              'container padding-top--md padding-bottom--lg',
+              styles.docItemWrapper,
+              {
+                [styles.docItemWrapperEnhanced]: hiddenSidebarContainer,
+              },
+            )}>
+            {!pageAvailable && dbtVersion && (
+              <div className={styles.versionBanner}>
+                <Admonition type="caution" title={`This feature is not available in the selected version: ${dbtVersion}`} ></Admonition>
```
jtcohen6: In addition to warning people that the page is unavailable for their current version, could we also tell them the first version in which the feature does become available? Then we get to frame this more positively:
```
🎉 New feature!
Unfortunately, it's not available in ${dbtVersion}
Upgrade to ${firstVersion} or later to use it
```

website/src/theme/DocPage/styles.module.css
```diff
@@ -0,0 +1,93 @@
...
+  .docItemWrapperEnhanced {
+    max-width: calc(
+      var(--ifm-container-width) + var(--doc-sidebar-width)
+    ) !important;
+  }
+}
+
+/* Page Not Available Banner */
+:local(.versionBanner) > div {
+  background: var(--ifm-menu-color);
```
jtcohen6: With the needed caveat that I am very much not a designer: I was expecting something in the yellow/orange/red color family, to contrast more with the standard menu color

website/sidebars.js
```diff
@@ -617,3 +617,6 @@ module.exports = {
     },
   ],
 };
+
+
+module.exports = sidebarSettings
```
jtcohen6, line 634: Whoa, versioned sidebar?? Fancy!

website/docs/reference/node-selection/defer.md
```diff
@@ -17,6 +17,24 @@ Defer requires that a manifest from a previous dbt invocation be passed to the `
...
+
+<VersionBlock firstVersion="0.21" lastVersion="1.0">
+
+```shell
+$ dbt run --models [...] --defer --state path/to/artifacts
+$ dbt test --models [...] --defer --state path/to/artifacts
+```
+
+</VersionBlock>
+
```
jtcohen6: Nice!! This works intuitively and is super easy to use.

The only thing that doesn't seem to work is specifying a `lastVersion` only. I don't know if/when we'd ever actually want to do that. Perhaps limitations, caveats, or known bugs that can be removed when we've resolved them, and no longer bear mentioning? Even so, it's easy enough to just write as `<VersionBlock firstVersion="0.0" lastVersion="...">`

website/dbt-versions.js
```diff
@@ -0,0 +1,19 @@
+exports.versions = [
+    "1.2",
+    "1.0",
+    "0.21",
+    "0.20"
+]
+
+exports.versionedPages = [
```
jtcohen6: In case others are confused at first, as I was:
- These are for pages documenting functionality are _entirely_ unsupported in older (or newer) versions of dbt
- It's not necessary to add every page to this list, every time there's a small difference in functionality across versions. That's what the `<VersionBlock>` is for

website/dbt-versions.js
```diff
@@ -0,0 +1,19 @@
+exports.versions = [
+    "1.2",
+    "1.0",
+    "0.21",
+    "0.20"
```
jtcohen6: Copying from Slack:

> I'm not strictly opposed to including v0.20 + v0.21, especially since we already have those modern/legacy tabs for the `--models`/`--select` rename. I wonder if we can do that, while still making clear that only v1.0 and newer are "officially supported" these days? Perhaps by showing a banner when a user selects v0.20 or v0.21? "You're viewing docs for an outdated version of dbt"?

Context: Starting with v1.0, a minor version of dbt will be "officially supported" for 12 months from its initial release. All pre-v1.0 versions are past their official support period. The new Core Versions page (#1061) will also help with clarifying + setting these expectations.

I could see us adding "end of life" dates to these version constants, and then warning people when looking at docs for a version that:
- has an EOL date fast approaching (say, within the next 3 months)
- is already past its official support period

@runleonarun @JKarlavige What do you think of showing those types of banners for _all_ pages? I think it could go a long way toward teaching users about our version policies. I'm happy to kick that further down the road, and say it's out of scope for this PR. Such a banner also ought to link people to the "Understanding dbt Core versions" page, which is soon to exist :)

jtcohen6 (commented):
This is fantastic work! So excited about what's already here, and about the next steps. I left a handful of comments, none of which is blocking

website/dbt-versions.js
jtcohen6: How [Jinja](https://jinja.palletsprojects.com/en/2.11.x/) does this, for instance:

Screenshot 2022-02-21 at 13 15 46

website/src/theme/DocPage/index.js
JKarlavige: I think this is a great way to frame it! I will update the banner with this content.

website/src/theme/DocPage/styles.module.css
JKarlavige: @runleonarun and I looked at some color options based on our brand colors. I agree we should adjust this to find a way to make it more of an alert, and separate from a standard note.

website/docs/reference/node-selection/defer.md
JKarlavige: At the moment, the `firstVersion` is required and the `lastVersion` is optional. Rather than having to set `firstVersion="0.0"`, it would be a better experience to be able to exclude the firstVersion prop altogether. That should definitely be possible, so I will revisit this to adjust

website/dbt-versions.js
JKarlavige: That is spot on, thanks for noting that here for others looking over it. I have an outline of documentation for the versioning [in progress here](https://www.notion.so/dbtlabs/Versioning-on-Docusaurus-c6a4a41a66cd4ea2970854cc42cb5b70), and these notes will definitely be great to have in there as well.

website/dbt-versions.js
JKarlavige: This would be a great feature to have, and i'm open to getting it added into this PR. Docusaurus has a banner for this which works with their built-in versioning. [Example here](https://61c1eade2ffed20008aaac30--docusaurus-2.netlify.app/docs/2.0.0-beta.10). 

Since we are already showing a banner for the versioned pages if they are not available for that version, it should be straight forward enough to adjust the versions const to add an "end of life" property for each version. If the end of life date has passed, we can show the alert.

In the meantime, it would be great to sort out the language we would like to use for this unsupported version banner.

runleonarun: > In the meantime, it would be great to sort out the language we would like to use for this unsupported version banner.

I can work on this!

website/dbt-global-variables.js
```diff
@@ -0,0 +1,15 @@
...
+      {
+        "name": "dbt Core Lite",
+        "version": "1.0"
+      },
+      {
+        "name": "dbt Lite",
+        "version": "0.20"
+      }
+    ]
+  }
```
joellabes: @runleonarun what sorts of things do we think we'll be defining in here? At EP we had a Brand variable for our automated emails and it wound up being more of a pain than a benefit, because we did a bad job of guessing what components were going to be renamed over time.

Made up example - if we had `dbtCloud` as a catchall variable for the SaaS product, and then we come up with a fun name for the IDE like `The Power Console` (🤢 ) then we still have to go and split out the new component into its smaller chunks... 

Slightly concerned at the thought that we might have recency bias around "what if the company name changes again" which will hopefully not happen again any time soon! 

100% open to being talked out of it if you've found them useful in a past life though

joellabes (commented):
I spent some more time messing around today. Some thoughts: 
- it's weird to me that the [metrics page disappears from the sidebar](https://deploy-preview-1052--docs-getdbt-com.netlify.app/docs/building-a-dbt-project/metrics) when I'm using an old version - it feels like my breadcrumbs are broken. 
- The version data isn't in the URL anywhere, which makes it impossible to share version-specific docs over Slack/support tickets 
- Design-wise, it's weird to have the version picker on the right hand sidebar, which is the jumping off point to _leave_ the site (dev blog, dbt learn, community links...)

JKarlavige:
> I spent some more time messing around today. Some thoughts:
> 
> * it's weird to me that the [metrics page disappears from the sidebar](https://deploy-preview-1052--docs-getdbt-com.netlify.app/docs/building-a-dbt-project/metrics) when I'm using an old version - it feels like my breadcrumbs are broken.
> * The version data isn't in the URL anywhere, which makes it impossible to share version-specific docs over Slack/support tickets
> * Design-wise, it's weird to have the version picker on the right hand sidebar, which is the jumping off point to _leave_ the site (dev blog, dbt learn, community links...)

Thanks for the feedback Joel, i'll add more context for each here:

- The sidebar items hide entirely if they're not available for the current version. However, I think it's still very much an open discussion on whether we want to hide them entirely, or if the banner which shows above the page would be enough
- In order to share a version-specific url, this can be done by adding a `version` param to the url, for example: https://deploy-preview-1052--docs-getdbt-com.netlify.app/?version=0.21
- For the version positioning in the nav, this can be switched from right to left within the `docusaurus.config.js` file on line 251. At the moment it's set to right, but can be moved to the left by updating the setting there

jtcohen6:
> In order to share a version-specific url, this can be done by adding a version param to the url, for example: https://deploy-preview-1052--docs-getdbt-com.netlify.app/?version=0.21

Makes sense! Would it be possible to update/store the value in the URL every time someone selects a version option? (Joel - really good catch - this is so so so important for troubleshooting/support)

joellabes:
Should the picker show when on the Dev blog? feels a bit strange given that changing it won't do anything: https://deploy-preview-1052--docs-getdbt-com.netlify.app/blog/customer-360-view-identity-resolution 

Also, the version takes a while to populate in the header when it's accessed by URL: https://www.loom.com/share/c411cab8d63b41c3a9e4e048f0f5c646

website/dbt-global-variables.js
runleonarun: @joellabes It is a good way to easily rename a feature without the hassle of missing many instances like we would with search & replace. But more importantly, it's a way to standardize on spelling and capitalization of a feature or any name in dbt. A great example is `dbt Labs` If we use a product variable for when we use the company name, then it will decrease the chance there will be typos (dBt labs`</gasp>`). 

Product variables will be really important as we scale and more folks are writing docs. For example, do we say dbt Core v1 or dbt Core v1.0 or dbt Core Version 1.0 ?!? Product variables will also be useful when we're ready to translate. The translators can ignore everything in a variable OR translate in one place, consistently.

website/src/theme/DocPage/styles.module.css
runleonarun: We did some more color spelunking today and Jason is going to make some changes to improve the notes and banners.

website/dbt-versions.js
JKarlavige: The latest two commits add this feature, with the EOD date being set in the `dbt-versions` file.

JKarlavige:
> Should the picker show when on the Dev blog? feels a bit strange given that changing it won't do anything: https://deploy-preview-1052--docs-getdbt-com.netlify.app/blog/customer-360-view-identity-resolution
> 
> Also, the version takes a while to populate in the header when it's accessed by URL: https://www.loom.com/share/c411cab8d63b41c3a9e4e048f0f5c646

Thanks for sharing this @joellabes, I don't think we need to show the version nav item if in the blog directory. @runleonarun what are your thoughts with this?

For the version nav item, I will hide the version nav item entirely until the data has loaded to clean this up a bit.

JKarlavige:
> > In order to share a version-specific url, this can be done by adding a version param to the url, for example: https://deploy-preview-1052--docs-getdbt-com.netlify.app/?version=0.21
> 
> Makes sense! Would it be possible to update/store the value in the URL every time someone selects a version option? (Joel - really good catch - this is so so so important for troubleshooting/support)

This is a great point. At the moment it is not setup to add the version to the URL, however I feel that this _should_ be possible. Will add it to the asana task to look into.

JKarlavige:
> > > In order to share a version-specific url, this can be done by adding a version param to the url, for example: https://deploy-preview-1052--docs-getdbt-com.netlify.app/?version=0.21
> > 
> > 
> > Makes sense! Would it be possible to update/store the value in the URL every time someone selects a version option? (Joel - really good catch - this is so so so important for troubleshooting/support)
> 
> This is a great point. At the moment it is not setup to add the version to the URL, however I feel that this _should_ be possible. Will add it to the asana task to look into.

Hey @jtcohen6, from looking into this further I don't believe it's something that will fall within the realm of the initial scope of this task. I wouldn't rule it out entirely, but it may be better fit for a "v2" round of updates for the versioning feature. 

I'm hesitant to attempt to fit this into the initial release, as it would be a global change affecting every URL. This is something I'd want to separate out into it's own project to see what options there would be and spend a good amount of time working on and testing. It may involve tapping into the Docusaurus routing to append the version onto every URL.

I believe @runleonarun has a list of potential features to add into this project. We should definitely note this as a higher-priority item to circle back to after we go live.

runleonarun:
> I believe @runleonarun has a list of potential features to add into this project. We should definitely note this as a higher-priority item to circle back to after we go live.

@JKarlavige Yep! I just [added it here](https://www.notion.so/dbtlabs/Docs-versioning-project-3f670deb35fe4df7bf8e184363b2dfb1#ab5b9d0f837f4b5291918b0c4e30a002). and I will create a GitHub issue to track Versioning V2.

website/src/theme/DocPage/index.js
```diff
@@ -0,0 +1,223 @@
...
+              'container padding-top--md padding-bottom--lg',
+              styles.docItemWrapper,
+              {
+                [styles.docItemWrapperEnhanced]: hiddenSidebarContainer,
+              },
+            )}>
+            {!pageAvailable && dbtVersion && firstAvailableVersion && (
+              <div className={styles.versionBanner}>
+                <Admonition type="caution" title={`New feature!`} icon="🎉 " >
+                  <p style={{'marginTop': '5px', 'marginBottom': '0'}}>Unfortunately, it's not available in {dbtVersion}</p>
```
runleonarun: @runleonarun revise this banner text!

website/src/theme/DocPage/index.js
```diff
@@ -0,0 +1,223 @@
...
+  })
+
+  useEffect(() => {
+    let threeMonths = new Date(EOLDate)
+    threeMonths.setMonth(threeMonths.getMonth() - 3)
+
+    if(new Date() > new Date(EOLDate)) {
+      setEOLData({
+        showEOLBanner: true,
+        EOLBannerText: `This version is no longer supported. The latest stable version is ${latestStableRelease}`
```
runleonarun: @runleonarun review this text!

website/src/theme/DocPage/index.js
```diff
@@ -0,0 +1,223 @@
...
+      })
+    } else if(new Date() > threeMonths) {
+      setEOLData({
+        showEOLBanner: true,
+        EOLBannerText: `This version is nearing the end of support. The latest stable version is ${latestStableRelease}`
```
runleonarun: @runleonarun review this text

website/snippets/test-snippet.md
```diff
@@ -0,0 +1,8 @@
+---
+---
+
+### Header 2
+
```
runleonarun, line 5: @runleonarun Use real life snippet!

website/dbt-versions.js
```diff
@@ -0,0 +1,31 @@
...
+  {
+    version: "1.2",
+    EOLDate: "2023-01-01"
+  },
+  {
+    version: "1.0",
+    EOLDate: "2022-05-20"
+  },
+  {
+    version: "0.21",
```
runleonarun, line 11: @runleonarun put real life EOL scenarios in here. 

cc @jtcohen6 for awareness

website/src/theme/DocPage/styles.module.css
runleonarun: We settled on brand colors and got approval form Design!

website/dbt-versions.js
```diff
@@ -0,0 +1,31 @@
+exports.versions = [
+  {
+    version: "1.2",
```
runleonarun:
```suggestion
    version: "1.1",
```

JKarlavige:
Sharing here for visibility. I have the first rough draft for documentation on this feature together here: https://www.notion.so/dbtlabs/Versioning-on-Docusaurus-c6a4a41a66cd4ea2970854cc42cb5b70

website/docs/reference/node-selection/defer.md
runleonarun: @JKarlavige Here are some versioning clarifications that I discussed with Jeremy this week. I also captured them in our [notes for the week](https://www.notion.so/dbtlabs/March-8th-2022-26f5c53fe8864843974e3767a4bac31b).
    * We will not be surfacing patch version number or x. So, use 1.1 and not 1.1.x. We will message this in the Core versioning policy, which I captured in a follow-up task https://github.com/dbt-labs/docs.getdbt.com/issues/1217
    * Versioning in the picker will include `.20`,`.21`, `1.0` and higher. BUT we will remove `.20` and `.21` on June 30th, and communicate this in a banner that says they will no longer be supported after 6/30 and in the core versioning policy page. Also captured in a follow-up task https://github.com/dbt-labs/docs.getdbt.com/issues/1217
   
Let me know if there's anything else we can clarify!

website/dbt-versions.js
runleonarun: @JKarlavige Should I comment this out until the 1.1 release?

JKarlavige: @runleonarun committed your suggestion and commented out 1.1!

JKarlavige:
Hey @jtcohen6 and @joellabes, thank you for your feedback and input so far on this project. We are aiming to take this live a week from now on March 17th. There is still a few places with placeholder data we need to swap out, but a majority of the functionality is ready to go. I opened this PR for a final run-through and review, and any help with the review would be greatly appreciated!

@lfracalossiFA adding you for review as well

runleonarun:
@amychen1776 I wonder if you'd also want to take a look at these upcoming changes to let us know if you notice anything. Jason linked this above, but it's worth relinking the [documentation he wrote up](https://www.notion.so/dbtlabs/Versioning-on-Docusaurus-c6a4a41a66cd4ea2970854cc42cb5b70) for us to use the new features.

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
...
+  //   version: "1.1",
+  //   EOLDate: "2023-01-01"
+  // },
+  {
+    version: "1.0",
+    EOLDate: "2022-05-20"
+  },
+  {
+    version: "0.21",
+    EOLDate: "2021-05-20"
```
runleonarun, line 12:
```suggestion
    EOLDate: "2022-06-30"
```

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+  {
+    version: "0.20",
+    EOLDate: "2020-10-17"
```
runleonarun, line 16:
```suggestion
    EOLDate: "2022-06-30"
```

runleonarun:
```suggestion
    EOLDate: "2022-06-30"
```

website/dbt-versions.js
runleonarun:
```suggestion
    EOLDate: "2022-06-30"
```

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
+exports.versions = [
+  // {
+  //   version: "1.1",
+  //   EOLDate: "2023-01-01"
+  // },
+  {
+    version: "1.0",
+    EOLDate: "2022-05-20"
```
runleonarun:
```suggestion
    EOLDate: "2022-12-03"
```

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
+exports.versions = [
+  // {
+  //   version: "1.1",
+  //   EOLDate: "2023-01-01"
```
runleonarun: @jtcohen6 I think we need this to be available for the beta, right?
```suggestion
  //   EOLDate: "2023-03-18"
```

lfracalossiFA:
I have a UX question, really for the whole team. I went to the [URL Jason put in the PR description](https://deploy-preview-1052--docs-getdbt-com.netlify.app/reference/node-selection/defer), and it's unclear to me what version I'm actually looking at here. It says "v1.0" in the dropdown, but it also says "This version is nearing the end of support. The latest stable version is 1.0." So am I not in fact viewing 1.0 docs when I look at that page?

Could this be confusing to users? Maybe I'm missing something that would be obvious to dbt power users, though.

JKarlavige:
> 

This is due to us still having placeholder dates for each version. @runleonarun has a note to update those to the correct dates. In the meantime, I will adjust that date forward a year to reflect more how it would actually look when live!

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
...
+  // },
+  {
+    version: "1.0",
+    EOLDate: "2023-05-20"
```
runleonarun: @jtcohen6 how long do you want to have 1.0 show up in the docs?
```suggestion
    EOLDate: "2023-12-03"
```

website/src/theme/DocPage/index.js
runleonarun:
```suggestion
        EOLBannerText: `This version of dbt Core is nearing the end of its [critical support period](/docs/core-versions). For better performance, improved security, and new features, you should upgrade to ${latestStableRelease}, the latest stable version.`
```

website/src/theme/DocPage/index.js
```diff
@@ -0,0 +1,223 @@
...
+              styles.docItemWrapper,
+              {
+                [styles.docItemWrapperEnhanced]: hiddenSidebarContainer,
+              },
+            )}>
+            {!pageAvailable && dbtVersion && firstAvailableVersion && (
+              <div className={styles.versionBanner}>
+                <Admonition type="caution" title={`New feature!`} icon="🎉 " >
+                  <p style={{'marginTop': '5px', 'marginBottom': '0'}}>Unfortunately, it's not available in {dbtVersion}</p>
+                  <p>Upgrade to {firstAvailableVersion} or later to use it</p>
```
runleonarun:
```suggestion
                  <p> You should upgrade to {firstAvailableVersion} or later if you want to use this feature.</p>
```

website/src/theme/DocPage/index.js
runleonarun:
```suggestion
                  <p style={{'marginTop': '5px', 'marginBottom': '0'}}>Unfortunately, this feature is not available in dbt Core version {dbtVersion}</p>
```

website/src/theme/DocPage/index.js
runleonarun:
```suggestion
        EOLBannerText: `This version of dbt Core is [no longer supported](/docs/core-versions). No patch releases will be made, even for critical security issues. For better performance, improved security, and new features, you should upgrade to ${latestStableRelease}, the latest stable version.`
```

website/src/theme/DocPage/index.js
runleonarun: @JKarlavige Can we add the the EOL date do we already have that as a variable?

runleonarun (approved):
@JKarlavige and @jtcohen6 I revised the support notes and added real life version scenarios.  Let me know if you feel like any of it should change!

Jason, I was wondering if I could add the EOL date in the note too.

website/src/theme/DocPage/index.js
JKarlavige: @runleonarun Yes, the EOL date is already available using the `EOLDate` variable, in the same way we add the `latestStableRelease` variable to the banner text. The EOL date format is currently how we set it in the dbt-versions file: `2023-05-20`. Let me know if you prefer to display the date differently and I will push an update.

JKarlavige:
@runleonarun Committed the suggestions made for the content in the banners. Holding off on committing the EOLDate updates. Once you and @jtcohen6 have those sorted out, feel free to commit them!

lfracalossiFA (approved):
Looks good to me! There is still a test snippet that it looks like Leona is working on still, but I assume you are already aware of it.

JKarlavige:
Put together a [Loom video](https://www.loom.com/share/514bcd6f62b4483e88a662003d555e3f) to walk through how the end-of-life dates, and versioned pages work. These won't be visible in the preview, as our current dates set won't trigger these banners.

JKarlavige:
> 

Awesome! And yep Leona is leaving that in place, and will being adding actual snippets down the road. It's okay to have that demo snippet, as long as we're not actually using that demo snippet within a `<Snippet>` component anywhere on a page.

breezyfasano (approved):
lgtm!!! great work.

website/dbt-versions.js
jtcohen6: ✅ correct!

website/dbt-versions.js
```diff
@@ -0,0 +1,20 @@
...
+    version: "1.0",
+    EOLDate: "2023-12-03"
+  },
+  {
+    version: "0.21",
+    EOLDate: "2022-06-30"
+  },
+  {
+    version: "0.20",
+    EOLDate: "2022-06-30"
```
jtcohen6, line 16: This feels like a fine middle ground for now. @runleonarun and I discussed this at length in our last 1:1.

My main concerns are:
- These versions have already reached EOL from a Core maintenance perspectively, but they'll continue to be supported in Cloud until June 30.
- The selection of v0.20 or v0.21 from the version dropdown implies that the user will see docs that are fully relevant for v0.20 or v0.21, which isn't quite true. We'll add some language to the `Core Versions` page to reflect this: [EOLDate](https://github.com/dbt-labs/docs.getdbt.com/issues/1217)

Ideally, we'd have a banner that we can show users for the next 3 months, while v0.20 + v0.21 are special exceptions. I don't think that should block shipping all the good work that's gone into this feature, and which we badly need to unblock other initiatives.

jtcohen6 (approved):
Let's ship it! Great work on this @JKarlavige @runleonarun

website/dbt-versions.js
JKarlavige: For sure! This initial release will be minimum as far as actual versioned content goes. The only versioned content this PR currently includes come from the Node Selector > Commands pages, converting the tabs to the versionBlock component. 

I can imagine the more versioned content we add, the more we will want to adjust. This is a great item to add to our list for continuous improvements on this project.

PR #1069, opened 2022-02-01
Add discourse comments

breezyfasano:
## Description & motivation
Adding comments to the dev blog via Discourse embeds

How to Review:
- Preview link: [click on any blog post!](https://deploy-preview-1069--docs-getdbt-com.netlify.app/blog/)
- Below each post content, there should be an embed that says 'start discussion'! That's the discourse embed. Let me know if you all can see it 

A note about Discourse's embed functionality from the Discourse docs:

> One important thing to note with this setup is that users have to navigate to your forum to post replies. This is intentional, as we feel that the posting interface on a Discourse forum is currently much richer than what we could embed via Javascript.

Edit 4:08pm CT:
Here are the comments in action! this does create a discourse topic, but I just commented to test the system works as planned and then deleted the topic it made. Screenshots below.

## On the Website
Screen Shot 2022-02-01 at 4 07 28 PM

## On Discourse Topic
Screen Shot 2022-02-01 at 4 07 21 PM

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -868,3 +868,7 @@ code {
   }
 }
 
+/* Discourse Comment Embed */
+.commentEmbed {
+  margin-top: 2rem;
+}
--- a/website/src/theme/BlogPostPage/index.js
+++ b/website/src/theme/BlogPostPage/index.js
@@ -4,17 +4,17 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React from 'react';
+import React, { useEffect } from 'react';
 import Seo from '@theme/Seo';
 import Head from '@docusaurus/Head';
 import BlogLayout from '@theme/BlogLayout';
 import BlogPostItem from '@theme/BlogPostItem';
 import BlogPostPaginator from '@theme/BlogPostPaginator';
-import {ThemeClassNames} from '@docusaurus/theme-common';
+import { ThemeClassNames } from '@docusaurus/theme-common';
 
 function BlogPostPage(props) {
-  const {content: BlogPostContents, sidebar} = props;
-  const {frontMatter, assets, metadata} = BlogPostContents;
+  const { content: BlogPostContents, sidebar } = props;
+  const { frontMatter, assets, metadata } = BlogPostContents;
   const {
     title,
     description,
@@ -24,8 +24,23 @@ function BlogPostPage(props) {
     tags,
     authors,
   } = metadata;
-  const {hide_table_of_contents: hideTableOfContents, keywords} = frontMatter;
+  const { hide_table_of_contents: hideTableOfContents, keywords } = frontMatter;
   const image = assets.image ?? frontMatter.image;
+
+  useEffect(() => {
+    window.DiscourseEmbed = {
+      discourseUrl: 'https://discourse.getdbt.com/',
+      discourseEmbedUrl: `${window.location.href}`
+    }
+    
+
+    var d = document.createElement('script');
+    d.type = 'text/javascript';
+    d.async = true;
+    d.src = DiscourseEmbed.discourseUrl + 'javascripts/embed.js';
+    (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(d);
+    
+  }, [])
   return (
     <BlogLayout
       wrapperClassName={ThemeClassNames.wrapper.blogPages}
@@ -62,7 +77,7 @@ function BlogPostPage(props) {
           />
         )}
       </Seo>
-      
+
       {/* dbt Custom */}
       <Head>
         <title>{title} | dbt Developer Blog</title>
@@ -81,8 +96,13 @@ function BlogPostPage(props) {
       {(nextItem || prevItem) && (
         <BlogPostPaginator nextItem={nextItem} prevItem={prevItem} />
       )}
+
+
+      <div className="commentEmbed" id='discourse-comments'></div>
+
+
     </BlogLayout>
   );
 }
 
-export default BlogPostPage;
+export default BlogPostPage;          


lfracalossiFA:
I can see the Discourse embed! Is it okay to add comments to it?

breezyfasano:
> I can see the Discourse embed! Is it okay to add comments to it?

That is a very good question. I think yes! My only reservation is that it may or may not create something on the dbt Discourse's forums. But thats ok, we can delete any "testing" comments

krevitt:
Looks great!  A couple usability notes here:

* Discourse topics are unlisted (meaning they don't appear in the discourse feed), so we can feel free to test as needed.
image

* One risk here is that Discourse posts are included in the sitemap ([sitemap here](https://discourse.getdbt.com/sitemap_1.xml))+ indexed by google - since they include the full text of the original post, if they were indexed this would count as duplicate content.  They don't appear in the sitemap (probably bc they're unlisted on Discourse), but we'll want to confirm before going live w/ this.

JKarlavige:
>     * One risk here is that Discourse posts are included in the sitemap ([sitemap here](https://discourse.getdbt.com/sitemap_1.xml))+ indexed by google - since they include the full text of the original post, if they were indexed this would count as duplicate content.  They don't appear in the sitemap (probably bc they're unlisted on Discourse), but we'll want to confirm before going live w/ this.

Did some digging on this, as I was thinking there's no way Discourse didn't think this through. Low and behold, it was indeed an issue, with the discussion [here](https://meta.discourse.org/t/add-option-to-set-canonical-url-to-embed-url/103915/13).

The fix was merged in back in 2020 ([PR Here](https://github.com/discourse/discourse/pull/9134/files)).

I believe we will have to check the option in Discourse > Settings > Posting > "Set the canonical URL for embedded topics to the embedded content's URL." to ensure the correct canonical is set.

lfracalossiFA (changes requested):
Just wanted to add context here: I spoke with Breezy about the styling of the embed, and it looks like this can be controlled within the "embedded CSS" in the Discourse theme. So @breezyfasano I'm requesting changes based on that -- basically just making the "Start the Discussion" link, and any comments, match the docs styling.

JKarlavige (changes requested):
All the code and functionality looks good! Just marking this for requested changes to ensure the canonical url/duplicate content issue Krevitt noted is tested and verified, whether it's using the Discourse setting or another solution.

breezyfasano:
Thanks for the comments and guidance folks :)
Update: I have styled the embed widget to match the docs styling a bit more, and I found the canonical URL setting in the backend! Requesting re-review.

Screen Shot 2022-02-02 at 11 26 52 AM

JKarlavige (approved):
lgtm! Can confirm the canonical urls on the Discourse post now point to the original domain from the embed
image

lfracalossiFA (changes requested):
Hi Breezy,
This is looking more on brand! I think we're almost there. 

Any chance of using the Source Sans Pro font for that embed? I know we use it on Discourse, but I'm not sure if we have access to it in the embedded CSS. You could try adding `font-family: var(--font-family)` to the element in the embedded CSS, as it looks like it might already be set up as a CSS variable. If the variable doesn't work, you can try manually setting the font-family to `SourceSansPro, Arial, sans-serif `.

(As a total aside, I wonder if it might be nice to present that text as a button like we use on both docs and Discourse -- teal background, white text, flip the colors on hover. @krevitt any thoughts on if that's a good use of Breezy's time?)

krevitt:
Button styling would be great if it’s quick!  If it looks like it’ll be a drag we can skip it.

breezyfasano:
Okie doke! I went ahead and matched the button styling to the coalesce CTA button on docs! Unfortunately, I couldn't get the font to pull in to the embed :( but I set it to arial, which is the fallback font on the main site for Source Sans Pro. Let me know what you think!

lfracalossiFA:
Hey @breezyfasano,
It looks like the button styling only applies to the "continue discussion" on posts that already have comments. The "Start Discussion" text looks like it's still text only. Can you take a look? 

Thanks!
Lise

breezyfasano:
Done and done!

lfracalossiFA (approved):
lgtm!

breezyfasano:
Thank you! Merging in, and will QA shortly

gwenwindflower:
I apologize for swooping in late with this, but we're fairly settled on the Community team with archiving Discourse and promoting GH Discussions as our centralized area for forum-like discussion as well as Stack Overflow-like question asking. I just would hate for this go live and then we archive Discourse in a month. Hopefully there's a way to swap out this or similar frontend interface but feed it [using the GH API](https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions) instead of Discourse?



The primary concern here is that we want to consolidate the places where knowledge lives, which also maps to the places that the community and Community team need to monitor. Focusing this on GH consolidates knowledge and activity, minimizes surface area. Additionally, there would be a super helpfully transparent and connected record of the discussion and activity around a piece of knowledge's conception, creation, and continued life and growth after publish.

Super happy to discuss more on a call as this is a big change. Also definitely not something that needs to necessarily block this going live, but very much something I want y'all to be aware of, as it's very likely Discourse will be set it into a frozen archive state soon.

lfracalossiFA:
Hey @gwenwindflower -- let me swoop in with another consideration! This is the first I'm hearing of archiving Discourse, and in fact we have a plan to migrate it to the new dbt.com domain in two weeks! (See [here](https://www.notion.so/dbtlabs/getdbt-com-dbt-com-migration-c474d4076ae14042a4654e7ef0e7cf8a#96651489745b467bb1328b0b8995f493)). 

I have no strong feelings one way or another about this, but if we are archiving discourse.dbt.com, I think @krevitt needs to be aware of that, so that we can make a call not only on this PR (which has been merged but reverted), but also on the dbt.com migration runbook.

PR #1101, opened 2022-02-09
Add session method

JCZuurmond:
## Description & motivation

Documentation updates that go along with https://github.com/dbt-labs/dbt-spark/pull/279

## To-do before merge

- [ ] Ensure PR https://github.com/dbt-labs/dbt-spark/pull/279 is merged.

## Pre-release docs
- [ ] Yes: please
- [x] Unsure: we'll let you know! --> It's part of a `dbt-spark` release.

### DIFF
--- a/website/docs/reference/warehouse-profiles/spark-profile.md
+++ b/website/docs/reference/warehouse-profiles/spark-profile.md
@@ -21,6 +21,11 @@ dbt-spark can connect to Spark clusters by three different methods:
 - `odbc` is the preferred method when connecting to Databricks. It supports connecting to a SQL Endpoint or an all-purpose interactive cluster.
 - `http` is a more generic method for connecting to a managed service that provides an HTTP endpoint. Currently, this includes connections to a Databricks interactive cluster.
 - `thrift` connects directly to the lead node of a cluster, either locally hosted / on premise or in the cloud (e.g. Amazon EMR).
+- `session` *FOR ADVANCED USERS* use a pySpark session.
+
+:::info session
+The `session` connection method is intended for advanced users. This connection method is not supported by dbt cloud.
+:::
 
 ### ODBC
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/warehouse-profiles/spark-profile.md
```diff
@@ -21,6 +21,11 @@ dbt-spark can connect to Spark clusters by three different methods:
 - `odbc` is the preferred method when connecting to Databricks. It supports connecting to a SQL Endpoint or an all-purpose interactive cluster.
 - `http` is a more generic method for connecting to a managed service that provides an HTTP endpoint. Currently, this includes connections to a Databricks interactive cluster.
 - `thrift` connects directly to the lead node of a cluster, either locally hosted / on premise or in the cloud (e.g. Amazon EMR).
+- `session` *DO NOT USE FOR PRODUCTION WORKLOADS* use a pySpark session.
+
+:::info session
+The `session` connection method is not intended for production workloads because data models might be read into memory.
```
JCZuurmond: @jtcohen6 : The only time this happens when it's not _really_ required is when `fetchone` is called. (Then all data is fetched into memory and the first row is popped from that list.) It would also be a problem when `fetchall` is called with other connection methods.

My question: why do we consider this connection method to be not ready for production workloads? Or vice versa, when do we consider a connection method to be ready for production workloads?

JCZuurmond (commented):
Please tell me we what should be added to better explain the `session` connection method.

website/docs/reference/warehouse-profiles/spark-profile.md
jtcohen6: Fair question! I have two concerns, motivating this cautionary note:
- dbt should not be processing any data transformation itself. For production workloads, it should be running in a compute resource firmly separated from any compute actually performing the data transformation. Interaction with a local `pyspark` session blurs that boundary, though I understand that it's quite similar in most practice. (I appreciate the clarifying point about `fetchone`, as shifting to the other side of that boundary line.) For instance, I do not see us supporting this connection method in dbt Cloud.
- Practically, the maintainers of `dbt-spark` (me + a few colleagues) aren't pySpark experts (yet), so I have concerns about our ability to defend, debug, and explain this functionality when folks run into rough edges. With that goal in mind, I'd also be willing to call this functionality "experimental" or "for advanced users." If you feel that concern is misplaced, I'm open to hearing it!

JCZuurmond: ok, thanks for your reply, on the points:

- Spark sessions can be configured to connect to a non-local Spark cluster. It does not have to be on the same compute as dbt. 
- I also do not see this being supported in the dbt cloud
- I like the "for advanced users". 

I have updated the PR accordingly

jtcohen6:
Planning to:
- Update this branch with latest changes from `current`
- Add in a "VersionBlock" so that users will only see this when they select v1.1 from the dropdown (since that's now a thing we can do!)

jtcohen6:
Thanks @JCZuurmond! Closing in favor of https://github.com/dbt-labs/docs.getdbt.com/pull/1288, which adds (new) versioned-docs logic, and a note to the v1.1 migration guide

PR #1184, opened 2022-02-28
remove TM from header

jasnonaz:
Updates the page title / nav to remove the TM mark while we evaluate how to showcase this in a way that is less visually distracting


runleonarun (approved):
LGTM

PR #1195, opened 2022-03-02
initial testing setup

JKarlavige:
## Description & motivation
Initial testing setup for Docs

## Pre-release docs
- [ ] Yes: please

## Preview Link
https://deploy-preview-1195--docs-getdbt-com.netlify.app/docs/building-a-dbt-project/building-models/configuring-incremental-models#defining-a-uniqueness-constraint-optional

For the `Defining a uniqueness constraint (optional)` header on page above, a green `tip` block should show for v1.1, but not for any other versions.


website/src/components/variable/index.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
```
eiof: We could probably remove this and `Var` before merging, right?

website/package.json
```diff
@@ -30,14 +31,31 @@
...
     "path-browserify": "^1.0.1",
     "process": "^0.11.10",
     "stream-http": "^3.2.0",
     "style-loader": "^1.1.3",
     "svg-inline-loader": "^0.8.2",
     "tty-browserify": "0.0.1"
   },
+  "jest": {
+    "transform": {
+      "^.+\\.(js|jsx)$": "babel-jest"
```
eiof: It looks like Docusaurus uses [SWC](https://swc.rs/) instead of Babel [for its transformation](https://github.com/facebook/docusaurus/blob/main/jest.config.mjs#L42)? [This looks like it was a very recent change](https://github.com/facebook/docusaurus/pull/6944). There could be minor differences in transformation output, which could mean the component we test is different than what's in production. If so, I think we should try to use SWC as well?

website/babel.config.js
```diff
@@ -0,0 +1,13 @@
...
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current'
+        }
+      }
+    ]
+  ],
+};
```
eiof: If we try using SWC, we might no longer need this file, and need to figure out how to [apply a similar config to what Docusaurus is doing](https://github.com/facebook/docusaurus/pull/6944/files#diff-bc4913d026ae51bef2bc8b7a1c040fdf52b999d99e6fa40d87b379f98a60e0ea).

website/src/components/variable/index.test.js
JKarlavige: Yeah good call. I'll run through the components we added for versioning, and I'll either update this to include some actual tests, or remove it altogether.

website/package.json
JKarlavige: That makes sense, I will try switching that out and removing the babel config for a similar config as Docusaurus. I appreciate the input with this!

JKarlavige:
@eiof Been making progress on this. I switched out babel for swc and the tests run locally. However, when the Netlify build runs it fails with the error `npm ERR! notsup Unsupported platform for @swc/core-android-arm-eabi@1.2.170: wanted {"os":"android","arch":"arm"} (current: {"os":"linux","arch":"x64"})`

JKarlavige:
@eiof _Finally_ got this to build. Was running into the three following errors:

- `npm ERR! notsup Unsupported platform for @swc/core-android-arm-eabi@1.2.174` - removing package.lock and node_modules and letting them regenerate seemed to resolve this.
- `Error: Can't resolve '@theme/Seo' in '/opt/build/repo/website/src/theme/BlogPostPage'` Docusaurus updated to beta.18 during one of the npm installs. We need to make a few updates to swizzled components before being able to upgrade to beta.18.
- `Module not found: Error: Can't resolve 'react-is' in '/opt/build/repo/website/node_modules/styled-components/dist'` - not sure what the source of this was, but installing the `react-is` package got the build to finally pass.

I believe this is finally ready for review!

lfracalossiFA:
Am I seeing the correct "tip" here? I think so, I just want to double check. 

Screen Shot 2022-05-04 at 9 41 07 AM

I assume there's no real way to test the testing setup (heh) from my end of things, right?

website/package.json
```diff
@@ -2,13 +2,14 @@
...
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-beta.17",
-    "@docusaurus/plugin-ideal-image": "^2.0.0-beta.17",
-    "@docusaurus/preset-classic": "^2.0.0-beta.17",
-    "@docusaurus/theme-search-algolia": "^2.0.0-beta.17",
+    "@docusaurus/core": "2.0.0-beta.17",
+    "@docusaurus/plugin-ideal-image": "2.0.0-beta.17",
+    "@docusaurus/preset-classic": "2.0.0-beta.17",
+    "@docusaurus/theme-search-algolia": "2.0.0-beta.17",
```
eiof, line 12: What's the motivation for locking these versions? Should we have concerns about this?

eiof (approved):
Yay! This config using `swc` feels much better than overriding babel configs 😅 . I think this is a good start; we could discuss adding testing to continuous integration next if the team is interested.

website/package.json
JKarlavige: > What's the motivation for locking these versions? Should we have concerns about this?

While working on this, the docusaurus packages updated to `beta.18` during one of the package installs/updates. I was then seeing an error:
`Error: Can't resolve '@theme/Seo' in '/opt/build/repo/website/src/theme/BlogPostPage'` and for the `DocItem` component as well.

We have a lot of swizzled components at this point, and beta.18 introduces a breaking change where `@theme/Seo` is no longer imported. We have been handling the docusaurus updates manually anyway due to this occurring for each update, so I figured I might as well prevent it from being updated automatically altogether.

JKarlavige:
Added the `.swcrc` config file, needed to switch the line `jsx: false` to true for it to work with React. Confirmed i'm able to run the tests on my mac, as well as windows desktop.

@lfracalossiFA If you wanted to test if the tests are running on your end, you can pull this branch down and run the following commands:
`cd website`
`npm i`
`npx jest --watch`

and hitting "a" will run all tests.

image

PR #1197, opened 2022-03-02
Replacing category-specific feeds w/ a global reverse-chronological blogroll

JKarlavige:
## Description & motivation
This replaces feature category-specific feeds on the dev blog homepage with a classic reverse-chronological blogroll.

## Pre-release docs
Is this change related to an unreleased version of dbt?
- [ ] Yes: please
    - update the base branch to `next`
    - add Changelog components: `<Changelog>[New/Changed] in v0.x.0</Changelog>`
    - add links to the "New and changed documentation" section of the latest [Migration Guide](../website/docs/docs/guides/migration-guide)
- [x] No: please ensure the base branch is `current`
- [ ] Unsure: we'll let you know!

## Preview
https://deploy-preview-1197--docs-getdbt-com.netlify.app/blog

PR #1199, opened 2022-03-02
Fix: Remove release notes category

gwenwindflower:
## Description & motivation
It's causing 404s when clicked since no posts use this category yet

### DIFF
--- a/website/blog/categories.yml
+++ b/website/blog/categories.yml
@@ -23,7 +23,3 @@
   display_title: SQL magic
   description: Stories of dbt developers making SQL sing across warehouses.
   is_featured: true
-- name: release notes
-  display_title: Release notes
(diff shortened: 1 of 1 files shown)


jasnonaz (approved):
Excellent thank you!

runleonarun (approved):
@gwenwindflower I see it's creating a bad link until we have content 😢  Thanks for catching this! I will re-add when I add a new release note blog.

gwenwindflower:
we're getting the same error as @JKarlavige 's #1197 so waiting to see if somebody can get to the bottom of that

gwenwindflower:
ah okay this error is just from the Add PR to Project Action, that is okay to fail for now. Merging and we'll figure that out!

PR #1170, opened 2022-02-23
removing ssh

schottj:
removing ssh method for private packages

## Pre-release docs
Is this change related to an unreleased version of dbt?
- [ ] Yes: please
    - update the base branch to `next`
    - add Changelog components: `<Changelog>[New/Changed] in v0.x.0</Changelog>`
    - add links to the "New and changed documentation" section of the latest [Migration Guide](../website/docs/docs/guides/migration-guide)
- [X] No: please ensure the base branch is `current`
- [ ] Unsure: we'll let you know!

PR #1201, opened 2022-03-02
removing ssh

runleonarun:
Reverts dbt-labs/docs.getdbt.com#1196 which reverted https://github.com/dbt-labs/docs.getdbt.com/pull/1170

We need to add more context before removing SSH. 

*  this method should not be used in dbt Cloud and possibly link into something in dbt Cloud land that shows how to set environment variables for the other cited cases on that page.
* Is that true in the Core context?

See this [internal conversation](https://dbt-labs.slack.com/archives/C02SRNY2EQ4/p1645477660510649)


runleonarun:
Addressed in this PR: https://github.com/dbt-labs/docs.getdbt.com/pull/1538

PR #1230, opened 2022-03-14
Update fromyaml.md

bprohan:
Bugfix: Incorrectly formatted delimiter


nghi-ly (approved):
great catch! thanks for fixing, @bprohan !

Issue #1261, opened 2022-03-22
Remove "next" site


jtcohen6:
### Contributions

- [x] I have read the contribution docs, and understand what's expected of me.

### What page(s) or areas on docs.getdbt.com are affected?

Now that docs are versioned, we no longer need the twin deployments at docs.getdbt.com and next.docs.getdbt.com... right?

### What changes are you suggesting?

We can remove logic put by yours truly in:
- `docusaurus.config.js`
- `netlify.toml` (`PRERELEASE = "true"`)
- Netlify to deploy the `next` branch
- Algolia, which runs a separate search index for the "next" site... that never worked _quite_ right

Let's also:
- update the PR template, to mention `VersionBlock` instead of the `next` branch
- set up a CNAME redirect from next.docs.getdbt.com to docs.getdbt.com
- just have one branch named `main` :)

PR #1272, opened 2022-03-26
Add v1.1 migration guide

### LINKED ISSUE #1260: Add v1.1, initialize migration guide
### Contributions

Let's create a migration guide of new & changed documentation, as we've done for past versions: https://docs.getdbt.com/docs/guides/migration-guide/upgrading-to-1-0-0

There shouldn't be any breaking changes for end users. There will be some notable/breaking changes for adapter plugin maintainers and artifact consumers, which we'll want to call out accordingly.

The page needs initialization

### Additional information

Do we like how we've organized these pages in the past? Are there things we want to do differently?

jtcohen6:
resolves https://github.com/dbt-labs/docs.getdbt.com/issues/1260

Let's merge this PR, before all [other v1.1 updates](https://github.com/dbt-labs/docs.getdbt.com/labels/dbt-core%20v1.1.x), since those PRs should update the migration guide.

## Description & motivation

In separate commits, should we want to split any of these off into their own PRs:
- Add v1.1 to version dropdown
- Add "isPrerelease" to VersionContext, and show a warning banner if a user has selected a prerelease version from the dropdown (@JKarlavige would love your eyes on this!)
- Add v1.1 migration guide, with one initial change
- Try using `versionedPages` to show v1.1 migration guide only when v1.1 selected in dropdown (and warn otherwise). This doesn't seem to work as I'd expect, but I'm also not sure it even makes sense for a page like this—we probably do want to show all version guides to all users, no matter which version they're using right now.
- Rename v1.0 migration guide, to follow same `Major.Minor` naming format (with sidebar update + redirect)

## Pre-release docs
- [x] Yes: please

Yes! Let's change the way we do this: https://github.com/dbt-labs/docs.getdbt.com/issues/1261

## Checklist

- [x] The page has been removed from `website/sidebars.js`
- [x] An entry has been added to `_redirects`

### DIFF
--- a/website/src/theme/DocPage/index.js
+++ b/website/src/theme/DocPage/index.js
@@ -59,6 +59,12 @@ function DocPageContent({
     isPrereleaseBannerText: ''
   })
 
+  // Check End of Life date and show unsupported banner if deprecated version
+  const [EOLData, setEOLData] = useState({
+    showEOLBanner: false,
+    EOLBannerText: ''
+  })
+
   useEffect(() => {
     // If version is not isPrerelease, do not show banner
     if(!isPrerelease) {
@@ -69,18 +75,9 @@ function DocPageContent({
     } else {
         setPreData({
           showisPrereleaseBanner: true,
-          isPrereleaseBannerText: `This is a prerelease version. The latest stable version is ${latestStableRelease}`
+          isPrereleaseBannerText  : `This is a prerelease version. The latest stable version is ${latestStableRelease}`
         })
     }
-  })
-
-  // Check End of Life date and show unsupported banner if deprecated version
-  const [EOLData, setEOLData] = useState({
-    showEOLBanner: false,
-    EOLBannerText: ''
-  })
-
-  useEffect(() => {
     // If EOLDate not set for version, do not show banner
     if(!EOLDate) {
       setEOLData({
--- a/website/src/theme/DocPage/index.js
+++ b/website/src/theme/DocPage/index.js
@@ -50,10 +50,31 @@ function DocPageContent({
 
   // Check if page available for current version
   const { versionedPages } = usePluginData('docusaurus-build-global-data-plugin');
-  const { version: dbtVersion, EOLDate, latestStableRelease } = useContext(VersionContext)
+  const { version: dbtVersion, EOLDate, isPrerelease, latestStableRelease } = useContext(VersionContext)
   const { pageAvailable, firstAvailableVersion } = pageVersionCheck(dbtVersion, versionedPages, currentDocRoute.path)
 
-  // Check End of Life date and show unsupported banner if depricated version
+  // Check whether this version is a isPrerelease, and show banner if so
+  const [PreData, setPreData] = useState({
+    showisPrereleaseBanner: false,
+    isPrereleaseBannerText: ''
+  })
+
+  useEffect(() => {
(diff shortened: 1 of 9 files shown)


website/src/theme/DocPage/index.js
```diff
@@ -50,10 +50,31 @@ function DocPageContent({
...
+        showisPrereleaseBanner: false,
+        isPrereleaseBannerText: ''
+      })
+    } else {
+        setPreData({
+          showisPrereleaseBanner: true,
+          isPrereleaseBannerText: `This is a prerelease version. The latest stable version is ${latestStableRelease}`
+        })
+    }
+  })
```
JKarlavige: For this useEffect, it looks like this is causing a `Maximum update depth exceeded` error, due to not having a dependency array.

There's two options to resolve this from what I see. You can either:
- Add the dependency array to the end of this useEffect, making line 75 `}, [dbtVersion])`
- Or, move the if/else statement from this useEffect, into the useEffect below which uses the same `[dbtVersion]` dependency array

This should prevent the useEffect from getting stuck in a constant loop, and only firing when the version has changed.

image

JKarlavige (changes requested):
This is awesome @jtcohen6, nice work! Added one requested change for the useEffect in the DocPage file. Once that is adjusted i'd call this good to go.

website/src/theme/DocPage/index.js
jtcohen6: @JKarlavige Ah, so that's what `}, [dbtVersion])` is doing! There's a reason you're the one who gets to do this stuff for real ;)

Thanks for the quick pointer. I like option 2, since that keeps the logic a bit more consolidated. We shouldn't ever have a version that's _both_ a prerelease _and_ within 3 months of EOL.

JKarlavige (approved):
Verified the depth error is no longer occurring, all looks good to me!

website/src/theme/DocPage/index.js
```diff
@@ -50,16 +50,34 @@ function DocPageContent({
...
+    // If version is not isPrerelease, do not show banner
+    if(!isPrerelease) {
+      setPreData({
+        showisPrereleaseBanner: false,
+        isPrereleaseBannerText: ''
+      })
+    } else {
+        setPreData({
+          showisPrereleaseBanner: true,
+          isPrereleaseBannerText  : `This is a prerelease version. The latest stable version is ${latestStableRelease}`
```
runleonarun: @JKarlavige or @jtcohen6 How can we programmatically add the prerelease version here? 
```suggestion
          isPrereleaseBannerText  : `You are currently viewing v1.1, which is a prerelease of dbt Core. The latest stable version is ${latestStableRelease}`
```

jtcohen6: @runleonarun I think I can do this! comma ça ?

Screenshot 2022-03-29 at 20 38 01

jtcohen6:
Thanks for the feedback both! I'm going to merge this in, in order to
- unblock other v1.1 PRs
- have a "v1.1 landing page" to start linking to from elsewhere

jtcohen6:
Caught one quick thing I needed to fix after merging: https://github.com/dbt-labs/docs.getdbt.com/commit/ebd00034760a0401c52dad44b84f37c27af86b2a

JKarlavige:
Looks great, nice work 👍

PR #1293, opened 2022-03-31
New in v1.1: generic tests with custom names, alt format

### LINKED ISSUE #1269: Generic tests: custom names, alt format
### Contributions

https://docs.getdbt.com/reference/resource-properties/tests

Reflect that this is possible:
```yml
models:
  - name: my_model
    columns:
      - name: id
        tests:
          - unique:
              name: my_cool_test
```

Specifically to:
- give a prettier, more recognizable name to tests with lots of arguments, which have hash-y synthetic names by default
- disambiguate multiple tests that differ _only_ in their configs (which dbt doesn't consider a difference by default, but users might) — good callout in https://github.com/dbt-labs/dbt-core/issues/4102#issuecomment-1072245658 that we'll want a clear example/guidance on this

Also, reflect that dbt now supports an alt format for defining generic tests with lots of properties:
```yml
models:
  - name: my_model
    columns:
      - name: id
        tests:
          # og format
          - accepted_values:
              name: only_allow_abc
              values: ['a', 'b', 'c']
              config:
                severity: warn
                where: 1=1
          # alt format
          - name: only_allow_abc
            test_name: accepted_values
            values: ['a', 'b', 'c']
            config:
              severity: warn
              where: 1=1
```

### Additional information

_No response_

jtcohen6:
resolves #1269

## Description & motivation
Adds two new versioned sections to "Tests" resource property page:
- Defining a custom name for one test — this is pretty long right now, lots of examples and motivation
- Alternative format for defining tests

Both are new in v1.1. I'm observed that version blocks don't prevent headers from appearing on the right-hand sidebar. So if the user has selected an earlier version from the dropdown, I have those same sections appear and display the test `Only supported by v1.1 or newer.`. This feels similar to our approach for [versioning entire pages](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md#versioning-entire-pages).

While here, I found that `database`/`schema`/`alias` are missing from available test configs. They're only relevant if you're using `--store-failures`, but still, worth documenting.

Updates to the migration guide across all these PRs are going to cause such merge conflicts, but I don't think it's a huge deal (worthy of `changie`-like approach). It's important that we show all relevant info to end users, in a way that's approachable and easy to digest.

## Pre-release docs
Is this change related to an unreleased version of dbt? _Yes, following new paradigm!_

### DIFF
--- a/website/docs/reference/resource-properties/tests.md
+++ b/website/docs/reference/resource-properties/tests.md
@@ -152,11 +152,13 @@ This feature is not implemented for analyses.
 
 ## Description
 
-The `tests` property defines assertions about a column, table, or view. The property contains a list of generic tests (referenced by name), which can include the four built-in generic tests available in dbt. It can also include any arguments or [configurations](test-configs) passed to those tests.
+The `tests` property defines assertions about a column, table, or view. The property contains a list of [generic tests](building-a-dbt-project/tests#generic-tests), referenced by name, which can include the four built-in generic tests available in dbt. Any arguments or [configurations](test-configs) passed to those tests should be nested below the test name.
 
 Once these tests are defined, you can validate their correctness by running `dbt test`.
 
 ## test_name
+
+There are four generic tests that are available out of the box, for everyone using dbt.
 
 ### `not_null`
 
@@ -270,7 +272,7 @@ models:
 
 ### Define and use a custom generic test
 
-If you define your own custom generic ("schema") test, you can use that as the `test_name`:
+If you've defined your own custom generic test, you can use that as the `test_name`:
 
 <File name='models/<filename>.yml'>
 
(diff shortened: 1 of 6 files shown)


website/docs/reference/resource-properties/tests.md
```diff
@@ -282,10 +284,195 @@ models:
...
+
+<VersionBlock firstVersion="1.1">
+
+By default, dbt will synthesize a name for your generic test by concatenating:
+- test name (`not_null`, `unique`, etc)
+- model name (or source/seed/snapshot)
+- column name (if relevant)
+- arguments (if relevant, e.g. `values` for `accepted_values`)
+
+It does not include any configurations for the test. If the concatenated name is too long, dbt will use a truncated and hashed version instead. The goal is to preserve unique identifiers for all resources in your project, including tests.
```
runleonarun, line 305: This seems like good practice. Do we also prune tests that haven't been used for awhile?

website/docs/reference/resource-properties/tests.md
```diff
@@ -152,12 +152,14 @@ This feature is not implemented for analyses.
 
 ## Description
 
-The `tests` property defines assertions about a column, table, or view. The property contains a list of generic tests (referenced by name), which can include the four built-in generic tests available in dbt. It can also include any arguments or [configurations](test-configs) passed to those tests.
+The `tests` property defines assertions about a column, table, or view. The property contains a list of [generic tests](building-a-dbt-project/tests#generic-tests), referenced by name, which can include the four built-in generic tests available in dbt. Any arguments or [configurations](test-configs) passed to those tests should be nested below the test name.
```
runleonarun: > The `tests` property defines assertions about a column, table, or view.

Should we add a minor example like, "For example, you can test for duplicate or null values." I know we mention it later but it could help people who pop in the page for a second.

website/docs/reference/resource-properties/tests.md
```diff
@@ -152,12 +152,14 @@ This feature is not implemented for analyses.
...
 
 Once these tests are defined, you can validate their correctness by running `dbt test`.
 
 ## test_name
```
runleonarun: I see "name" in the code. Should this title be "Test name" or is there a meaning here?

runleonarun (approved):
I didn't have too much here. Eventually we might want to redesign our reference docs so they're easier to navigate. Maybe Winnie and I can come up with a few ideas.

website/docs/reference/resource-properties/tests.md
jtcohen6: Yes! — _and_, this is a great philosophical question about the info arch of our docs. We talk in great depth about what tests are and why you'd want them in `docs/docs/building-a-dbt-project-tests`. This page is in the "reference" section; its intended reader is someone who's already been watching the show, knows the basic plot of what a test is, and just wants a bare-bones overview of the syntax to copy-paste.

That info arch served us ok for the past few years, but definitely definitely worth a rethink! Was just talking about exactly this with @gwenwindflower the other day :)

website/docs/reference/resource-properties/tests.md
jtcohen6: A more useful section header here might be "Built-in generic tests" or "Out-of-the-box generic tests."

The reason it's `test_name` is because that's the actual technical property for what the thing is inside dbt. That property is set explicitly in the example at the very bottom. You can define a `unique` test as either:
```yml
tests:
  # these are identical, just alternative syntaxes
  - unique
  - test_name: unique
```
Then you can select all `unique` tests in your project via:
```
dbt test --select test_name:unique
```

PR #1301, opened 2022-04-04
Source Status Fresher Docs

sungchun12:
## Description & motivation

All the documentation to equip developers with using the `source_status:fresher` selector. 

## To-do before merge

- [x] Ensure PR [#4256](https://github.com/dbt-labs/dbt-core/pull/4256) is merged

## Pre-release docs
Is this change related to an unreleased version of dbt?
- [x] Yes: please
    - update the base branch to `next`
    - add Changelog components: `<Changelog>[New/Changed] in v0.x.0</Changelog>`
    - add links to the "New and changed documentation" section of the latest [Migration Guide](../website/docs/docs/guides/migration-guide)
- [ ] No: please ensure the base branch is `current`
- [ ] Unsure: we'll let you know!

PR #1327, opened 2022-04-12
Cache the selected database

### LINKED ISSUE #1273: New global config: `cache_selected_only`
### Contributions

- https://docs.getdbt.com/reference/global-configs

- Add new "global" config, `cache_selected_only`, based on the addition in https://github.com/dbt-labs/dbt-core/pull/4860
- Call out that this is "experimental" functionality, which may significantly reduce runtime overhead in large warehouses but may also have some unanticipated / unaccounted-for rough edges. We could use a callout similar to the one we use on https://docs.getdbt.com/docs/building-a-dbt-project/metrics

### Additional information

The PR is not yet merged, but likely to be included in v1.1.0rc1

runleonarun:
closes #1273 

## Description & motivation

Adding content without too much context. 

## Pre-release docs
- [ ] Yes: please

## Checklist

### DIFF
--- a/website/docs/reference/global-configs.md
+++ b/website/docs/reference/global-configs.md
@@ -111,8 +111,20 @@ config:
 This should be considered an _unstable_ config flag because it is experimental and subject to change. We reserve the right to make breaking changes to this config flag.
 :::
 
-When true, caches database objects relevant to selected resource. Otherwise, will cache all database objects related to the project.
+When true, dbt caches database objects relevant to selected resource. Otherwise, dbt will cache all database objects related to the project.
 
+For example, to improve speed and performance while focused on Salesforce models you would select the source and use the `cache-selected-only` flag:
+
+```text
+
+$ dbt build --select source:salesforce 
+$ dbt --cache-selected-only run
+
+```
+
+The previous example enables you to start working with your salesforce models, and dbt will only cache those models instead of the entire project.
+
+To set the config in your `profile.yml`:
 <File name='profiles.yml'>
 
 ```yaml
@@ -124,13 +136,6 @@ config:
 
 </File>
 
-<File name='Usage'>
-
-```text
-dbt --cache-selected-only run
-```
-
-</File>
 </VersionBlock>
 
 ### Checking version compatibility
--- a/website/docs/reference/global-configs.md
+++ b/website/docs/reference/global-configs.md
@@ -103,6 +103,30 @@ config:
 
 </File>
 
+### Cache database objects for selected resource
+
+When true, caches database objects relevant to selected resource. Otherwise, will cache all database objects related to the project.
+
+<File name='profiles.yml'>
+
+```yaml
+
+config:
+  cache_selected_only: true
+
+```
+
+</File>
+
+<File name='Usage'>
+
+```text
+dbt --cache-selected-only run
+```
+
+</File>
+
+
 ### Checking version compatibility
 
 Projects are recommended to set [dbt version requirements](require-dbt-version), especially if they use features that are newer, or which may break in future versions of dbt Core. By default, if you run a project with an incompatible dbt version, dbt will raise an error.
--- a/website/docs/reference/global-configs.md
+++ b/website/docs/reference/global-configs.md
@@ -105,6 +105,10 @@ config:
 
 ### Cache database objects for selected resource
 
+:::caution Experimental config flag
+This should be considered an _unstable_ config flag because it is experimental and subject to change. We reserve the right to make breaking changes to this config flag.
+:::
+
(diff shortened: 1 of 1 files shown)


website/docs/reference/global-configs.md
```diff
@@ -103,6 +103,34 @@ config:
 
 </File>
 
+### Cache database objects for selected resource
+
+:::caution Experimental config flag
+This should be considered an _unstable_ config flag because it is experimental and subject to change. We reserve the right to make breaking changes to this config flag.
+:::
+
+When true, caches database objects relevant to selected resource. Otherwise, will cache all database objects related to the project.
```
runleonarun: @jtcohen6 and possibly @gwenwindflower or @dbeatty10, can you add any context or color here? I'm also curious how or where you'd select this resource to be cached? Thanks!

website/docs/reference/global-configs.md
```diff
@@ -102,6 +102,34 @@ config:
 ```
 
 </File>
+<VersionBlock firstVersion="1.1">
```
runleonarun, line 106:
```suggestion

<VersionBlock firstVersion="1.1">

```

website/docs/reference/global-configs.md
gwenwindflower: definitely!

dbt needs a bunch of metadata about all the actual database objects (relations in dbt) when you run commands, which requires hitting the database, which is always significantly slower than staying local, so it caches all this and relies on that cache when its able to for speed improvements (the specifics of that metadata and the conditions when it can use the cache vs needing to refresh it are not something i've given any thought to so i can't really go deeper than that -- jeremy and doug will be much more useful there).

if you're consistently working on a subset of your DAG though, you don't really need it to cache the whole project, just the part you're working on. this is not a big help on a small project, but could be a huge advantage on massive projects where a developer would never actually run the full project. 

so the selection here as to what is cached is tied to what the developer is selecting in their command(s) -- if i run `dbt build --select source:salesforce` to start working with all my salesforce models, only those will be cached, rather than the entire project as before -- and thus speed 🏃🏻‍♀️! performance 🏋🏻‍♀️!

jtcohen6: @gwenwindflower said it very well! There's an (internal) doc here I started putting together a few weeks ago, would be great to prepare more of this for public documentation: https://www.notion.so/dbtlabs/The-Adapter-Cache-8de07620c80049c4ae66cf576fff3ac5

website/docs/reference/global-configs.md
```diff
@@ -103,6 +103,36 @@ config:
 
 </File>
 
+<VersionBlock firstVersion="1.1">
+
+### Cache database objects for selected resource
+
+:::caution Experimental config flag
+This should be considered an _unstable_ config flag because it is experimental and subject to change. We reserve the right to make breaking changes to this config flag.
+:::
```
jtcohen6, line 112: This hits on most of the right points for me. This is a thing we will be doing not infrequently, as a way to continue shipping innovative stuff in major-version-1 software. e.g.
- metrics
- https://github.com/dbt-labs/docs.getdbt.com/pull/1301

jtcohen6:
This got merged into `dbt-core`, so we're good to document it for v1.1: https://github.com/dbt-labs/dbt-core/pull/5036

website/docs/reference/global-configs.md
```diff
@@ -103,6 +103,36 @@ config:
...
+
+When true, caches database objects relevant to selected resource. Otherwise, will cache all database objects related to the project.
```
runleonarun, line 114:
```suggestion
When true, dbt caches database objects relevant to selected resource. Otherwise, dbt will cache all database objects related to the project. 

For example, if you wanted to improve speed and performance while focused on Salesforce models:
```
$ dbt build --select source:salesforce` 
$ dbt run --cache-selected-only` 
```
After selecting the source and using the `cache-selected-only` flag, you can start working with your salesforce models, and dbt will only cache those models instead of the entire project.

```

website/docs/reference/global-configs.md
```diff
@@ -103,6 +103,41 @@ config:
...
+
+### Cache database objects for selected resource
+
+:::caution Experimental config flag
+This should be considered an _unstable_ config flag because it is experimental and subject to change. We reserve the right to make breaking changes to this config flag.
+:::
+
+When true, dbt caches database objects relevant to selected resource. Otherwise, dbt will cache all database objects related to the project.
+
+For example, to improve speed and performance while focused on Salesforce models you would select the source and use the `cache-selected-only` flag:
```
runleonarun, line 116: @gwenwindflower I added your example. Does this look OK to you?

PR #1382, opened 2022-04-27
Update Using Airflow section with dbt Cloud provider

kentdanas:
## Description & motivation
Replaces the `airflow-dbt` Python package bullet with a reference to the new dbt Cloud Provider, which is the preferred way of orchestrating dbt Cloud jobs from Airflow. Also update a few links and add link out to guide with detailed instructions.


amychen1776 (approved):
LGTM!

PR #1394, opened 2022-04-28
Update getting-help.md

bbrewington:
## Description & motivation
Fixed slack channel names - looks like dbt renamed slack channels so they sort better, and the ones referenced in the website weren't updated

## To-do before merge
(n/a)

## Prerelease docs
(n/a)

## Checklist
(n/a)


jasnonaz (approved):
great catch - thank you!!

PR #1401, opened 2022-04-29
dbt-materialize: update Materialize profile and add configs page

morsapaes:
## Description & motivation

We recently deprecated the custom `index` materialization in the `dbt-materialize` adapter and made index creation part of the model definition as a config (see [Materialize MaterializeInc/database-issues#3371](https://github.com/MaterializeInc/database-issues/issues/3371)) to follow the implementation in `dbt-postgres`.

This PR adds documentation for the new index configuration, and opportunistically introduces some updates to the adapter profile.

**Note:** I tried to respect the order of Materialize in `Profiles (CLI only)` in `Adapter-specific configs`, which is why it wasn't just appended to the existing list!

## Checklist
If you added new pages:


runleonarun:
@morsapaes Thank you so much for submitting this PR! We will get it reviewed and merged as soon as we can!

@amychen1776 and @dataders Is this something one of you can give a technical review before I review?

website/docs/reference/warehouse-profiles/materialize-profile.md
```diff
@@ -43,20 +57,30 @@ dbt-materialize:
...
+dbt debug
+```
+
+If the output reads "All checks passed!", you’re good to go! Check the [dbt and Materialize guide](https://materialize.com/docs/guides/dbt/) to learn more and get started.
+
 ## Supported Features
 
 ### Materializations
 
+Because Materialize is optimized for transformations on streaming data and the core of dbt is built around batch, the `dbt-materialize` adapter implements a few custom materialization types:
```
dataders, line 72: is there a link from the profile info page to the config page? if not, this'd be a great place to add it!

website/docs/reference/resource-configs/materialize-configs.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Performance optimizations
+
+### Indexes
+
+:::info Advanced feature
+  Manually creating indexes in Materialize is an advanced feature that most users **do not** need. See the [Materialize documentation](https://materialize.com/docs/sql/create-index/) for more details.
+:::
+
+Materialized views (`materializedview`), views (`view`) and sources (`source`) may have a list of `indexes` defined. Each [Materialize index](https://materialize.com/docs/sql/create-index/) can have the following components:
```
dataders, line 14: maybe naive, but what does an index on a source mean? What is the impact from an end-user perspective?

dataders, line 14: maybe naive, but what does an index on a source mean?

dataders (approved):
Love the addition here. I also added technical implementation suggestions directly on https://github.com/MaterializeInc/materialize/pull/12030#pullrequestreview-959645237, but these docs look good to me as a first pass, beyond the small issues I brought up

website/docs/reference/resource-configs/materialize-configs.md
morsapaes: An index on a source "materializes" the source, which allows you to _e.g._ directly query it. This isn't something that is recommended for other than prototyping and testing (or in very specific situations where it might make sense!), since it means we'll store every single event available in the source in memory.

website/docs/reference/warehouse-profiles/materialize-profile.md
morsapaes: There's a link in the indexes entry, but that's it! Right now, the configurations page doesn't really provide more than index-specific information; do you think it'd make sense to link it elsewhere? 

I was actually wondering if this section as a whole would actually be a better fit in the configurations page, rather than the profile.

morsapaes: Every other documented database seems to do its own thing, so I wasn't too sure!

morsapaes:
@dataders, just solved a small conflict but otherwise it seems like there are no blockers to merge this? LMK if my answers helped clarify your questions, or if you still think this could use further tweaking!

PR #1403, opened 2022-05-02
rm staging cta, add setup tutorial cta

glin7:
## Description & motivation
Stop promoting Staging since the event has passed. Start promoting the new warehouse-specific tutorials instead.


jasnonaz:
ahh great catch - thanks @glin7! I think @krevitt was thinking we'd switch this over to Coalesce promo curious to hear thoughts between the two

runleonarun (approved):
Great idea! I'm OK with this if everyone else is!

glin7:
No strong opinion! Feel free to keep this PR on the back burner if we need to promote Coalesce instead. But if we're already pointing to Coalesce in other prominent positions on the site, might be nice to throw a little attention to other resources in smaller spots like this.

krevitt:
Yep this works, we can mix in Coalesce soon but let's definitely swap this out

PR #1404, opened 2022-05-02
Mirnawong1 faq addition

runleonarun:
## Prerelease docs

## Checklist


runleonarun:
@JKarlavige, this builds locally, but the Netlify build fails. I want to be able to reference the `id` instead of the path to the file or filename when we use faqs inline for example `<FAQ src="dbt-specific-jinja" />` (not sure what it's using now). I wasn't sure how to make that happen for FAQs.  I'm assuming this is what's causing the build errors?

Hi @mirnawong1 and @sophiad96! I just wanted to capture some of the changes I made so you can understand the fixes I made, and feel free to ask questions about them!
- Added directories that will contain the content for each subcategory (like "models" and "troubleshooting")
- Made all of the content in each directory automatically generated, so when you add content to the folder, and it will show up in the category sidebar.
- I didn't move all of the content into directories because I wasn't sure where it all went.
- You no longer have to edit the `sidebar.js` unless you want to add another category (like "security" Note:  I didn't add "security" because you can't have an empty folder so I didn't want it to break.)
- Shortened some titles using a `sidebar_label: 'How to name models directory'` or by making the title shorter. (see the Models and Tests directories for examples) when you specify a sidebar_label, it overrides the title in the left bar.
- Fixed some of the markdown files to remove the lingering code `</File>`
- Fixed links that no longer worked.
- I also started adding ids to the FAQs.md files in the header. We might need these in each FAQ depending on Jason's answer above.

You can see the changes I made [here](https://github.com/dbt-labs/docs.getdbt.com/pull/1404/files/f6b941a3eb7e0f5db8256beaf54846cb2c778ba1..a894da9ddb17d56c9ae3afd8f8acd395dfe4bd65).

JKarlavige:
@runleonarun you are correct, the FAQs component is referencing the filename with the following code `const file = require('../../../docs/faqs/' + src + '.md')`

Since the faqs are moved into subdirectories, the path above is no longer accurate and causing the build to fail, as it is unable to find those files. We will have to update the FAQs component to support this. 

So to confirm, this should be adjusted to get the FAQ by the `id` in the frontmatter instead of the filename? I will put a ticket into our asana board to get that adjusted. I will open a separate PR to adjust that, and will set it up so it supports both the `id` and filename, so it will work with the current production code, as well as this code once it's merged in. I _think_ that will be possible.

runleonarun:
> So to confirm, this should be adjusted to get the FAQ by the id in the frontmatter instead of the filename? I will put a ticket into our asana board to get that adjusted. I will open a separate PR to adjust that, and will set it up so it supports both the id and filename, so it will work with the current production code, as well as this code once it's merged in. I think that will be possible.

Yep! if that fixes the errors, then I think that's the preferred path. I'd also like to fallback on the filename if no id is provided. Just in case.

website/src/pages/index.js
```diff
@@ -261,7 +261,7 @@ function Home() {
                       <div className="card__footer">
                           <Link
                             className="button button--primary"
-                            to={useBaseUrl('/faqs/example-projects')}>
+                            to={useBaseUrl('/faqs/project/example-projects')}>
```
runleonarun:
```suggestion
                            to={useBaseUrl('/faqs/example-projects')}>
```

website/docs/docs/available-adapters.md
```diff
@@ -5,7 +5,7 @@ id: "available-adapters"
 
 dbt connects to and runs SQL against your database, warehouse, platform, or query engine. It works by using a dedicated **adapter** for each technology. All the adapters listed below are open source and free to use, just like dbt.
 
-If you have a new adapter, please add it to this list via a pull request! See [Documenting your adapter](/website/docs/docs/contributing/documenting-a-new-adapter.md)  for more information.
+If you have a new adapter, please add it to this list via a pull request! See [Documenting your adapter](/docs/contributing/documenting-a-new-adapter.md)  for more information.
```
runleonarun:
```suggestion
If you have a new adapter, please add it to this list using a pull request! See [Documenting your adapter](/docs/contributing/documenting-a-new-adapter.md)  for more information.
```

mirnawong1 (commented):
Approving review per Leona's feedback

mirnawong1 (approved):
Approving and merging pr

PR #1402, opened 2022-05-01
Databricks

schottj:
## Description & motivation
Adding details for the new databricks adapter in dbt Cloud. Keeping info about the current connection details, which we'll deprecate alongside how to connect to the new adapter.

PR #1407, opened 2022-05-02
fix grammatical error in Databricks profile page

dataders:
## Description & motivation
lil' typo that I noticed while reviewing #1402


schottj (approved):
lgtm

PR #1409, opened 2022-05-03
change `quote-columns` to `quote_columns` in v1.0 migration guide

sophiad96:
## Description & motivation
A user reached out to Support pointing out that in the v1.0 migration guide, under the "Renamed fields in `dbt_project.yml` file" section, the `quote_columns` config is referred to as `quote-columns`. Based on our documentation, `quote_columns` is the correct config. This PR will make that very minor change.


runleonarun (approved):
LGTM! Thanks for fixing this @sophiad96!

PR #1183, opened 2022-02-28
Update connecting-your-database.md

mirnawong1:
adding comment to instruct users to add commented lines when adding private key field:

:::**Note:** At this time ONLY Encrypted Private Keys are supported by dbt Cloud. In order to successfully fill in the Private Key field, you  **must** include the commented lines below when you add the passphrase. The passphrase you add, will be used to decrypt the and leaving the `PRIVATE KEY PASSPHRASE` field empty will will return an error.
:::
**Example:**
```sql
-----BEGIN ENCRYPTED PRIVATE KEY-----
< encrypted private key contents here >
-----END ENCRYPTED PRIVATE KEY-----
```
## Pre-release docs
Is this change related to an unreleased version of dbt?
- [ ] Yes: please
    - update the base branch to `next`
    - add Changelog components: `<Changelog>[New/Changed] in v0.x.0</Changelog>`
    - add links to the "New and changed documentation" section of the latest [Migration Guide](../website/docs/docs/guides/migration-guide)
- [x ] No: please ensure the base branch is `current`
- [ ] Unsure: we'll let you know!

## Checklist
If you added new pages (delete if not applicable):
- [ ] The page has been added to `website/sidebars.js`
- [ ] The new page has a unique filename

If you removed existing pages (delete if not applicable):
- [ ] The page has been removed from `website/sidebars.js`
- [ ] An entry has been added to `_redirects`

PR #1412, opened 2022-05-03
Update connecting-your-database.md

schottj:
adding key size support to the docs


dataders:
@schottj looks like some of your changes have already been incorporated via #1183. I went ahead and merged in those changes as best as I could. Does the current change make grammatical sense as one sentence?

dataders (approved):
let's smash this in!

PR #1418, opened 2022-05-04
Updating to correct Python version

leahwicz:
## Description & motivation
Just noticed that the Python version here still said Python 3.6 when we are now only supporting 3.7 and above.

## Prerelease docs

## Checklist
n/a


jtcohen6 (approved):
Good catch!

Maybe this should link to the [other FAQ about Python version compatibility](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/faqs/install-python-compatibility.md)?
```
<FAQ src="install-python-compatibility" />
```

PR #1419, opened 2022-05-04
Adjust FAQs on Docs site to use `id` instead of filename

JKarlavige:
[Asana task](https://app.asana.com/0/1200099998847559/1202220638270389/f)

Original issue: #1404 

Since the faqs are moved into subdirectories, the path above is no longer accurate for these FAQs withing the FAQs component, causing the build to fail as it is unable to find those files. This updates the FAQ component to resolve this.

Adjusted the FAQ component to support getting the content by both the faq frontmatter id and filename. This will allow it to work with the current production code which gets FAQ items by filename, and by frontmatter IDs once they are set.

To test this PR, I set an id for the `/docs/faqs/test/install-pip-os-prereqs.md` faq file, and moved it into a subdirectory: `/test`.

Once this is live, #1404 should merge `current` to include these updates. The FAQs in subdirectories should begin working as expected and the build should pass.
 
## Before Merge

- [x] Remove ID in `/docs/faqs/test/install-pip-os-prereqs.md` file.
- [x] Move `/docs/faqs/test/install-pip-os-prereqs.md` into original faqs directory.

## Preview Link
https://deploy-preview-1419--docs-getdbt-com.netlify.app/dbt-cli/install/from-source

The first FAQ at the bottom of this page `Does my operating system have prerequisites?` is pulling the faq by id, while the other two are still pulling the faq by filename.

### DIFF
--- a/website/docs/faqs/install-pip-os-prereqs.md
+++ b/website/docs/faqs/install-pip-os-prereqs.md
@@ -1,5 +1,4 @@
 ---
-id: install-pip-os-prereqs
 title: "Does my operating system have prerequisites?"
 Description: "You can check whether your operating system has prerequisites for installing dbt Core."
 
--- a/website/docs/faqs/test/install-pip-os-prereqs.md
+++ b/website/docs/faqs/test/install-pip-os-prereqs.md
@@ -1,5 +1,5 @@
 ---
-
+id: install-pip-os-prereqs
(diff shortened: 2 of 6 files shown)


runleonarun (approved):
This works for me! Thanks for taking care of this so quickly, @JKarlavige!

PR #1421, opened 2022-05-04
Version Nav Item Edit

JKarlavige:
## Description & motivation
Currently, when a version is selected from the navbar, the dropdown does not close automatically. Adjust this to have the dropdown automatically close when a version is selected.

## Preview Link
https://deploy-preview-1421--docs-getdbt-com.netlify.app/docs/building-a-dbt-project/building-models/configuring-incremental-models#defining-a-uniqueness-constraint-optional

This page has versioned components


website/src/css/custom.css
```diff
@@ -43,6 +43,12 @@
   --color-nav-text: #bfc0c0;
 }
 
+.dropdown--version--hide .dropdown__menu {
+  opacity: 0 !important;
```
lfracalossiFA: Jason, Breezy, Is it necessary to use the !important flags here? I generally like to stay away from them if possible. But it's also possible it's entirely necessary in this situation. What do you think?

JKarlavige: Looks like it still works without the `!important` flag, removed in latest commit

lfracalossiFA (approved):
lgtm!

PR #1423, opened 2022-05-05
fixing data type for non null fields for metadata api

vadim82:
## Description & motivation

Updating the graphql introspection query to account for non null fields in the SchemaTable for metadata API documentation


ynambiar (approved):
LGTM!

PR #1425, opened 2022-05-06
missing import for BaseAdapterMethod

codeforkjeff:
code sample for test_basic.py is missing the import for BaseAdapterMethod

## Description & motivation

The import is needed for the code sample to work.


jtcohen6 (approved):
Good catch @codeforkjeff! Thanks for this!

PR #1428, opened 2022-05-06
apprise and authorize Anders of dbt Cloud config docs changes

dataders:
## Description & motivation

cloud configuration has a lot of adapter and versioning related info is my wheelhouse, I should know when it changes and be informed-enough to assist.


amychen1776 (approved):
🚢

PR #1434, opened 2022-05-10
Remove gitlab published docs from example list

w0ut0:
## Description & motivation
Previously, Gitlab had a very nice, elaborate dbt docs website. Unfortunately, it's now secured and not publicly available without authentication.


w0ut0:
Seems available again!

PR #1443, opened 2022-05-12
Prevent homepage tile links from breaking on Safari

john-rock:
## Description & motivation
@cesher pointed out that currently the tile links (e.g. `Read Up` button in the `Docs` tile) on the homepage are not working while using Safari. 

This PR removes the lookbehind regex that is currently not supported on Safari and allows the links to work across all browsers.

## Deploy Preview
https://deploy-preview-1443--docs-getdbt-com.netlify.app/


lfracalossiFA (approved):
Hey John! I did some tests in Chrome, Firefox, and Safari, and the bug seems to be fixed. So from my perspective it's good to go, though I'd definitely want Cesar to sign off on this, too.

I just wanted to ask if you had an understanding of what the : operator in regex does, because I don’t, and Googling it isn’t helping. This is mostly for my curiosity, though.

thisispvb:
`(?:)` creates a non-capturing group (you won't get the result of it saved to a variable like $1).

I really like https://regexr.com/ to work on and test regular expressions and they have some simple docs about the most common use cases as well.

lfracalossiFA:
> `(?:)` creates a non-capturing group (you won't get the result of it saved to a variable like $1).

Ah, so that's what capturing means; I've always wondered!

Do you think that will cause a problem, in this instance?

thisispvb:
I think you can simplify the regex a little bit to `docs\/faqs\/(.*)\.md$` and you will also have to change the index of the data lookup (`0` is the full match and without the lookbehind it will include `docs/faq/`, so `1` is what you want here to only get the first group's content)

Example: https://regexr.com/6lhhc

john-rock:
Wow this is great info thanks @thisispvb. Regex is NOT like riding a bike for me. Anytime coming back to it I lose my balance, I need to work on that.

thisispvb (approved):

lfracalossiFA (approved):
Oops, thought I had approved this already. Apparently not! lgtm!

PR #1448, opened 2022-05-13
Update the exposure page to remove the metrics word

b-per:
Remove the word "metrics" from the exposure page to avoid confusion with the dbt metrics

## Description & motivation

Because [metrics](https://docs.getdbt.com/docs/building-a-dbt-project/metrics) are now a dbt node themselves, some new users could get confused seeing the work `metrics` when looking at the docs on `exposures`


jasnonaz (dismissed):
super smart catch - ty!!

matthewshaver (approved):
Thank you for this clarification @b-per

PR #1449, opened 2022-05-13
Add line break to have the types listed in a list.

arikfr:
## Description & motivation

Currently the list of materialization types is shown like this:
CleanShot 2022-05-13 at 12 01 36@2x

~I assume this is meant to be a list and my guess is that it's not because of the missing line break.~

This is how it looks like with this change:

CleanShot 2022-05-13 at 12 15 29@2x


nghi-ly (approved):

nghi-ly:
@arikfr : thanks for fixing this list! ❤️ 

i'm not sure why the Deploy Preview is failing. will look into fixing this.

website/docs/docs/building-a-dbt-project/building-models/materializations.md
```diff
@@ -5,6 +5,7 @@ id: "materializations"
 
 ## Overview
 <Term id="materialization">Materializations</Term> are strategies for persisting dbt models in a warehouse. There are four types of materializations built into dbt. They are:
```
runleonarun, line 8:
```suggestion

<Term id="materialization">Materializations</Term> are strategies for persisting dbt models in a warehouse. There are four types of materializations built into dbt. They are:
```

PR #1463, opened 2022-05-18
Adds warehouse prop support for WHCode component

john-rock:
## Description & motivation
The problem is that people who are unfamiliar with our docs-as-code setup might be confused as to what is going on in the code and not sure what they should edit.

This adds support to pass in a prop (`warehouse`) to define which code is being used for which warehouse in markdown files where WHCode component is used.

The `warehouse` prop is required and the value of the prop will be used as the label of the tab on the frontend.

This also removes the 4 warehouse limit currently in place and allows us to add new warehouses if needed in the future.


JKarlavige (approved):
This looks good to me! Here is a preview link where the WHCode component is being used: https://deploy-preview-1463--docs-getdbt-com.netlify.app/tutorial/building-your-first-project/build-your-first-models

john-rock:
Hi @runleonarun when you have a moment let me know if this looks like what you were requesting in [this task](https://app.asana.com/0/1200099998847559/1202225356712124/f).

Also, let me know if you have any questions about this change.

website/snippets/tutorial-sql-query.md
```diff
@@ -1,7 +1,7 @@
 <WHCode>
 
 <!-- Bigquery snippet -->
-<div>
+<div warehouse="BigQuery">
```
runleonarun, line 4: @john-rock This looks great!

runleonarun (approved):
Just some questions:
* Is it required to add the `warehouse="warehouse_name"` now? 
* Does the order you put the `<div>s in matter? and will it use the order we display in the docs or will it always use the same order?

john-rock:
Hi @runleonarun no problem, 

>     * Is it required to add the `warehouse="warehouse_name"` now?
Yes, that prop is now required. If the prop does not exist or is empty it will throw an error. `(Expected warehouse prop to be defined for each child of WHCode component)`


>     * Does the order you put the `s in matter? and will it use the order we display in the docs or will it always use the same order?
The order on the frontend is defined by the order of the children within the WHCode component. If we always want it BigQuery, Databricks, Redshift, Snowflake then that must be the order of the children in the WHCode component.


One other thing to mention here is that it will also take the case and spelling of the warehouse prop value for what is displayed on the frontend. For example, if someone uses `warehouse="bigquery"` then that will be the spelling and case that is displayed on the frontend tab.

runleonarun:
@john-rock That sounds good! I think I need to capture all of that in the [Contributing guide](https://github.com/dbt-labs/docs.getdbt.com#writing-content). I think this is good to merge and I will follow up with that content.

PR #1468, opened 2022-05-19
Fix links & grammar on Redshift Optimization blog post

johnblust:
Fixed the grammar in the `even` distribution style sections.
Fixed the section links to point to on-page headers instead of notion.


christineberger (approved):
Looks good to me!

PR #1471, opened 2022-05-20
Update idempotent.md: add word processing example

vergenzt:
Hi! I happened to just recently come up with the word processing example to explain idempotency to some colleagues, and then today noticed there's a dbt glossary page on the subject that was soliciting improvements. 🙂 

Let me know what you think!


vergenzt:
Updated to fix casing on the failed "DML" term reference, remove the "This page needs some love" note, and revise the hover summary.

johnblust (approved):
This is awesome, @vergenzt Thank you so much for taking the time to work on this!! Your contribution here is definitely an improvement from what we had. Go ahead and hit that big green merge button whenever you're ready :)

vergenzt:
If only I had permission (to click merge) I would! As it stands, however, I don’t have write access, so I can’t. 🙂

vergenzt:
Thanks for the review! Happy to help. 👍

johnblust:
Duh, sorry totally forgot about that. I'll get this merged. Thanks again!

PR #1474, opened 2022-05-22
update outdated docs

uncletoxa:
## Description & motivation

Documentation contains an example of  the outdated parameter name


nghi-ly:
thanks so much for catching and providing a fix for this, @uncletoxa  !

i think PR #2435 has addressed these same issues. i will go ahead and close. please don't hesitate to reopen if there's still something remaining.

PR #1494, opened 2022-05-27
Remove old recommendation from `init` command

### LINKED ISSUE #995: dbt init profile behavior changed in dbt 1.0.0, 'default' no longer default 'profile' value
### Contributions

https://docs.getdbt.com/reference/project-configs/profile

> If you are working on more than one project, do not use **`profile: default`** as your profile name (as set by the dbt init command), as it will become hard to manage multiple profiles.

In dbt 1.0.0 I'm seeing different behavior. The `dbt init` command sets the `profile` to the project name, eg `dbt init -s test_project` results in `profile: 'test_project'` not `profile: 'default'`

### Additional information

I'm not quite sure what's the best way to update this line in the document, but it definitely appears invalid now that 1.0.0 is released.

gwenwindflower:
Fixes #995


jasnonaz (approved):
great catch!

PR #1499, opened 2022-05-30
dbt-impala documentation

tovganesh:
## Description & motivation
Add dbt-impala documentation to dbtlabs doc site

## Checklist
- [x] The page has been added to `impala-configs.md` and `impala-profile.md`


website/docs/reference/resource-configs/impala-configs.md
```diff
@@ -0,0 +1,67 @@
...
+title: "Apache Impala configurations"
+id: "impala-configs"
+---
+
+## Configuring tables
+
+When materializing a model as `table`, you may include several optional configs that are specific to the dbt-impala plugin, in addition to the standard [model configs](model-configs).
+
+| Option  | Description                                        | Required?               | Example                  |
+|---------|----------------------------------------------------|-------------------------|--------------------------|
```
dataders, line 11: might be worth adding a column that shows the values these options can take? like which can be `str` or a `List[str]`?

website/docs/reference/resource-configs/impala-configs.md
```diff
@@ -0,0 +1,67 @@
...
+| external | is this an external table - true / false | No | external=true |
+|---------|----------------------------------------------------|-------------------------|--------------------------|
+
+For Cloudera specific options for above parameters see documentation of CREATE TABLE (https://docs.cloudera.com/documentation/enterprise/6/6.3/topics/impala_create_table.html)
+
+## Incremental models
+
+Supported modes for incremental model:
+ - **`append`** (default): Insert new records without updating or overwriting any existing data.
+ - **`insert_overwrite`**: On new data, insert records. If data is updated or deleted, overwrite the entire table.
```
dataders, line 32: > If data is updated or deleted, overwrite the entire table.

AFAIK, according [dbt-spark's docs' description of `insert_overwrite`](https://docs.getdbt.com/reference/resource-configs/spark-configs#the-insert_overwrite-strategy) this is an edge case that happens when a `unique_key` field isn't passed, not the standard behavior for `incremental_strategy=insert_overwrite`. Nothing to do here at this moment, but on dbt-impala it might be worth opening an issue there to zero this discussion out, after which the docs can be updated

website/docs/reference/resource-configs/impala-configs.md
```diff
@@ -0,0 +1,67 @@
...
+     union all
+     select 4 as id, "Name 4" as name, "City 1" as city,
+)
+
+select * from source_data
+```
+
+</File>
+
+In the above example, a sample table is created with partition_by and other config options. One thing to note when using partition_by option is that the select query should always have the column name used in partition_by option as the last one, as can be seen for the ```city``` column name used in the above query. If the partition_by clause is not the same as the last column in select statement, Impala will flag an error when trying to create the model.
```
dataders, line 67: @ericf-firebolt encountered a similar issue when defining External Tables in Firebolt. I wanted to link you to [his solution](https://github.com/firebolt-db/dbt-firebolt/blob/35a3e268b7c001ac405ca9747875633bf14a415c/dbt/include/firebolt/macros/dbt_external_tables/create_external_table.sql#L3-L9) that might be beneficial to you? It might be easier for users to not have to specify the partition in the column list definition

website/docs/reference/warehouse-profiles/impala-profile.md
```diff
@@ -0,0 +1,137 @@
...
+id: "impala-profile"
+---
+
+## Overview of dbt-impala
+
+**Maintained by:** Cloudera    
+**Author:** Cloudera    
+**Source:** [Github](https://github.com/cloudera/dbt-impala)    
+**dbt Cloud:** Currently un-supported    
+**dbt Slack channel** [Link to channel](https://slack.com)
```
dataders, line 12: doesn't have to happen in this PR, but I recommend asking for a `#db-impala` channel to be made in [#channel-requests](https://getdbt.slack.com/archives/C01D8J8AJDA).

website/docs/reference/warehouse-profiles/impala-profile.md
```diff
@@ -0,0 +1,137 @@
...
+
+dbt-impala can connect to Apache Impala and Cloudera Data Platform clusters.
+
+The [Impyla](https://github.com/cloudera/impyla/) library is used to establish connections to Impala.
+
+Two transport mechanisms are supported:
+- binary
+- HTTP(S)
+
+The default mechanism is `binary`. To use HTTP transport, use the boolean option `use_http_transport: [true / false]`.
```
dataders, line 27: food for thought, perhaps make the variable `transport` with values of `https` and `binary` with the latter being the default? Can you explain more about the binary transport mechanism?

dataders (approved):
left some comments and made some merge conflict commits. you can either revise this PR to incorporate my feedback or open a new PR after this one is merged. thanks @tovganesh!

website/docs/reference/resource-configs/impala-configs.md
tovganesh: I will edit this as now we also support insert_overwrite with partition clause (https://github.com/cloudera/dbt-impala/pull/29)

website/docs/reference/warehouse-profiles/impala-profile.md
tovganesh: I have made a request to that channel, will update the docs when the same is created.

PR #1512, opened 2022-06-02
Reorg `Getting Started` sidebar groupings

gwenwindflower:
## Description & motivation
Per @patkearns10 's excellent observation, the Getting Setup page seems to push you towards BQ rather than present all the options equally in design terms -- this PR fixes that, and moves the dbt Core CLI setup out of Learning More and into the group with the warehouses, which is a better fit imo and gives it more equal weight as a setup option. This fits with the overall plan to build out the CLI side of Getting Started in coming weeks/months.


runleonarun:
Just wanted to add a status that we have a few decisions to make before making this change. 
* Perhaps the Core Getting Started section should be its own guide.
* Where should we guide people who don't have a database preference? We had decided BigQuery (which is why the current guide provide that path.)
* Maybe we can suppress the bottom of the page navigation and add a mini toc that offers all database options

runleonarun:
We're taking the guides in a different direction so I'm closing this PR. Please let me know if we should re-open.

PR #1516, opened 2022-06-02
fix link to contribution guide

jasnonaz:
## Description & motivation
Tried to get fancy with my relative linking and it didn't work so putting in an absolute link


gwenwindflower (approved):
:shipit:

PR #1519, opened 2022-06-02
fix broken non-breaking space character

plattegruber:
## Description & motivation
An extra space in an HTML entity was causing it to render incorrectly. We should remove that space so that dropdown titles render correctly.

I noticed it in the faq section of this page: https://docs.getdbt.com/docs/building-a-dbt-project/projects

Let me know if you need anything else for this PR!


runleonarun (approved):
Thank you so much for fixing that error, @plattegruber! I will get this merged soon!

PR #1530, opened 2022-06-06
fix the data links

amychen1776:
Fixing the redshift guide links with the right links. I thought I fixed them in the wrong PR but apparently now. 😓


dataders (approved):
:shipit:

PR #1531, opened 2022-06-06
Update snowflake guide to not reference databricks

Topherhindman:
## Description & motivation

The setup guide for Snowflake was referencing databricks at the end of it when summarizing what a user had accomplished after following the guide.


dataders (approved):
great catch -- thanks @Topherhindman!

PR #1543, opened 2022-06-08
Add banner on dbt docs site

JKarlavige:
## Description & motivation
This sets up the Docusaurus `announcementBar` and adds additional customization to control the bar. The `AnnouncementBar` component is swizzled to add the following custom updates:
- Wrap the entire announcement bar with the link
- Add `announcementBarLink` & `announcementBarActive` properties to the config to provide additional control.

## To-do before merge
Include any notes about things that need to happen before this PR is merged, e.g.:
- [x] Confirm content is correct

## Preview
https://deploy-preview-1543--docs-getdbt-com.netlify.app/

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -69,6 +69,7 @@ var siteSettings = {
       textColor: "#fff",
       isCloseable: true
     },
+    announcementBarActive: true,
     announcementBarLink: "https://www.getdbt.com/resources/dbt-live-expert-series-june-na/",
     prism: {
       theme: (() => {
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -61,6 +61,15 @@ var siteSettings = {
       appId: ALGOLIA_APP_ID ? ALGOLIA_APP_ID : 'dbt'
       //debug: true,
     },
+    announcementBar: {
+      id: "live_qa",
+      content:
+        "Friday, Jun 10, 2022: Join a live Q&A with dbt experts, who'll answer all your 🌶  questions.",
+      backgroundColor: "#047377",
+      textColor: "#fff",
+      isCloseable: true
+    },
+    announcementBarLink: "https://www.getdbt.com/resources/dbt-live-expert-series-june-na/",
     prism: {
       theme: (() => {
         var theme = require('prism-react-renderer/themes/nightOwl');
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -64,13 +64,13 @@ var siteSettings = {
     announcementBar: {
       id: "live_qa",
       content:
-        "Friday, Jun 10, 2022: Join a live Q&A with dbt experts, who'll answer all your 🌶  questions.",
+        "Have questions you want answered live? Join us for a dbt Live: Expert Series session!",
       backgroundColor: "#047377",
       textColor: "#fff",
       isCloseable: true
     },
     announcementBarActive: true,
-    announcementBarLink: "https://www.getdbt.com/resources/dbt-live-expert-series-june-na/",
+    announcementBarLink: "https://www.getdbt.com/events/",
     prism: {
       theme: (() => {
         var theme = require('prism-react-renderer/themes/nightOwl');
--- a/website/src/theme/AnnouncementBar/index.js
+++ b/website/src/theme/AnnouncementBar/index.js
@@ -64,7 +64,6 @@ export default function AnnouncementBar() {
 }
 
 function AnnouncementBarContent({ content, styles }) {
-  console.log('styles', styles)
   return (
     <div
       className={styles.announcementBarContent}
--- a/website/src/theme/AnnouncementBar/index.js
+++ b/website/src/theme/AnnouncementBar/index.js
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import React from 'react';
+import clsx from 'clsx';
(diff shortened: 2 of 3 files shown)


website/docusaurus.config.js
```diff
@@ -61,6 +61,16 @@ var siteSettings = {
       appId: ALGOLIA_APP_ID ? ALGOLIA_APP_ID : 'dbt'
       //debug: true,
     },
+    announcementBar: {
+      id: "live_qa",
```
lfracalossiFA, line 65: Is there any significance to the name you chose for the `id`?

website/src/theme/AnnouncementBar/index.js
```diff
@@ -0,0 +1,76 @@
...
+    return null;
+  }
+
+  const {content, backgroundColor, textColor, isCloseable} = announcementBar;
+
+  return (
+    <div
+      className={styles.announcementBar}
+      style={{
+        backgroundColor,
```
lfracalossiFA, line 27: Is there a reason `backgroundColor` isn't a key-value pair, like `color`? This may be my own ignorance of React/Docusaurus. (I imagine a big chunk of this file is from the original component we swizzled this from.

website/src/theme/AnnouncementBar/index.js
```diff
@@ -0,0 +1,76 @@
...
+            color: textColor,
+          }}
+          >
+          <AnnouncementBarContent content={content} styles={styles} />
+        </a>
+      ) : (
+        <AnnouncementBarContent content={content} styles={styles} />
+      )}
+      
+      {isCloseable ? (
```
lfracalossiFA, line 49: I'm not sure I'm familiar with this syntax (the wrapped in a { } aspect, not the ternary operator -- I understand the latter) and I'd love to know more!

website/src/theme/AnnouncementBar/index.js
```diff
@@ -0,0 +1,76 @@
...
+    <div
+      className={styles.announcementBar}
+      style={{
+        backgroundColor,
+        color: textColor,
+      }}
+      role="banner">
+      {isCloseable && <div className={styles.announcementBarPlaceholder} />}
+      {announcementBarLink ? (
+        <a
```
lfracalossiFA, line 33: On the rendered site, it looks like we get `div.announcementBarContent_dpRF` inside the new `a` you set up. Is there any reason we couldn't collapse that into a single `a` tag? I tend to cut out unnecessary markup as much as possible. ("It would be a pain to refactor" is a very good reason -- I don't want you to have to swizzle `AnnouncementBarContent` as well!)

lfracalossiFA (approved):
This looks good to me! I have a few questions, but they are mostly for my own edification.

website/docusaurus.config.js
JKarlavige: Negative, the docs state `Any value that will identify this message.`
https://docusaurus.io/docs/api/themes/configuration#announcement-bar-id

website/src/theme/AnnouncementBar/index.js
JKarlavige: Yeah you're correct with a majority of this being the default from Docusaurus. With JS, if the `key` name is = to the `value`, you can use shorthand. Since the css selector for `background-color` is `backgroundColor` in React, shorthand can be used here.

I'm guessing the Docusaurus team used `textColor` instead of `color` was just to be more specific when setting that color in the config: 
```
announcementBar: {
      id: "live_qa",
      content:
        "Have questions you want answered live? Join us for a dbt Live: Expert Series session!",
      backgroundColor: "#047377",
      textColor: "#fff",
      isCloseable: true
    },
```

website/src/theme/AnnouncementBar/index.js
JKarlavige: The `{}` structure is the JSX way of using javavscript within html. So a common use case is for displaying html content conditionally based on a JS variable state. In this case, i'm adding additional html (the anchor link) if `announcementBarLink` is set in the config.

Here is a good read on when to use `{}` in JSX: https://reactjs.org/docs/introducing-jsx.html

lfracalossiFA:
Thank you for answering my super basic questions!

website/src/theme/AnnouncementBar/index.js
lfracalossiFA: Ah, I probably should have known that. Guess I'd never seen it used for multi-line conditionals before. Thanks!

website/src/theme/AnnouncementBar/index.js
JKarlavige: My main reason for keeping it as-is here is i'm not 100% sure if Docusaurus is depending on the `announcementBarContent` being a `div` under the hood anywhere. If any JS they are using for the banner functionality is looking for a `div`, and I change it to an anchor, it could potentially break it.

So what I did here was broke out the `announcementBarContent` div into it's own component, and am wrapping that component with an anchor if `announcementBarLink` is set in the config. That way none of the existing setup is altered, and just adding additional functionality for the new `announcementBarLink` config option I added.

Docusaurus doesn't support having the entire banner as a link, and expects users to add links inline with the text, such as their example in the docs. This doesn't match our functionality on WWW where the whole banner is clickable, which was the reason for swizzling the component and wrapping the content with an anchor.
```
announcementBar: {
      id: 'support_us',
      content:
        'We are looking to revamp our docs, please fill <a target="_blank" rel="noopener noreferrer" href="#">this survey</a>',
      backgroundColor: '#fafbfc',
      textColor: '#091E42',
      isCloseable: false,
    },
```

lfracalossiFA: Ah. I figured it was More Complicated than it looked!

JKarlavige:
Approved by @JakkiJay as well via Slack!

PR #1545, opened 2022-06-09
Support policy

schottj:
Adding more detail on our policy extension for running dbt <1.0 in dbt Cloud.


website/docs/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-choosing-a-dbt-version.md
```diff
@@ -25,28 +25,31 @@ The example job seen in the screenshot above belongs to the environment "Redshif
...
+
+We will continue to update this table so that customers know when we plan to stop running different versions of Core in dbt Cloud.
 
 <Snippet src="core-versions-table" />
 
-Starting in v1.0, dbt Cloud will ensure that you're always using the latest compatible patch release of `dbt-core` and plugins, including all the latest fixes. You may choose to try prereleases of those patch releases before they are generally available.
+:::warning ⚠️ v0.X Non-Supported Period
+ We are giving accounts until the end of June 2022 to upgrade to dbt 1.0 or later. Pre-dbt 1.0 versions will not receive patch fixes, and our support team will no longer assist with dbt version specific help on non-supported versions of dbt. Additionally, jobs running dbt versions prior to 1.0 will start experiencing service disruptions before the end of the year 2022 and will be removed from the dbt Cloud context by end of the year 2022. You will receive additional email and in app notification before disruption to your production jobs.
```
dataders, line 37: this means after june 2022, there's no more patch fixes? or are we already not patching pre 1.0 versions?

website/docs/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-choosing-a-dbt-version.md
```diff
@@ -70,4 +73,4 @@ Additionally upgrading to more recent versions of dbt Core will enable better pe
 
 #### Need help upgrading?
 
-If you want more advice on how to upgrade your dbt projects, check out our [migration guides](docs/guides/migration-guide/upgrading-to-0-21-0) and our [upgrading Q&A page](docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-upgrading-dbt-versions).
+If you want more advice on how to upgrade your dbt projects, check out our [migration guides](/guides/migration/versions/upgrading-to-v1.1) and our [upgrading Q&A page](docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-upgrading-dbt-versions).
```
dataders: perhaps the migrations landing page is more evergreen?
image

```suggestion
If you want more advice on how to upgrade your dbt projects, check out our [migration guides](/guides/migration/versions/) and our [upgrading Q&A page](docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-upgrading-dbt-versions).
```

dataders (approved):
LGTM! just 1 question and 1 suggestion

website/docs/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-choosing-a-dbt-version.md
schottj: We've already stopped releasing patches to pre 1.0 versions. For instance, we only patched the Networkx bug in 1.0 and up.

jeremyyeo: This is a strange one because it is correct that we actually pinned a lower version of the networkx library/package in the dbt-core for versions >= 1.0 meaning if you `pip install dbt-core==0.21` locally, you will still be experiencing the bug.

In dbt Cloud though, we did lower the networkx library as well in order to fix this for all dbt versions available in cloud (https://github.com/dbt-labs/dbt-cloud/pull/5364/files).

The status quo right now is if you install dbt-core < 1.0 on your own, you will have the selection bug but if your job is running in cloud with dbt < 1.0, you will not have the selection bug.

jeremyyeo: In saying that, this sounds like the right language here - there is no expectation that dbt < 1.0 in cloud will work / get patch fixes (such as we did in the situation I described above) come this period.

jeremyyeo (approved):
LGTM

website/docs/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-choosing-a-dbt-version.md
schottj: right - I think post end of June, we will no longer fix core problems in <1.0, so if a Networkx-like problem happens in July, we'll tell folks this 0.x version of dbt is in a non-support period, you need to upgrade.

PR #1556, opened 2022-06-11
Update setting-up-bigquery.md

novica:
## Description & motivation
I was going through the tutorial and noticed something that looks like a simple typo. I think it should be `data`, not `date`.


novica:
Additionally, and, I am not sure how / where to propose this, as I am not entirely sure it is correct: 
While going through the Big Query tutorial I noticed tables are never added to the datasets. Later, when trying different examples from the tutorial (notably: sources) I got errors. So I poked around Big Query, and added tables to the datasets. This **seems** to resolve the issues for me.

PR #1435, opened 2022-05-10
Add guidance on moving from dbt-spark to dbt-databricks

dataders:
## Prerelease docs
If this change is related to functionality in a prerelease version of dbt (delete if not applicable):
- [ ] I've added versioning components, as described in ["Versioning Docs"](../contributing/versioningdocs.md)
- [ ] I've added a note to the prerelease version's [Migration Guide](../website/docs/docs/guides/migration-guide)

## Checklist
If you added new pages (delete if not applicable):
- [ ] The page has been added to `website/sidebars.js`
- [ ] The new page has a unique filename

If you removed existing pages (delete if not applicable):
- [ ] The page has been removed from `website/sidebars.js`
- [ ] An entry has been added to `_redirects`

PR #1576, opened 2022-06-14
Add example with the threads config

b-per:
`threads` is supported on the databricks adapter but not being in the supplied example, people might forget to configure it and will in that case be automatically configuring only 1 thread

## Description & motivation

This change should help user set up their databricks project more easily by just copy/pasting the example and defining the number of threads they want to use.


dataders:
thanks @b-per! two questions:
1. is there any guidance on suggested number of threads depending on the compute and workload type that is specific to databricks?
2. should threading be a consideration for users migrating from dbt-spark to dbt-databricks? I have #1435 open to help with that

b-per:
Hi Anders.

1. I am not an expert in databricks tuning but a quick search led me to [this page](https://docs.microsoft.com/en-us/azure/databricks/sql/admin/sql-endpoints) and this section:

> To handle more concurrent users for a given query, increase the cluster count. Databricks recommends a cluster for every ten concurrent queries.

Which to my understanding could translate that with 1 cluster, up to 10 threads is OK.


2. I don't know if there is a difference between concurrency on an HTTP endpoint and a "typical" databricks compute cluster. My gutt feel is no, and whatever number of threads in the spark profile could be replicated to the databricks profile.

dataders (approved):
thanks @b-per!

PR #1579, opened 2022-06-14
Fix 'getting help' link

rmcgibony:
This URL was bad. Looks like it should point to the 'legacy' subdirectory (?)

## Description & motivation
Fixing a bad link ("Page Not Found")


nghi-ly:
thanks so much for flagging, @rmcgibony! really appreciate the love you're giving the docs <3 

was looking 👀 into this on your behalf and found that this has since been fixed you created this PR. closing. please reopen if there's anything else that hasn't been addressed!

PR #1580, opened 2022-06-15
Update the docs for dbt-impala

tovganesh:
## Description & motivation

* add slack channel for dbt-impala
* update doc for insert_overwrite


dataders (approved):
thanks @tovganesh!

PR #1618, opened 2022-06-24
Update snapshotting-freshness-for-one-source.md

jeremyholtzman:
I believe this is missing "source:" syntax


jeremyholtzman:
Hey @runleonarun would you be able to review this? This is my first time contributing so if I'm missing anything please let me know :)

coapacetic (approved):
LGTM 🚀

runleonarun (approved):
This looks great @jeremyholtzman 

Did you do a search to see if we need to update the language elsewhere? Does this apply to all versions of dbt?

jeremyholtzman:
@runleonarun good call, I just did a search and found other references which I just added. I also looked through our documentation and found that this only applies to v 0.21.0 and later according to https://docs.getdbt.com/reference/commands/source.

runleonarun:
@jeremyholtzman Thanks for updating! Being that .20 is almost EOL I think it's OK to merge this PR without versioning the old syntax for .20. 

We can add a note similar to the one on the Source page:

>:::info As of dbt Core v0.21, you need to prefix sources with the `source:` selection method. In previous versions of dbt, sources were specified by name only. :::

jeremyholtzman:
@runleonarun, sure thing, I just added the note you mentioned on both pages

runleonarun (approved):
🚢 it!

PR #1619, opened 2022-06-24
Initialize documentation for v1.2

dbeatty10:
Continues #1609
## Things to review

[Link to Netlify Preview](https://deploy-preview-1619--docs-getdbt-com.netlify.app/reference/dbt-jinja-functions/cross-database-macros)

<details>
  <summary>Location within navigation</summary>

  ### Questions
  - The content is currently within Jinja Reference > dbt Jinja functions > cross-database macros. Is this where we want it? Or somewhere else?

### Screenshot

image

</details>

<details>
  <summary>Group by category or alphabetical listing (or both)</summary>

  ### Questions
  - The first draft contains both an alphabetical listing and grouped into categories. Do we favor one over the other?
      - I prefer retaining a grouping by functional category as it makes it easier for me understand the usage of each macro when it is surrounded by more context.
      - Snowflake includes both types of listings: by category ([here](https://docs.snowflake.com/en/sql-reference/functions.html) and [here](https://docs.snowflake.com/en/sql-reference/functions-date-time.html#list-of-functions)) and [alphabetical](https://docs.snowflake.com/en/sql-reference/functions-all.html).
  - Single page listing all macros? Or separate pages based on functional category?
      - My only preference is that the detailed entires are grouped into categories rather than alphabetical. An alphabetical listing of links feels desirable too.

### Screenshots

Alphabetical:

image

Functional grouping:

image

</details>

<details>
  <summary>Format of examples</summary>

### Questions
  - Include `dbt.` prefix for macros? Or drop?
      - The result feels more readable without the prefix to me.
  - Convert to a SQL query that can actually execute?
      - It could be hard to create examples that will work cross-database without making them a bit complicated with common table expressions (CTE) and usage of other macros 
  - Include the filename wrappers (i.e., `models/example.sql`)?
      - this feels a bit inaccurate unless we supply SQL examples that are fully executable
 - Multiple variations per example? Or just a single variation per example?
     - I strongly prefer multiple variations to make it easier to see how to express column names vs. SQL string literals vs. an expression containing a function.

### Screenshot
image

</details>

<details>
  <summary>Versioning this page</summary>

  ### Questions
  - I added an entry to `dbt-versions.js`, but it doesn't appear to be taking effect. Versioning didn't appear to take effect for [this page](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.1) either (even though it is listed within `dbt-versions.js`).
      - Do we know how to get this to work?

### Screenshots

[Versioning working](https://docs.getdbt.com/reference/dbt-jinja-functions/selected_resources):

- image

[Versioning not working](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.1):

- image

</details>

## Description & motivation
We want to document the macros introduced [here](https://github.com/dbt-labs/dbt-core/pull/5298) so that:
- package maintainers (and other users) know what they do and how to use them
- adapter maintainers know how to implement them

These macros came from [dbt_utils](https://github.com/dbt-labs/dbt-utils/tree/0.8.6#cross-database-macros), but few of them had pre-existing documentation. Existing documentation was lifted-and-shifted when available, but but most of this content is net new.

The most novel feature is organization into functional groups (similar to the [groupings within Snowflake docs](https://docs.snowflake.com/en/sql-reference-functions.html)).

## To-do before merge

- [x] Determine if `current_timestamp*` macros should be included or not.
- [x] Determine if data type macros should be included or not (i.e., whatever abstraction on top of `api.Column.translate_type()` that we go with).

## Prerelease docs
- [x] I've added versioning components, as described in ["Versioning Docs"](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md)
- [x] I've added a note to the prerelease version's [Migration Guide](https://github.com/dbt-labs/docs.getdbt.com/tree/current/website/docs/guides/migration/versions)

## Checklist
- [x] The page ~has been added to `website/sidebars.js`~ is within a category listed within `website/sidebars.js`

### DIFF
--- a/website/dbt-versions.js
+++ b/website/dbt-versions.js
@@ -28,7 +28,7 @@ exports.versionedPages = [
     "firstVersion": "1.2",
   },
   {
-    "page": "docs/guides/migration-guide/upgrading-to-v1.2",
+    "page": "guides/migration/versions/upgrading-to-v1.2",
     "firstVersion": "1.2",
   },
   {
@@ -36,7 +36,7 @@ exports.versionedPages = [
     "firstVersion": "1.1",
   },
   {
-    "page": "docs/guides/migration-guide/upgrading-to-v1.1",
+    "page": "guides/migration/versions/upgrading-to-v1.1",
     "firstVersion": "1.1",
   },
   {
--- a/website/dbt-versions.js
+++ b/website/dbt-versions.js
@@ -2,7 +2,7 @@ exports.versions = [
   {
     version: "1.2",
     EOLDate: "2023-07-15",  // TODO estimated for now
-    isPrerelease: True,
+    isPrerelease: true,
   },
   {
     version: "1.1",
--- a/website/dbt-versions.js
+++ b/website/dbt-versions.js
@@ -1,7 +1,12 @@
 exports.versions = [
   {
+    version: "1.2",
+    EOLDate: "2023-07-15",  // TODO estimated for now
+    isPrerelease: True,
+  },
+  {
     version: "1.1",
-    EOLDate: "2023-04-28",  // TODO estimated for now
+    EOLDate: "2023-04-28",
   },
   {
     version: "1.0",
@@ -18,6 +23,10 @@ exports.versions = [
 ]
 
 exports.versionedPages = [
+  {
+    "page": "docs/guides/migration-guide/upgrading-to-v1.2",
+    "firstVersion": "1.2",
+  },
   {
     "page": "docs/contributing/testing-a-new-adapter",
     "firstVersion": "1.1",
--- a/website/dbt-versions.js
+++ b/website/dbt-versions.js
@@ -24,6 +24,10 @@ exports.versions = [
 
 exports.versionedPages = [
   {
+    "page": "docs/reference/dbt-jinja-functions/cross-database-macros",
+    "firstVersion": "1.2",
+  },
+  {
     "page": "docs/guides/migration-guide/upgrading-to-v1.2",
     "firstVersion": "1.2",
   },
--- a/website/dbt-versions.js
+++ b/website/dbt-versions.js
@@ -28,7 +28,7 @@ exports.versionedPages = [
     "firstVersion": "1.2",
   },
   {
-    "page": "docs/guides/migration-guide/upgrading-to-v1.2",
+    "page": "guides/migration/versions/upgrading-to-v1.2",
     "firstVersion": "1.2",
   },
   {
--- a/website/docs/guides/migration/versions/upgrading-to-v1.2.md
+++ b/website/docs/guides/migration/versions/upgrading-to-v1.2.md
@@ -1,5 +1,5 @@
 ---
-title: "Upgrading to v1.2 (latest)"
+title: "Upgrading to v1.2 (prerelease)"
 ---
 ### Resources
 
@@ -14,6 +14,10 @@ dbt Core v1.2 will soon be available as a **beta prerelease.** Join the #dbt-pre
 ## Breaking changes
 
 There are no breaking changes for end users of dbt. We are committed to providing backwards compatibility for all versions 1.x. If you encounter an error upon upgrading, please let us know by [opening an issue](https://github.com/dbt-labs/dbt-core/issues/new).
+
+### For maintainers of adapter plugins
+
+We added a collection of ["cross-database macros"](cross-database-macros) to dbt Core v1.2. Default implementations are automatically inherited by adapters and included in the testing suite. Adapter maintainers may need to override the implementation of one or more macros to align with database-specific syntax or optimize performance. For details on the testing suite, see: ["Testing a new adapter"](testing-a-new-adapter).
 
 ## New and changed documentation
 
--- a/website/docs/guides/migration/versions/upgrading-to-v1.2.md
+++ b/website/docs/guides/migration/versions/upgrading-to-v1.2.md
@@ -1,0 +1,22 @@
+---
+title: "Upgrading to v1.2 (latest)"
+---
+### Resources
+
+- [Changelog](https://github.com/dbt-labs/dbt-core/blob/main/CHANGELOG.md)
+- [CLI Installation guide](/dbt-cli/install/overview)
+- [Cloud upgrade guide](/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-choosing-a-dbt-version)
+
+:::info Beta
+dbt Core v1.2 will soon be available as a **beta prerelease.** Join the #dbt-prereleases channel in the Community Slack so that you're the first to know!
+:::
+
+## Breaking changes
+
+There are no breaking changes for end users of dbt. We are committed to providing backwards compatibility for all versions 1.x. If you encounter an error upon upgrading, please let us know by [opening an issue](https://github.com/dbt-labs/dbt-core/issues/new).
+
+## New and changed documentation
+
+_Under construction_
+
+https://github.com/dbt-labs/docs.getdbt.com/labels/dbt-core%20v1.2
--- a/website/docs/reference/dbt-jinja-functions/cross-database-macros.md
+++ b/website/docs/reference/dbt-jinja-functions/cross-database-macros.md
@@ -1,6 +1,6 @@
 ---
-title: "dbt"
-id: "dbt"
+title: "cross-database macros"
+id: "cross-database-macros"
 ---
 
 # Overview
@@ -68,7 +68,6 @@ compatibility. In general, you should not need these macros in your own dbt proj
 - [dbt.date_trunc](#date_trunc)
 - [dbt.last_day](#last_day)
 
-
 # Set functions
 
 ## except
@@ -110,7 +109,7 @@ __Args__:
 ## concat
 __Args__:
 
- * `fields`: Jinja array of [attribute names or expressions](#sql_expressions).
+ * `fields`: Jinja array of [attribute names or expressions](#sql-expressions).
 
 This macro combines a list of strings together.
 
@@ -130,9 +129,9 @@ This macro combines a list of strings together.
 ## hash
 __Args__:
 
- * `field`: [attribute name or expression](#sql_expressions).
+ * `field`: [attribute name or expression](#sql-expressions).
 
-This macro provides a hash (such as [MD5](https://en.wikipedia.org/wiki/MD5)) of an [expression](#sql_expressions) cast as a string.
+This macro provides a hash (such as [MD5](https://en.wikipedia.org/wiki/MD5)) of an [expression](#sql-expressions) cast as a string.
 
 **Usage**:
 
@@ -148,7 +147,7 @@ This macro provides a hash (such as [MD5](https://en.wikipedia.org/wiki/MD5)) of
 ## length
 __Args__:
 
- * `expression`: string [expression](#sql_expressions).
+ * `expression`: string [expression](#sql-expressions).
 
 
 This macro calculates the number of characters in a string.
@@ -166,8 +165,8 @@ This macro calculates the number of characters in a string.
 ## position
 __Args__:
 
- * `substring_text`: [attribute name or expression](#sql_expressions).
- * `string_text`: [attribute name or expression](#sql_expressions).
+ * `substring_text`: [attribute name or expression](#sql-expressions).
+ * `string_text`: [attribute name or expression](#sql-expressions).
 
 This macro searches for the first occurrence of `substring_text` within `string_text` and returns the 1-based position if found.
 
@@ -185,9 +184,9 @@ This macro searches for the first occurrence of `substring_text` within `string_
 ## replace
 __Args__:
 
- * `field`: [attribute name or expression](#sql_expressions).
- * `old_chars`: [attribute name or expression](#sql_expressions).
- * `new_chars`: [attribute name or expression](#sql_expressions).
+ * `field`: [attribute name or expression](#sql-expressions).
+ * `old_chars`: [attribute name or expression](#sql-expressions).
+ * `new_chars`: [attribute name or expression](#sql-expressions).
 
 This macro updates a string and replaces all occurrences of one substring with another. The precise behavior may vary slightly from one adapter to another.
 
@@ -205,8 +204,8 @@ This macro updates a string and replaces all occurrences of one substring with a
 ## right
 __Args__:
 
- * `string_text`: [attribute name or expression](#sql_expressions).
- * `length_expression`: numeric [expression](#sql_expressions).
+ * `string_text`: [attribute name or expression](#sql-expressions).
+ * `length_expression`: numeric [expression](#sql-expressions).
 
 This macro returns the N rightmost characters from a string.
 
@@ -221,13 +220,12 @@ This macro returns the N rightmost characters from a string.
 
 </File>
 
-
 # Aggregate and window functions
 
 ## any_value
 __Args__:
 
- * `expression`: an [expression](#sql_expressions).
+ * `expression`: an [expression](#sql-expressions).
 
 This macro returns some value of the expression from the group. The selected value is non-deterministic (rather than random).
 
@@ -244,7 +242,7 @@ This macro returns some value of the expression from the group. The selected val
 ## bool_or
 __Args__:
 
- * `expression`: [attribute name or expression](#sql_expressions).
+ * `expression`: [attribute name or expression](#sql-expressions).
 
 This macro returns the logical `OR` of all non-`NULL` expressions -- `true` if at least one record in the group evaluates to `true`.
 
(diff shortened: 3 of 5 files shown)


website/docs/guides/migration/versions/upgrading-to-v1.2.md
```diff
@@ -0,0 +1,26 @@
...
+dbt Core v1.2 will soon be available as a **beta prerelease.** Join the #dbt-prereleases channel in the Community Slack so that you're the first to know!
+:::
+
+## Breaking changes
+
+There are no breaking changes for end users of dbt. We are committed to providing backwards compatibility for all versions 1.x. If you encounter an error upon upgrading, please let us know by [opening an issue](https://github.com/dbt-labs/dbt-core/issues/new).
+
+### For maintainers of adapter plugins
+
+We added a collection of ["cross-database macros"](cross-database-macros) to dbt Core v1.2. Default implementations are automatically inherited by adapters and included in the testing suite. Adapter maintainers may need to override the implementation of one or more macros to align with database-specific syntax or optimize performance. For details on the testing suite, see: ["Testing a new adapter"](testing-a-new-adapter).
```
jtcohen6, line 20: Let's also reflect this in https://docs.getdbt.com/docs/contributing/building-a-new-adapter (or wherever you + @dataders feel is appropriate).

I agree that we shouldn't re-list every macro there, but we should at least link to the list and encourage adapter authors/maintainers to add support for them, and test them accordingly. In fact, our recommendation to adapter maintainers should probably look like:
- Inherit + run all `utility` tests, [like so](https://github.com/dbt-labs/dbt-redshift/blob/main/tests/functional/adapter/test_utils.py)
- For any failing tests, where the default implementation does not work, first try to reimplement the macro
- If it's not possible to get the test passing with a custom macro: mark the test to skip/xfail. Either the database doesn't support that functionality, or the core test has been written in a way that's not extensible.

website/dbt-versions.js
```diff
@@ -18,6 +23,14 @@ exports.versions = [
...
 
 exports.versionedPages = [
+  {
+    "page": "docs/reference/dbt-jinja-functions/cross-database-macros",
+    "firstVersion": "1.2",
+  },
+  {
+    "page": "docs/guides/migration-guide/upgrading-to-v1.2",
+    "firstVersion": "1.2",
+  },
```
jtcohen6: 👍 This looks right to me. I find that the versioning logic can be a bit wonky with browser settings/cookies (??) while I'm actively editing the docs site. It tends to work out in prod. _(famous last words)_

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,537 @@
+---
+title: "cross-database macros"
+id: "cross-database-macros"
+---
+
+# Overview
+
+These macros make it easier for package authors (especially those writing modeling packages) to implement cross-database
+compatibility. In general, you should not need these macros in your own dbt project (unless it is a package or you otherwise desire cross-database compatibility).
```
jtcohen6: 👍 Glad to see this context here!

The way I'd put this in my head:
- If you maintain a package, by using these macros, rather than a specific database's SQL syntax, your package is more likely to work on other adapters.
- If you maintain an adapter, by implementing (and testing) these macros, your adapter is more likely to support more packages.
- If you're an end user, more packages and adapters are likely to "just work" for you, without you having to do anything.
- Win-win!

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,537 @@
...
+
+## All functions (alphabetical)
+- [dbt.any_value](#any_value)
+- [dbt.bool_or](#bool_or)
+- [dbt.cast_bool_to_text](#cast_bool_to_text)
+- [dbt.concat](#concat)
+- [dbt.current_timestamp](#current_timestamp)
+- [dbt.current_timestamp_in_utc](#current_timestamp_in_utc)
```
jtcohen6: Let's swing back for `type_*` macros (https://github.com/dbt-labs/dbt-utils/issues/598) + `current_timestamp` macros (https://github.com/dbt-labs/dbt-utils/issues/599), once those are ready. (Might be in for v1.2, might be for a future release.) In the meantime, I don't think those need to block the initialization of this page / merging this PR.

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,537 @@
...
+## Aggregate and window functions
+- [dbt.any_value](#any_value)
+- [dbt.bool_or](#bool_or)
+- [dbt.listagg](#listagg)
+
+## Cast functions
+- [dbt.cast_bool_to_text](#cast_bool_to_text)
+- [dbt.safe_cast](#safe_cast)
+
+## Non-SQL Jinja functions
```
jtcohen6: Are these really "Non-SQL"? I thought these macros existed to produce SQL expressions for escaped strings and string literals

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,537 @@
...
+- [dbt.length](#length)
+- [dbt.listagg](#listagg)
+- [dbt.position](#position)
+- [dbt.replace](#replace)
+- [dbt.right](#right)
+- [dbt.safe_cast](#safe_cast)
+- [dbt.split_part](#split_part)
+- [dbt.string_literal](#string_literal)
+
+## Set functions
```
jtcohen6: Let's add links to the anchors down below for each category. That probably requires changing this header, though, to avoid conflicting anchors:
```suggestion
[**Set functions**](#set-functions)
```

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,537 @@
...
+ * None
+
+`except` is one of the set operators specified ANSI SQL-92 (along with `union` and `intersect`) and is akin to [set difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement).
+
+**Usage**:
+
+<File name='models/example.sql'>
+
+```sql
+{{ dbt.except() }}
```
jtcohen6: Maybe we should include _just_ the default "output" for each of these macros? In this case, it's hard to tell from just looking at the macro that all it does is return the string `except` (or database-specific equivalent)

jtcohen6 (commented):
Fantastic work here @dbeatty10! I left a handful of comments (mostly structural), and responded to your questions below. I didn't appreciate that most of these macros were effectively undocumented in `dbt-utils`, and so this is the first time we have real docs for them. Nice work!

(Food for far-future thought: Is there a better way for us to embed macro/method descriptions within dbt, so we don't need to hand-roll these docs every time?)

I'm for shipping this PR as soon as possible, and then swinging back with updates once we tackle the `type_*` + `current_timestamp` macros. It would be better to get the v1.2 docs live sooner.

### questions

> The content is currently within Jinja Reference > dbt Jinja functions > cross-database macros. Is this where we want it? Or somewhere else?

This feels fine to me

> The first draft contains both an alphabetical listing and grouped into categories. Do we favor one over the other?

I agree with categorical grouping as the primary point of entry. I think an alphabetical list of _every_ dbt-Jinja function is also warranted, but we may want that to live on its own separate page. Not a problem we must solve in this PR.

> Single page listing all macros? Or separate pages based on functional category?

From my point of view, as long as there are clear anchors (ways to link to specific categories + macros), one single page is an okay way to start. We can always split them up later on if/when the page gets unwieldy.

> Include dbt. prefix for macros? Or drop?

I prefer without the prefix, too.

> Convert to a SQL query that can actually execute?

I agree this gets complicated if we want the query to work across databases. I know this has given us a hard time for the macros documented in the `dbt-utils` README. Out of scope for this PR, I'd say.

> Include the filename wrappers (i.e., `models/example.sql`)?

I hear you on, if they aren't real/full SQL, it shouldn't be a full model file. I don't have such a strong feeling here. Let's keep it more generic for now.

> Multiple variations per example? Or just a single variation per example?

I buy what you're saying — multiple variations is good by me!

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
dbeatty10: Maybe you can help come up with a more accurate name?

Your understanding is correct, and this first-draft description is somewhat of a misnomer. All of these macros _are_ used to help build up executable SQL.

### Intent
I was trying to differentiate _where_ the actual escaping or literal-ing will happen for these two macros (pre-SQL via Jinja or SQL via database):
- `escape_single_quotes`
- `string_literal`

### Why this is hard to think about
So many cross-data macros are pretty meta (since they render _other_ code that will actually do the thing described (like `pivot`, etc).

These particular macros aren't meta -- they actually do what they say they will do _directly_, without passing along the actual work to the database.

website/dbt-versions.js
dbeatty10: Every great figure deserves some infamous words, and maybe these can be yours!

[Here's](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.1) an example of a page in production with an entry within [dbt-versions.js](https://github.com/dbt-labs/docs.getdbt.com/blob/fe4ec08639c00f7034064dc2b0f2ea924a7f6bdb/website/dbt-versions.js#L38-L41).

When I choose v1.0 in the drop-down, it doesn't change the content:
- image

Other pages do work for me, like [this](https://docs.getdbt.com/reference/dbt-jinja-functions/selected_resources) one:

- image

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
jtcohen6: > pre-SQL via Jinja or SQL via database

Ohhhhh. I see what you mean. Hm, maybe like "string literal" functions (as distinct from "string functions")? Makes me think of the subtleties around `try_cast` in Snowflake, which supports string literal inputs _only_

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
jtcohen6: Might also help adapter authors see "at a glance" if they need to reimplement a given macro (though starting by inheriting the functional tests and seeing what fails is definitely still the Right Way To Go)

website/dbt-versions.js
```diff
@@ -18,6 +23,14 @@ exports.versions = [
 ]
 
 exports.versionedPages = [
+  {
+    "page": "docs/reference/dbt-jinja-functions/cross-database-macros",
+    "firstVersion": "1.2",
+  },
+  {
+    "page": "docs/guides/migration-guide/upgrading-to-v1.2",
```
runleonarun: I think we need to change this a little:
```suggestion
    "page": "guides/migration/versions/upgrading-to-v1.2",
```
I'll merge and see if it works!

website/dbt-versions.js
runleonarun: Changing to `"page": "guides/migration/versions/upgrading-to-v1.2",` to see if that works.

website/sidebars.js
```diff
@@ -656,6 +656,7 @@ const sidebarSettings = {
             slug: '/guides/migration/versions',
           },
           items: [
```
runleonarun, line 658: @jtcohen6 and @dbeatty10 what do you think about autogenerating this sidebar? Then we don't need to update sidebar.js every release. It would looks something [like this](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/sidebars.js#L493-L508).

website/dbt-versions.js
runleonarun: @jtcohen6 and @dbeatty10 My fix above worked, so I fixed the v1.1 page as well. Thanks for flagging!

dbeatty10: Woohoo! Thank you, @runleonarun ! 🏆

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
dbeatty10: Included the default output from the Postgres adapter for each of these macros.

Is it pretty in all cases? Nope.

We've got options for how to handle that:
- update the whitespace handling on a per-macro basis and re-generate the example output
- format the example output however we want it to look even if it is not the exact output

Do you have a preference @jtcohen6 ?

Here's the input file I used:
<details>
  <summary>/analyses/output.sql</summary>

```sql

{{ except() }}

{{ intersect() }}

{{ concat(["column_1", "column_2"]) }}
{{ concat(["year_column", "'-'" , "month_column", "'-'" , "day_column"]) }}
{{ concat(["first_part_column", "'.'" , "second_part_column"]) }}
{{ concat(["first_part_column", "','" , "second_part_column"]) }}

{{ hash("column") }}
{{ hash("'Pennsylvania'") }}

{{ length("column") }}

{{ position("substring_column", "text_column") }}
{{ position("'-'", "text_column") }}

{{ replace("string_text_column", "old_chars_column", "new_chars_column") }}
{{ replace("string_text_column", "'-'", "'_'") }}

{{ right("string_text_column", "length_column") }}
{{ right("string_text_column", "3") }}

{{ escape_single_quotes("they're") }}
{{ escape_single_quotes("ain't ain't a word") }}

select {{ string_literal("Pennsylvania") }}

{{ any_value("column_name") }}

{{ bool_or("boolean_column") }}
{{ bool_or("integer_column = 3") }}
{{ bool_or("string_column = 'Pennsylvania'") }}
{{ bool_or("column1 = column2") }}

{{ listagg(measure="column_to_agg", delimiter_text="','", order_by_clause="order by order_by_column", limit_num=10) }}

{{ cast_bool_to_text("boolean_column_name") }}
{{ cast_bool_to_text("false") }}
{{ cast_bool_to_text("true") }}
{{ cast_bool_to_text("0 = 1") }}
{{ cast_bool_to_text("1 = 1") }}
{{ cast_bool_to_text("null") }}

{{ safe_cast("column_1", api.Column.translate_type("string")) }}
{{ safe_cast("column_2", api.Column.translate_type("integer")) }}
{{ safe_cast("'2016-03-09'", api.Column.translate_type("date")) }}

{{ dateadd(datepart="day", interval=1, from_date_or_timestamp="'2016-03-09'") }}
{{ dateadd(datepart="month", interval=-2, from_date_or_timestamp="'2016-03-09'") }}

{{ datediff("column_1", "column_2", "day") }}
{{ datediff("column", "'2016-03-09'", "month") }}
{{ datediff("'2016-03-09'", "column", "year") }}

{{ date_trunc("day", "updated_at") }}
{{ date_trunc("month", "updated_at") }}
{{ date_trunc("year", "'2016-03-09'") }}

{{ last_day("created_at", "month") }}
{{ last_day("'2016-03-09'", "year") }}
```
</details>

website/sidebars.js
dbeatty10: I don't have a preference. Whatever you both think is best.

dbeatty10:
@jtcohen6 I _think_ I've addressed all the outstanding feedback. Will you take a look?

website/sidebars.js
jtcohen6: I like the idea of auto-generating! I don't think that needs to block this PR if it's work we can pick up separately.

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
jtcohen6: @dbeatty10 Genius move to use dbt to quickly & efficiently get the compiled output! Let's not talk about how I was thinking of doing this......

I think it's okay to clean up _a little_ for display in the docs. I think it's also okay if it's a bit ugly (truth in advertising). All the more reason to motivate us doing The Right Thing, which is definitely this:
> update the whitespace handling on a per-macro basis and re-generate the example output

Great opportunities for community contributors!

How will we keep this up to date going forward? Short of... the semantic layer? :)

website/docs/guides/migration/versions/upgrading-to-v1.2.md
jtcohen6, line 20: @dataders Could you take a note for updating the adapter documentation, to notify adapter maintainers that there are:
- more tests to start running ("utils")
- likely more macros that need reimplementing

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
```diff
@@ -0,0 +1,670 @@
...
+
+`except` is one of the set operators specified ANSI SQL-92 (along with `union` and `intersect`) and is akin to [set difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement).
+
+**Usage**:
+
+```sql
+{{ except() }}
+```
+
+**Default Output**:
```
jtcohen6: Nit: Since this is actually the output we generated using Postgres, maybe this should say `**Sample Output (PostgreSQL)**` instead of `**Default Output**`?

jtcohen6 (approved):
Fit to ship! Left two (non-blocking) comments. One's just a nit, the other is a TODO for updating the docs targeting adapter maintainers, which can happen in a subsequent PR.

website/docs/guides/migration/versions/upgrading-to-v1.2.md
jtcohen6: Update: opened https://github.com/dbt-labs/docs.getdbt.com/issues/1634 to track this work

website/docs/reference/dbt-jinja-functions/cross-database-macros.md
dbeatty10: A handful of ideas for keeping up-to-date going forward (generally from most manual to most automated):
1. Use the same strategy as this PR:
    - `dbt init` a dbt project
    - copy-paste `/analyses/output.sql` into the project
    - execute `dbt compile --select output`
    - copy-paste from output in `target` directory into each relevant section of the docs
1. Create a git repo for the dbt project I used. Put instructions for generating the sample output in the README.
1. Build an elaborate mechanism that is mostly hands-off
1. Throw up our hands and declare the current examples in the docs are "good enough" and leave them as-is.

Option 2 (or 4!) might hit the sweet spot for us for the time being.

PR #1630, opened 2022-06-28
Update setup-azure.md

mirnawong1:
hi @schottj @jschottenstein - wanted to check with you if it was ok to add padded additional clarification that the user should copy the value field in azure devops and not the secret id (which i thought when i went through the this loom video https://www.loom.com/share/7d35434c554f438d99e2637f5ae7c42b - thank you!

## Prerelease docs

## Checklist


dataders (approved):
very helpful clarification @mirnawong1. Azure AD is certainly a rabbit hole, so anything that helps keep folks from falling into it is more than welcome!

PR #1637, opened 2022-06-29
Add onetrust cookie settings link to footer

JKarlavige:
## Description & motivation
 This task is to have the Docs Cookie Settings link in the footer styled correctly and ready to merge live when OneTrust is taken live.

This will not be functional until OneTrust is active. We are only checking to make sure the styles look correct for this PR.

Once OneTrust is live, this link will trigger the OneTrust Preference Center, similar to getdbt.com


website/src/css/custom.css
```diff
@@ -830,3 +830,24 @@ code {
...
+  background: none;
+  border: none;
+  color: var(--ifm-footer-color);
+  text-decoration: none;
+  font-weight: var(--ifm-font-weight-bold);
+  font-family: var(--ifm-font-family-base);
+  padding: 0;
+  line-height: inherit;
+  font-size: 1rem;
+}
```
john-rock: Should we add add `cursor:pointer` to this?

john-rock (commented):
Just one comment on adding a css property. Everything else looks good!

website/src/css/custom.css
JKarlavige: Yeah good call, will add that in

john-rock (approved):
lgtm!

runleonarun:
@JKarlavige and @john-rock I updated the GitHub Action to have a much higher threshold for the spammy comment. Sorry about that!

PR #1641, opened 2022-06-30
Foundation for Grants as configs (core 1.2 feature)

runleonarun:
Closes https://github.com/dbt-labs/docs.getdbt.com/issues/1527

## Description & motivation

This provides the foundation for the "Grants as configs" feature. The content is [in this PR](https://github.com/dbt-labs/docs.getdbt.com/pull/1642) for easier reviewing.

## Prerelease docs
If this change is related to functionality in a prerelease version of dbt (delete if not applicable):
- [x] I've added versioning components, as described in ["Versioning Docs"](../contributing/versioningdocs.md)
- [x] I've added a note to the prerelease version's [Migration Guide](../website/docs/docs/guides/migration-guide)

## Checklist
If you added new pages (delete if not applicable):
- [x] The page has been added to `website/sidebars.js`
- [x] The new page has a unique filename

If you removed existing pages (delete if not applicable):
- [ ] The page has been removed from `website/sidebars.js`
- [ ] An entry has been added to `_redirects`

PR #1642, opened 2022-06-30
Grants as configs - Core v1.2

### LINKED ISSUE #1527: `grants` as configs, run during materializations
### Contributions

- [X] I am a PM or subject matter expert at dbt who is responsible for this feature.

### Where does this content belong?

**Main location:** New page in "general configs," similar to the page for [`meta`](https://docs.getdbt.com/reference/resource-configs/meta)

Update copy in:
- [Hooks & operations](https://docs.getdbt.com/docs/building-a-dbt-project/hooks-operations), specifically the recommendations + examples that these are for granting privileges. We want to turn these hooks into an "advanced" feature, for people who know what they're doing, since there are a lot of gotchas.
- Snowflake `copy grants` config: https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants
- ["Building a new adapter"](https://docs.getdbt.com/docs/contributing/building-a-new-adapter): there are 4 new macros, and adapter maintainers may need to reimplement anywhere between none and all of them to support `grants` on their database

Longer term: We should update ["Creating new materializations"](https://docs.getdbt.com/guides/legacy/creating-new-materializations), since features like `persist_docs` + `grants` are not reflected there currently.

## Who will use this feature?

dbt developers who want to manage access to the valuable datasets they're producing with dbt. Oftentimes, that means _restricting_ access on upstream/intermediate models, and granting access to reporting / BI / downstream querying tools on final models that are ready for primetime.

## Feature description: how it will help people

Traditionally, the way you had to do this was pretty jank:
```sql
{{ config(post_hook = 'grant select on {{ this }} to role reporter') }}
```
(See those nested curlies? [Ew!](https://docs.getdbt.com/docs/building-a-dbt-project/dont-nest-your-curlies#an-exception))

This feature will enable people to define grants as configs—the way they know & love—and dbt takes care of the rest.

## Customer research

Hooks applying grants ends up being some of the gnarliest code in advanced/mature dbt projects. Yet, people have A Bad Time using hooks. See the issues linked from https://github.com/dbt-labs/dbt-core/issues/5090.

More background for dbt Labs employees: https://www.notion.so/dbtlabs/Grants-permissions-becbe54d4a444ede9baf46078e5e69ef

## How can we test or use this feature?

See the example code below! Once the feature is merged, and available from source / in a beta, it should all Just Work on any of our supported databases.

## Source material

- https://github.com/dbt-labs/dbt-core/issues/5189
- https://github.com/dbt-labs/dbt-core/issues/5263

---

### Configuring the grant

Grants will be supported as node configs, in all the usual places:
```yml
# dbt_project.yml
models:
  export:
    +grants:
      select: ['reporter', 'bi']
```
```yml
# models/model.yml
models:
  - name: specific_model
    config:
      grants:
        select: ['reporter', 'bi']
```
```sql
-- models/specific_model.sql

{{ config(grants = {'select': ['other_user']}) }}
```

---

Per-adapter behavior: Most databases support a simple privilege named `select`, and recipients (users/groups/roles) with plaintext names. **BigQuery** is a little different for both naming conventions. There, this needs to be:
```yml
# models/model.yml
models:
  - name: specific_model
    config:
      grants:
        'roles/bigquery.dataViewer': ['user:jeremy@example.com']
```

We should just link to these docs: https://cloud.google.com/bigquery/docs/access-control

---

**There's one special behavior that we're experimenting with.** This will be supported for `grants` only to get started. If a user has configured some `grants` for all models in their project (in `dbt_project.yml`), and then they reconfigure certain `grants` for a specific model, the default behavior is to "clobber" one config with another: to completely replace the project-level grants with the model-level grants.

We're optionally supporting a `+` prefix (different from [the other "+" prefix that everyone hates](https://docs.getdbt.com/reference/resource-configs/plus-prefix)) to make that model-level config "additive" instead of "clobbering". E.g.:

```yml
# dbt_project.yml
models:
  +grants:
    select: ['reporter', 'bi']
```
```sql
-- models/specific_model.sql

{{ config(grants = {'+select': ['other_user']}) }}
```

This would "add" to the `select` list, instead of replacing it, and thereby return `select: ['reporter', 'bi', 'other_user']`. More conversation about why this is valuable, and speculation about extending it to other config types, in https://github.com/dbt-labs/dbt-core/issues/4108.

---

### Runtime behavior

A lot of this doesn't need documentation, since it's something that dbt handles behind the scenes. But there will be some opinionated judgment calls that we want to make clear.

dbt's goal in this feature is to always ensure that the grants actually applied on the database object match up exactly with the `grants` that the user has configured. No more, no less.

Depending on the database adapter, and the materialization of the resource, dbt may or may not need to do more complicated work: figure out grants on the preexisting object, and figure out if any have been added/removed. 

On Snowflake, if `copy_grants` is enabled, then dbt will NOT try to do this work. The presumption is that the end user knows what they're doing, and may be applying grants in an external system. dbt will ONLY add grants, not remove them.

If people need their `grants` to do other fancier things, they can always revert back to the current recommendation, of using a pre/post hook.

runleonarun:
## Description & motivation

Partially addresses #1527 

Writing content for "Grants as configs" feature in v1.2. Make sure to merge it into https://github.com/dbt-labs/docs.getdbt.com/pull/1641 and merge that PR too!

## To-do before merge
- [ ] Merge this PR into https://github.com/dbt-labs/docs.getdbt.com/pull/1641 and then merge #1641 for release.
- [ ] Add a new simpler example to hooks & operations
- [ ] Snowflake copy grants config: https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants
- [ ] ["Building a new adapter"](https://docs.getdbt.com/docs/contributing/building-a-new-adapter): there are 4 new macros, and adapter maintainers may need to reimplement anywhere between none and all of them to support grants on their database

## Prerelease docs
- [x] I've added versioning components, as described in ["Versioning Docs"](../contributing/versioningdocs.md)
- [x] I've added a note to the prerelease version's [Migration Guide](../website/docs/docs/guides/migration-guide)
- [x] Make sure to version the Grants.md page for 1.2!

## Checklist


website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -36,6 +37,12 @@ Hooks are snippets of SQL that are executed at different times:
 
 Hooks are defined in your `dbt_project.yml` file. Pre- and post-hooks can also be defined in a `config` block.
 
+<VersionBlock firstVersion="1.2">
+
+In order to streamline hooks and automatically apply grants when your dbt model runs, we recommend using [`grants` resource-config](/reference/resource-configs/grants).  
+### Advanced example for hooks
+
+</VersionBlock>
```
jtcohen6: 👍

- I think we should remove the examples related to `grants` below, if someone's selected v1.2+. We could add an example with more complex privilege granting that won't be handled by the built-in feature
- Left some more pre/post-hook examples in [our Slack thread](https://dbt-labs.slack.com/archives/C03J5GFKDDJ/p1656604642649699)

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,66 @@
...
-resource_types: all
-datatype: "{<dictionary>}"
+resource_types: [models,seeds,snapshots]
+datatype: "string"
 default_value: {}
 ---
 
+<Snippet src="available-prerelease-beta-banner" />
+
+You can manage access to the datasets you're producing with dbt by using grants. Define these grants as resource configs on each model, seed, or snapshot instead of using hooks to implement these permissions. You can define default grants in your `dbt_project.yml` and provide model-specific grants by defining them within each model's SQL or YAML file.
```
jtcohen6: > instead of using hooks

This will make sense to anyone who's been using dbt for a while, but it might be confusing to a brand-new user. (Going forward, we're framing pre/post-hooks as an advanced, "only if you need it" feature.)

Maybe a separate note, in the migration guide, to the effect of: "If you already use post-hook to apply simple grants, moving to the `grants` feature will allow you to DRY up your duplicated/boilerplate code."

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,66 @@
...
+
+
+
+The grant resource configs enable you to automatically apply grants when your dbt model runs. If you want to create even more granular permissions, you can still use grants with hooks, but these hooks can be complicated to implement and require testing thoroughly to make sure they're working as expected. For more information on hooks, see [Hooks & operations](/building-a-dbt-project/hooks-operations).
```
jtcohen6: I agree that linking to the `hook-operations` page is still appropriate, and calling out that it's necessary for use cases more advanced than what's shown on this page.

I think we can draw a clearer delineation between what we view as "simple" and "advanced" use cases for grants/permissions:
- **Simple:** grant privileges on this model to a set of recipients, as soon as the model finishes building
- **Advanced:** [future grants](https://docs.snowflake.com/en/sql-reference/sql/grant-privilege.html), more granular row- and column-level access/masking policies

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +69,33 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
 snapshots:
   - name: snapshot_name
-    config:
-      meta: {<dictionary>}
-
-    columns:
-      - name: column_name
-        meta: {<dictionary>}
-
+      grants:
```
jtcohen6, line 83: This needs to be nested under a `config` block:
```yml
snapshots:
  - name: snapshot_name
    config:
      grants:
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,66 @@
...
+        select: ['reporter', 'bi']
 ```
 
 </File>
 
-The `meta` config can also be defined:
+The `grants` config can also be defined:
```
jtcohen6, line 48: There's a new capability we're experimenting with for grants ONLY, described [in the issue here](https://github.com/dbt-labs/docs.getdbt.com/issues/1527#:~:text=There%27s%20one%20special%20behavior%20that%20we%27re%20experimenting%20with.).

If users set `grants` in multiple places, e.g. `dbt_project.yml` _and_ in a per-model `.sql`/`.yml` file, the default behavior for grants on a given privilege (e.g. `select`) is to replace the less-specific set of recipients with the more-specific set of recipients.

So for instance:
```
# dbt_project.yml
models:
  +grants:
    select: ['reporter', 'bi']
```
```
-- models/specific_model.sql

{{ config(grants = {'select': ['other_user']}) }}
```
This will result in `specific_model` granting the `select` privilege to ONLY `other_user`.

However, let's say we wanted to ADD `other_user` to the existing list, rather than replacing that list entirely. To accomplish that, we can use a `+` ("addition") symbol, prefixing the name of the privilege:
```
-- models/specific_model.sql

{{ config(grants = {'+select': ['other_user']}) }}
```
Now, the model will grant `select` to `reporter`, `bi`, and `other_user`!

**Note:** This will only take effect for privileges with the `+` prefix. If we were granting other privileges, in addition to `select`, their clobber vs. addition behavior is independent.

**Note:** That second use of `+` needs to be disambiguated from the use of `+` in `dbt_project.yml` (first use shown), which is documented and described here: https://docs.getdbt.com/reference/resource-configs/plus-prefix

It's unfortunate that we already used `+` to do that in `dbt_project.yml`, when we could have picked a different symbol. (They're also distinct from [`+` used in node selection syntax](https://docs.getdbt.com/reference/node-selection/graph-operators#the-plus-operator), though that's contextually quite different.) But, `+` felt like the only natural choice for the "additive config" functionality.

website/sidebars.js
```diff
@@ -340,6 +340,7 @@ const sidebarSettings = {
             "reference/resource-configs/database",
             "reference/resource-configs/enabled",
             "reference/resource-configs/full_refresh",
+            "reference/resource-configs/grants",
```
jtcohen6, line 344: Do we want to version the whole page, too? Since it's totally unusable for anyone on v1.1 or older

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +69,33 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
-
-### Override one meta attribute for a single model
-
-<File name='models/my_model.yml'>
 
 ```sql
-{{ config(meta = {
-    'single_key': 'override'
+{{ config(materialized = 'incremental', grants = {
+    'select': ['other_user']
```
jtcohen6: As with most configs in dbt that accept lists (e.g. [`sort` keys on Redshift](https://docs.getdbt.com/reference/resource-configs/redshift-configs#using-sortkey-and-distkey)), if the user is specifying only a single entry in the list, they can also provide the argument as a string, and dbt will "listify" it for them:
```
'select': 'other_user'
```
(That is, without the `[]` brackets around `'other_user'`)

IMO it's always better/clearer to provide the argument as an explicit list, though!

(The right place to note this is wherever seems easiest — but I think it's worth noting _somewhere_, rather than leaving the behavior totally implicit / undocumented)

jtcohen6 (commented):
Thanks for putting the PR up @runleonarun! I left a first round of comments.

Let me know if there are pieces you want to discuss further; I appreciate that the `+` piece is pretty tricky. (Slack might be better this week, while I'm swimming in a sea of GitHub notifications.)

---

### Ignore the following — updated in more recent review comment below

Three **adapter-specific pieces** that have come up in our implementation work, which we'll need to document:
- **BigQuery:** ~Users will need to specify a `location` in their connection profile. (This isn't our choice; it's an implementation detail in GCP, because resource permissions are stored/accessed by region... I think. Not something we need to go into here.) That's totally doable ([docs](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile#dataset-locations)), it's just optional today, so many users probably don't. dbt will raise a helpful error message if they try to use `grants` without setting a `location`.~
- **Spark/Databricks:** ~Grants are currently only supported on the Databricks runtime, and not yet on OSS Apache Spark (because they're actually implemented in the metastore layer ... again, implementation details, not something we need to get into here). dbt won't try to raise an error message here, but they might see a database error to the effect of, `GRANT isn't a supported such-and-such`, [or similar](https://www.mail-archive.com/user@spark.apache.org/msg74276.html).~ When using Databricks interactive clusters, there is some setup that admins/users will need to do: https://docs.databricks.com/administration-guide/access-control/table-acl.html + https://docs.databricks.com/security/access-control/table-acls/table-acl.html
- **Snowflake:** ~We need to nail down the exact behavior / interaction with the [`copy_grants`](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants) config. I don't know if there's a way to avoid squirrelly language here: "When fully replacing objects, dbt will ignore any grants copied over, and not try to revoke any superfluous grants." Subtext is: If you're using this config, you know what you're doing.~ If users enable the [`copy_grants`](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants) config, dbt will take that into account when calculating which grants to add/remove.

Very open to your thoughts on where feels most appropriate to include that information. It's not high-level feature support, so adding a note to the adapter-specific profile/config pages feels like the wrong move. We really just want a note/callout that catches the attention of the relevant users, and is comfortably ignored by all others.

website/sidebars.js
runleonarun: Yes! This was an oversight. Thanks for the reminder!!

jtcohen6:
@runleonarun I wanted to share a top-level comment that came up in [a recent implementation discussion](https://dbt-labs.slack.com/archives/C03J5GFKDDJ/p1657118772825579?thread_ts=1657019275.987459&cid=C03J5GFKDDJ). I realize that I've done a bad job laying this out clearly. It really motivates our design, and should communicate in a sentence or two what users can expect to happen.

- If the `grants` config is not supplied, `none`, or empty (`{}`) --> do nothing. dbt will perform **zero** revoke or grant statements by its own volition. It assumes that you do not want dbt "managing" the grants for this model. (You can still specify whatever SQL you want in a post-hook, including `revoke` + `grant`.)
- If the `grants` config is supplied with _any_ values --> When dbt finishes running this model, **the actual grants on the model will match exactly what's specified in the `grants` config.** (dbt will find the most efficient path to make that a reality — checking what's there, removing some, adding others — and it may look different for different materializations/databases — but that simple statement is all users should need to think about.)

website/docs/reference/resource-configs/grants.md
runleonarun: OK Done. Let me know how that looks!

website/docs/reference/resource-configs/grants.md
runleonarun: This makes sense. I reframed this a bit.

website/docs/reference/resource-configs/grants.md
runleonarun: :thumbs_up:

website/docs/reference/resource-configs/grants.md
runleonarun: I tried to add both examples here, but I think you're right that it should be called out somewhere else. 

I'm not sure where yet, and we can follow up with this after 1.2 is released. I [captured it here](https://github.com/dbt-labs/docs.getdbt.com/issues/1686). Do we have a syntax or expressions page that lays out content/standards like this? (I can't find one, but that doesn't mean we don't have one)

website/docs/docs/building-a-dbt-project/hooks-operations.md
runleonarun: I removed grant examples for 1.2 and higher. Still need to add pre- and post-hook examples for 1.2+ .

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,117 @@ Effective database administration sometimes requires additional SQL statements t
...
+<WHCode>
+
+<div warehouse="BigQuery">
+
+<File name='models/<model_name>.sql'>
+
+```sql
+{{ config(
+    post_hook=[
+      "grant select on {{ this }} to role reporter"
```
jtcohen6: BigQuery has different names for its privileges + grantees. So if you're writing the hook SQL yourself, this actually needs to look like:
```sql
'grant `roles/bigquery.dataViewer` on {{ this.type }} {{ this }} to "user:someone@yourcompany.com"'
```
If you're using the new-in-1.2 grants config, it will look like:
```yml
{{ config(grants = {'roles/bigquery.dataViewer': ['user:someone@yourcompany.com']}) }}
```

**This is relevant to our `grants` docs, too.** We need to call out that, while our examples use words like `select` and `another_user` because they're common on many databases, users will need to be aware of their own database's support syntax for:
- **privileges** (= a thing you can do in the database)
- **grantees** (= the recipients of granted privileges, also called "principals") — which could be a user, a group of users, a role held by users (Snowflake), a service account (GCP), ... 

(Very confusing: Google uses the word "role" to mean "privilege," whereas Snowflake uses the word "role" to mean "grantee" = a way of receiving privileges. Let's just try avoiding the word "role" if possible!)

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,117 @@ Effective database administration sometimes requires additional SQL statements t
...
+</div>
+
+<div warehouse="Databricks">
+
+<File name='models/<model_name>.sql'>
+
+```sql
+{{ config(
+    post_hook=[
+      "grant select on {{ this }} to role reporter"
```
jtcohen6: On Databricks, users are identified by their logins (= email addresses):
```sql
"grant select on {{ this }} to `someone@yourcompany.com`"
```

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,117 @@ Effective database administration sometimes requires additional SQL statements t
...
+</div>
+
+<div warehouse="Redshift">
+
+<File name='models/<model_name>.sql'>
+
+```sql
+{{ config(
+    post_hook=[
+      "grant select on {{ this }} to role reporter"
```
jtcohen6: Redshift doesn't use the word `role` in the way it's appearing here. This should be just:
```sql
"grant select on {{ this }} to reporter"
```

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,117 @@ Effective database administration sometimes requires additional SQL statements t
...
+</div>
+
+<div warehouse="Snowflake">
+
+<File name='models/<model_name>.sql'>
+
+```sql
+{{ config(
+    post_hook=[
+      "grant select on {{ this }} to role reporter"
```
jtcohen6: The word `role` is fine here for Snowflake, but it's not strictly necessary. This also works:
```sql
"grant select on {{ this }} to reporter"
```

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,117 @@ Effective database administration sometimes requires additional SQL statements t
...
+
+You can use hooks to trigger actions at certain times when running an operation or building a model, seed, or snapshot.
+
+For more information about when hooks can be triggered, see reference sections for [`on-run-start` and `on-run-end` hooks](on-run-start-on-run-end) and [`pre-hook`s and `post-hook`s](pre-hook-post-hook).
+
+You can use hooks to provide database-specific functionality not available out-of-the-box with dbt. 
+
+For example, you can use a `config` block to run an `ALTER TABLE` statement right after building an individual model using a `post-hook`:
+
+<WHCode>
```
jtcohen6: I left some specific comments correcting the syntax on each warehouse, but I also want to be clear that all the examples below are _simple grant statements_, and we should be encouraging anyone on v1.2+ to use the `grants` config instead.

We should explicitly discourage them from using `post-hook` for simple grants. Instead, they should use `post-hook` if and only if:
- they need to apply grants in a more complex way, which the v1.2 `grants` config does not (yet) support
- they need to do other custom post-processing that dbt does not (yet) support via out-of-the-box configuration — e.g. [`analyze table`](https://spark.apache.org/docs/latest/sql-ref-syntax-aux-analyze-table.html), [`alter table set property`](https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-alter-table.html), [`alter table ... add row access policy`](https://docs.snowflake.com/en/sql-reference/sql/alter-table.html), etc

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +70,45 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
-### Override one meta attribute for a single model
-
-<File name='models/my_model.yml'>
+Granting multiple users the same permission:
 
 ```sql
-{{ config(meta = {
-    'single_key': 'override'
+{{ config(materialized = 'incremental', grants = {
+    'select': ['other_user','this_user']
```
jtcohen6: I find the name `'this_user'` to be a bit confusing, because it sounds like the current user that dbt is running as. We should be clear that `grants` are only for granting access to OTHER users/groups/roles

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +70,45 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
-
-Depending on the resource you're configuring, `meta` may be available within the `config` property, or as a top-level key. (For backwards compatibility, `meta` is always supported as a top-level key, though without the capabilities of config inheritance.)
-
-<Changelog>
 
-* `v0.16.0`: This property was introduced
-* `v0.21.0`: Introduced the `config` property, and gave `meta` the capabilities of a config
-
-</Changelog>
+You can use the `grants` field to set permissions or grants for a resource. These grants will be compiled into the `manifest.json` file complied by dbt, which you can view in the automatically generated documentation.
```
jtcohen6: > which you can view in the automatically generated documentation.

This isn't true (yet), so we shouldn't say it here. We could make it true with a little bit of HTML/CSS/JS in the `dbt-docs` site.

Everything up to this point, about compiling it into `manifest.json` / making it accessible as metadata, is true.

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,67 @@
...
+datatype: "string"
 default_value: {}
+id: "grants"
 ---
 
+<Snippet src="available-prerelease-beta-banner" />
+
+You can manage access to the datasets you're producing with dbt by using grants. To implement these permissions, define grants as resource configs on each model, seed, or snapshot. Define the default grants that apply to the entire project in your `dbt_project.yml`, and define model-specific grants within each model's SQL or YAML file.
+
+The grant resource configs enable you to apply permissions to a specific model and set of recipients as soon as the model finishes building.
```
jtcohen6: High-level message:

> As soon as your model, seed, or snapshot finishes building, dbt ensures that the grants on its view or table match **exactly** the `grants` you have configured.

> Behind the scenes, dbt will try to take the most efficient approach to updating grants. This approach will vary based on the adapter you're using, and whether dbt is replacing or updating an object that already exists. You can always check [the debug logs](https://docs.getdbt.com/reference/events-logging) for the full set of `grant` and `revoke` statements that dbt runs.

The language there is a bit clunky, but that's the big idea. As a user, you can trust in an exact match. dbt handles the nuance of the implementation.

jtcohen6 (commented):
Coming along!

### Per-adapter notes

Open to your thoughts on how to structure these, given that they're highly specific to users of specific databases.

#### Snowflake

- Interaction with [`copy_grants`](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#copying-grants) config: dbt WILL take this config into account when calculating which grants need to be added/removed.

#### BigQuery

- Grantee and privilege names are quite different (e.g. `roles/bigquery.dataViewer` instead of `select`, `user:jeremy@dbtlabs.com` instead of `jerco_user`). We need to call that out for BQ users.
- (!!) Interaction with [`grant_access_to` config](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#authorized-views): ZERO. These are completely different features / functionalities. The `grant_access_to` config is specifically for setting up **authorized views**. When configured, dbt gives an authorized view access to other datasets, without leaking additional data from those datasets. Access on that authorized view could, in turn, be granted to users/groups/service accounts via the `grants` config.
- ~Depending on the outcome of https://github.com/dbt-labs/dbt-bigquery/pull/218, users either will or will not be required to set [`location` in `profiles.yml`](https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile#dataset-locations), if they don't already, in order to use the `grants` feature. I'm hoping the answer is that they don't!~

#### Spark/Databricks

- Not supported on OSS Apache Spark / Delta Lake — at least not yet, and probably not for some time
- In Databricks runtime: Grants are automatically enabled on SQL endpoints. For interactive clusters, admins need to first enable grant functionality via two setup steps: https://docs.databricks.com/administration-guide/access-control/table-acl.html + https://docs.databricks.com/security/access-control/table-acls/table-acl.html

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,67 @@
...
+      grants:
+        select: ['reporter', 'bi']
 ```
 
 </File>
 
-The `meta` config can also be defined under the `seeds` config block in `dbt_project.yml`. See [configs and properties](configs-and-properties) for details.
+The `grants` config can also be defined under the `seeds` config block in `dbt_project.yml`. See [configs and properties](configs-and-properties) for details.
 
 </TabItem>
```
runleonarun, line 73: @jtcohen6  I'm not 100% sure I got this example correct

website/docs/docs/building-a-dbt-project/hooks-operations.md
runleonarun: Addressed in https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/f4c9578e38f22addf04835bcc9e0de536aef8528

website/docs/docs/building-a-dbt-project/hooks-operations.md
runleonarun: Addressed in https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/f4c9578e38f22addf04835bcc9e0de536aef8528

website/docs/docs/building-a-dbt-project/hooks-operations.md
runleonarun: BigQuery distinction addressed in https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/f4c9578e38f22addf04835bcc9e0de536aef8528

Overall examples note addressed here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/8af57ee062cc36b91d3fad081d2969e723dc50eb

website/docs/reference/resource-configs/grants.md
runleonarun: Removed here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/5ffaa49b6ecbc1525e0c3e8cb1c724e049bd5bdd

Is there a follow-up task to add this to the `dbt-docs` site?

website/docs/reference/resource-configs/grants.md
runleonarun: I changed to "admin_user" here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/64fcd6d90666c933e0c25e4324adb0936dc2338b

But I'd like a real-world example here.  What would the people do? @dbeatty10 any ideas for me? :)

website/docs/reference/resource-configs/grants.md
runleonarun: Added here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/64fcd6d90666c933e0c25e4324adb0936dc2338b

website/docs/reference/resource-configs/grants.md
dbeatty10: I've seen the following as a real-worldy example for grants:
```'select': ['bi', 'reporter']```

File: website/docs/docs/building-a-dbt-project/hooks-operations.md (hunk 12 above)
Comment (runleonarun) [reply website/docs/docs/building-a-dbt-project/hooks-operations.md]: Added here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/44e2ef211cb53d609332aedd7eec5e7e2e4af5d5 but need to edit more

File: website/docs/reference/resource-configs/grants.md (hunk 5 above)
Comment (runleonarun) [reply website/docs/reference/resource-configs/grants.md:48]: I still need to add this!

File: website/docs/reference/resource-configs/grants.md [hunk 17]
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
-</TabItem>
+<div warehouse="BigQuery">
 
-<TabItem value="exposures">
+- Be aware that the `grants` config is unrelated to the `grant_access_to` config:
+  - **`grants_access_to`:** Enables you to set up authorized views. When configured, dbt provides authorized view access to other datasets, without leaking additional data from those datasets. Fore more on this, see [BigQuery configurations: Authorized views](/reference/resource-configs/bigquery-configs#authorized-views)
+  - **`grants`:** Provides specific permissions to users, groups, or service accounts for managing access to datasets you're producing with dbt.You could grant a user, group or service account access to an authorized view set up by the `grants_access_to` feature.
+- Use BigQuery-specific grantee and privilege names. 
+  * Use `user:jeremy@dbtlabs.com` (do not use `jerco_user`)
+  * Use  `roles/bigquery.dataViewer` (do not use `select`)
```
jtcohen6: This `<WHCode>` mechanism is really cool!

A good bet, here and for all database-specific idiosyncrasies, is to link out to their docs:
```suggestion
- Use BigQuery-specific grantee and privilege names. 
  * Use `user:jeremy@dbtlabs.com` (do not use `jerco_user`)
  * Use  `roles/bigquery.dataViewer` (do not use `select`)
```
- Understanding GCP "roles" (what we call "privileges"): https://cloud.google.com/iam/docs/understanding-roles
- How to format grantees: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-control-language#user_list

website/docs/docs/building-a-dbt-project/hooks-operations.md
jtcohen6: It's still a bit weird to see the `post-hook` examples below all featuring simple `grant` statements, when we've explicitly said above that users should NOT use `post-hook` to achieve that functionality.

It would make more sense to me if the section below was `<VersionBlock lastVersion="1.1">`, and then for v1.2+, we just had a generic placeholder post-hook (not `WHCode`) like:
```sql
{{ config(
    post_hook=[
      "alter table {{ this }} ..."
    ]
) }}
```

website/docs/reference/resource-configs/grants.md
jtcohen6: This looks right to me!

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
+* Grantees: A way to manage privileges. Recipients of granted privileges, also called "principals." Grantees can be a user, a group of users, a role held by users (Snowflake), a service account (GCP), and more.
 
-</File>
+<WHCode>
 
-</TabItem>
+<div warehouse="BigQuery">
 
-<TabItem value="exposures">
+- Be aware that the `grants` config is unrelated to the `grant_access_to` config:
```
jtcohen6: This is great! Let's add a note to this same effect in the docs for `grant_access_to`. Perhaps a snippet / FAQ in both spots?

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
+  - **`grants_access_to`:** Enables you to set up authorized views. When configured, dbt provides authorized view access to other datasets, without leaking additional data from those datasets. Fore more on this, see [BigQuery configurations: Authorized views](/reference/resource-configs/bigquery-configs#authorized-views)
```
jtcohen6: - Typo
- "Leaking" can be a strong word
```suggestion
- **`grants_access_to`:** Enables you to set up authorized views. When configured, dbt provides an authorized view access to show partial information from other datasets, without providing end users with full access to those underlying datasets. For more on this, see [BigQuery configurations: Authorized views](/reference/resource-configs/bigquery-configs#authorized-views)
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
+  - **`grants`:** Provides specific permissions to users, groups, or service accounts for managing access to datasets you're producing with dbt.You could grant a user, group or service account access to an authorized view set up by the `grants_access_to` feature.
```
jtcohen6:
```suggestion
- **`grants`:** Provides specific permissions to users, groups, or service accounts for managing access to datasets you're producing with dbt.

You can use the two features together: "authorize" a view model with the `grants_access_to` configuration, and then add `grants` to that view model to share its query results (and _only_ its query results) with other users, groups, or service accounts.
```
What this looks like in code, if helpful:
```sql
{{ config(
  materialized = 'view',
  grant_access_to = [{'project': 'sensitive-project', 'dataset': 'sensitive-dataset'}],
  grants = {'roles/bigquery.dataViewer': ['group:analysts@yourcompany.com']}
) }}

select * from {{ ref('model_with_sensitive_data') }} 
where is_sensitive_data = False
```
Big idea: members of the Google group `analysts@yourcompany.com` get to see the data included in this view, but they _don't_ get access to the underlying data in the sensitive model / dataset

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
 </File>
 
+</div>
 
-### Designate a source column as containing PII
+<div warehouse="Databricks">
 
-<File name='models/schema.yml'>
+- OSS Apache Spark / Delta Lake do not support `grants`.
+- Databricks automatically enables `grants` on SQL endpoints. For interactive clusters, admins should enable grant functionality using these two setup steps in teh Dataabricks documentation:
```
jtcohen6: Looks good! Just a typo
```suggestion
- Databricks automatically enables `grants` on SQL endpoints. For interactive clusters, admins should enable grant functionality using these two setup steps in the Databricks documentation:
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -98,152 +72,117 @@ The `meta` config can also be defined under the `seeds` config block in `dbt_pro
...
+* No special requirements at this time.
 
-```
+</div>
 
-</File>
+<div warehouse="Snowflake">
 
-### Configure one meta attribute for all seeds
+* dbt recognizes the `copy_grants` when calculating which grants need to be added or removed.
```
jtcohen6:
```suggestion
* dbt accounts for the [`copy_grants` configuration](/reference/resource-configs/snowflake-configs#copying-grants) when calculating which grants need to be added or removed.
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,69 @@
...
+
+The grant resource configs enable you to apply permissions at build time to a specific set of recipients and model, seed, or snapshot. When your model, seed, or snapshot finishes building, dbt ensures that the grants on its view or table match exactly the grants you have configured.
+
+dbt always finds the most efficient approach when updating grants, which varies based on the adapter you're using, and whether dbt is replacing or updating an object that already exists. You can always check the debug logs for the full set of grant and revoke statements that dbt runs.
+
+When you have a more advanced problem to solve, you might instead use grants with hooks instead of grants as resource configs. For example, if you want to create more granular row- and column-level access, use masking policies, or apply future grants. For more information on hooks, see [Hooks & operations](/building-a-dbt-project/hooks-operations).
```
jtcohen6: > When you have a more advanced problem to solve, you might instead use grants with hooks instead of grants as resource configs.

I think there are two conditions for this, where we'd recommend taking matters into your own hands (= hooks):
- if you need to apply grants in a more complex or custom manner, beyond what the built-in `grants` capability can provide
- you want to take advantage of more-advanced permission capabilities offered by your data platform, for which dbt does not (yet!) offer out-of-the-box support via resource configuration

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,69 @@
...
+id: "grants"
 ---
 
+<Snippet src="available-prerelease-beta-banner" />
+
+You can manage access to the datasets you're producing with dbt by using grants. To implement these permissions, define grants as resource configs on each model, seed, or snapshot. Define the default grants that apply to the entire project in your `dbt_project.yml`, and define model-specific grants within each model's SQL or YAML file.
+
+The grant resource configs enable you to apply permissions at build time to a specific set of recipients and model, seed, or snapshot. When your model, seed, or snapshot finishes building, dbt ensures that the grants on its view or table match exactly the grants you have configured.
+
+dbt always finds the most efficient approach when updating grants, which varies based on the adapter you're using, and whether dbt is replacing or updating an object that already exists. You can always check the debug logs for the full set of grant and revoke statements that dbt runs.
```
jtcohen6: 💯 Thanks for including this! I think the Big Idea expressed here is important for users to grok what's expected of them, and what's going on behind the scenes.

I'd ask that we just soften this promise very slightly:

> dbt always finds the most efficient approach when updating grants

"dbt aims to use the most efficient approach"

Just because we know it won't be 100% perfect, there are always edge cases, and we'll always prefer to be safe rather than sorry — meaning, more queries / less efficient to still guarantee the correct set of `grants` are applied

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -27,14 +27,118 @@ Effective database administration sometimes requires additional SQL statements t
...
+## About hooks
+
 Hooks are snippets of SQL that are executed at different times:
   * `pre-hook`: executed _before_ a model, seed or snapshot is built.
   * `post-hook`: executed _after_ a model, seed or snapshot is built.
   * `on-run-start`: executed at the _start_ of `dbt run`, `dbt seed` or `dbt snapshot`
   * `on-run-end`: executed at the _end_ of `dbt run`, `dbt seed` or `dbt snapshot`
 
-Hooks are defined in your `dbt_project.yml` file. Pre- and post-hooks can also be defined in a `config` block.
+Hooks can help you leverage database-specific actions to expand dbt. You can define hooks in your `dbt_project.yml` file. Pre- and post-hooks can also be defined in a `config` block.
```
jtcohen6: My ongoing agenda to reset user expectations around hooks:
```suggestion
Hooks are a more-advanced capability that enable you to run custom SQL, and leverage database-specific actions, beyond what dbt makes available out-of-the-box with standard materializations and configurations.
```

jtcohen6 (commented):
I think we're getting really close! We may be able to get this over the finish line during / soon after our 1:1 pairing time later today.

### Remaining TODOs

I left a bunch of comments. These are the broader categories of things I think we need to do before merging:

- [ ] Additive vs. clobber behavior when merging `grants` configured in multiple places - https://github.com/dbt-labs/docs.getdbt.com/pull/1642#discussion_r913657517
- [x] Finalize language around `hooks-operations` for pre-v1.2 and v1.2+
- [x] Other pages needing update:
   - [x] Add note + link to "New and changed documentation" section of [v1.2 migration guide](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.2#new-and-changed-documentation)
   - [x] Add `grants` to list of "general configs" for [models](https://docs.getdbt.com/reference/model-configs), [seeds](https://docs.getdbt.com/reference/seed-configs), [snapshots](https://docs.getdbt.com/reference/snapshot-configs)

website/docs/reference/resource-configs/grants.md
runleonarun: Great idea! Addressed here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/d05223dfa7c8c91717bc8d0609b18556a5a91edf

website/docs/docs/building-a-dbt-project/hooks-operations.md
runleonarun: I think I got it right this time. Added here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/668e2472672a32838b5544d29dfff2e28f70bf06

website/docs/reference/resource-configs/grants.md
runleonarun: Love it! Addressed here: https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/b0ca061d74376fab084006a1d8a77d3e1da2d0f7

dataders:
> - [ ] ["Building a new adapter"](https://docs.getdbt.com/docs/contributing/building-a-new-adapter): there are 4 new macros, and adapter maintainers may need to reimplement anywhere between none and all of them to support grants on their database

@runleonarun I just opened #1691, perhaps I can grab some of the work there and bring it into the ["Building a new adapter"](https://docs.getdbt.com/docs/contributing/building-a-new-adapter) page in a separate PR into this branch?

website/docs/reference/resource-configs/grants.md
jtcohen6: That can be food for future thought for the Core team—or maybe for the community member who opens the issue :)

website/docs/reference/resource-configs/grants.md
runleonarun: TY Doug, I changed to those examples.

runleonarun:
@dataders Go for it!

website/docs/reference/resource-configs/grants.md
runleonarun: caught in a previous sweep!

website/docs/reference/resource-configs/grants.md
runleonarun: This was captured!

website/docs/reference/resource-configs/grants.md
runleonarun: Fixed in https://github.com/dbt-labs/docs.getdbt.com/pull/1642/commits/1121db576dbc7c69973edbfd75cdd29abc0c0d65

website/docs/docs/building-a-dbt-project/hooks-operations.md
```diff
@@ -25,18 +25,55 @@ Effective database administration sometimes requires additional SQL statements t
...
 
-### Hooks
 Hooks are snippets of SQL that are executed at different times:
   * `pre-hook`: executed _before_ a model, seed or snapshot is built.
   * `post-hook`: executed _after_ a model, seed or snapshot is built.
   * `on-run-start`: executed at the _start_ of `dbt run`, `dbt seed` or `dbt snapshot`
   * `on-run-end`: executed at the _end_ of `dbt run`, `dbt seed` or `dbt snapshot`
 
-Hooks are defined in your `dbt_project.yml` file. Pre- and post-hooks can also be defined in a `config` block.
+Hooks are a more-advanced capability that enable you to run custom SQL, and leverage database-specific actions, beyond what dbt makes available out-of-the-box with standard materializations and configurations.
```
jtcohen6, line 38: 👍

website/docs/reference/resource-configs/grants.md
```diff
@@ -1,95 +1,74 @@
...
+
+dbt aims to use the most efficient approach when updating grants, which varies based on the adapter you're using, and whether dbt is replacing or updating an object that already exists. You can always check the debug logs for the full set of grant and revoke statements that dbt runs.
+
+We encourage you to use grants as resource configs whenever possible now that they are available. In versions prior to v1.2, you were limited to using hooks for grants. Currently, you still might need to write grants statements manually and run them using hooks. For example, when you want to:
+* Apply grants in a more complex or custom manner, beyond what the built-in grants capability can provide.
+* Apply grants on other database objects besides views and tables.
+* Take advantage of more-advanced permission capabilities offered by your data platform, for which dbt does not (yet!) offer out-of-the-box support via resource configuration
+
+For example, if you want to create more granular row- and column-level access, use masking policies, or apply future grants. For more information on hooks, see [Hooks & operations](/building-a-dbt-project/hooks-operations).
```
jtcohen6, line 21: [Non-blocking] It almost feels like this could go into a nested FAQ element: "Should I use the `grants` config, or should I apply my grants in hooks instead?"

website/docs/reference/seed-configs.md
```diff
@@ -84,6 +84,7 @@ seeds:
     [+](plus-prefix)[persist_docs](persist_docs): <dict>
     [+](plus-prefix)[full_refresh](full_refresh): <boolean>
     [+](plus-prefix)[meta](meta): {<dictionary>}
+    [+](plus-prefix)[grant](grant): {<dictionary>}
```
jtcohen6: [Blocking] Here + elsewhere, this should be plural `grants`:
```suggestion
    [+](plus-prefix)[grants](grants): {<dictionary>}
```

jtcohen6 (approved):
@runleonarun I think the "additive" vs. "clobber" behavior might be the only outstanding TODO. I'd be happy to split out that work into a separate issue, and open a subsequent PR to add it in (adapted from the example code in my comment above).

Pending one quick fix (`grant` &rarr; `grants` in "general" configs for models/snapshots/seeds), I'd be happy to get this merged in so that we can announce the RC!

website/docs/reference/snapshot-configs.md
```diff
@@ -134,6 +134,7 @@ snapshots:
       [pre-hook](pre-hook-post-hook): <sql-statement> | [<sql-statement>]
       [post-hook](pre-hook-post-hook): <sql-statement> | [<sql-statement>]
       [persist_docs](persist_docs): {<dict>}
+      [grant](grant): {<dictionary>}
```
runleonarun:
```suggestion
      [grants](grants): {<dictionary>}
```

website/docs/reference/snapshot-configs.md
```diff
@@ -151,6 +152,7 @@ snapshots:
     [pre_hook](pre-hook-post-hook)="<sql-statement>" | ["<sql-statement>"],
     [post_hook](pre-hook-post-hook)="<sql-statement>" | ["<sql-statement>"]
     [persist_docs](persist_docs)={<dict>}
+    [grant](grant)={<dict>}
```
runleonarun:
```suggestion
    [grants](grants)={<dict>}
```

website/docs/reference/snapshot-configs.md
```diff
@@ -113,7 +113,7 @@ snapshots:
     [+](plus-prefix)[pre-hook](pre-hook-post-hook): <sql-statement> | [<sql-statement>]
     [+](plus-prefix)[post-hook](pre-hook-post-hook): <sql-statement> | [<sql-statement>]
     [+](plus-prefix)[persist_docs](persist_docs): {<dict>}
-
+    [+](plus-prefix)[grant](grant): {<dict>}
```
runleonarun:
```suggestion
    [+](plus-prefix)[grants](grants): {<dict>}
```

website/docs/reference/seed-configs.md
```diff
@@ -112,6 +113,7 @@ seeds:
       [persist_docs](persist_docs): <dict>
       [full_refresh](full_refresh): <boolean>
       [meta](meta): {<dictionary>}
+      [grant](grant): {<dictionary>}
```
runleonarun:
```suggestion
      [grants](grants): {<dictionary>}
```

runleonarun:
@jtcohen6 and @dataders I am OK merging this PR into https://github.com/dbt-labs/docs.getdbt.com/pull/1641 and then both into `current`.

I will capture follow-up work in a new issue. OK to merge?

PR #1420, opened 2022-05-04
Reorg Step 1: Create top level Guides section and tuck FAQs into Docs for phase out

gwenwindflower:
## Description & motivation
This is step 1 of the reorganization process Leona and I are working to implement across the Docs and Guides portion of the Developer Hub. We're collaborating on a more cohesive, modular, DRY, and maintainable Information Architecture (IA) that makes knowledge easier to create, share, and collaborate on for not just us, but the entire dbt Community.

### Step 1 Todos (this PR)
- [x] Create Guides section and add to top nav
- [x] Get rid of 'tutorial' structure and move 'Getting Started' into Guides
- [x] Update 'Learn' tab in top nav to 'Courses'
- [x] Add 'Airflow and dbt' Guide to Guide section
- [x] Finish 'How we structure our dbt projects' revamp and add to Guides section
- [x] Update 'tutorial' language throughout site and use 'guide'/'Getting Started guide', for consistency (Partially complete but needs a full dedicated pass)
- [x] Move FAQs into Docs and out of the top nav.
- [x] Double-check all links are updated and redirects are set (Will do this when we move out of draft)

### Step 2 (What's coming)
- Finalize plan for IA
- Regroup Docs to fit IA
- Move dbt CLI and dbt Cloud sections into Docs and Guides as needed/as fits IA
- Move FAQs into Docs and Guides as needed/as fits IA

PR #1643, opened 2022-06-30
Add redirects for /guides/legacy/

jtcohen6:
I don't understand redirects well enough to know why the changes to `_redirects` in https://github.com/dbt-labs/docs.getdbt.com/pull/1420 aren't working quite right, but it's clear that these aren't being redirected right now:

- https://docs.getdbt.com/docs/guides/getting-help
- https://docs.getdbt.com/docs/guides/best-practices

We link to these in a variety of external places, so it would be best to get the redirects working again. This was a manual attempt based on searching / combing through the file.


jtcohen6:
@runleonarun Looks like there's a merge conflict between the redirects I'm proposing, and the one in the `current` branch:
```
<<<<<<< fix/guides-redirects
/docs/best-practices	/guides/legacy/best-practices	302
=======
/docs/best-practices	/guides/best-practices	302
>>>>>>> current
```
There's a **lot** of content in the ["legacy" best practices page](https://docs.getdbt.com/guides/legacy/best-practices), which is previously where folks would land when they were linked to `/docs/best-practices`. Compared to which, the new ["best practice guides" landing page](https://docs.getdbt.com/guides/best-practices) has only the new guide to project structure. Would it be possible to add a link to the legacy best practices there, too? If so, then I'd feel more comfortable accepting the redirect currently in `current`.

runleonarun (approved):
@jtcohen6 I think I resolved all the conflict and removed the duplicated redirects. Let's merge this and see how it goes!

PR #1676, opened 2022-07-07
Adding Expression Metrics & Metric Types to Metric Page

### LINKED ISSUE #1638: Expression (?) metrics
### Contributions

https://docs.getdbt.com/docs/building-a-dbt-project/metrics

Fantastic expression/derived/composite metrics, and how to define them :)

### Additional information

https://github.com/dbt-labs/dbt-core/pull/5027

callum-mcdata:
## Description & motivation
This PR adds information to the `metrics.md` file about expression metrics. Additionally, it adds a table of supported metric types. NOTE: These metric type definitions actually live within the metrics package but I believe it is important for that information to live in the docs and not just the package README. 

This PR satisfies Issue #1638 

## Prerelease docs
- [X] I've added versioning components, as described in ["Versioning Docs"](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md)


jasnonaz:
This is so awesome thanks for putting together @callum-mcdata!! Looks like the netlify build failed - any thoughts on what we might need to tweak?

callum-mcdata:
Hmm if I had to hazard a guess I would say this is probably because of some `<VersionBlock>` shenanigans. I wonder if `1.2` hadn't been defined as a version block and is failing because of it. Let me see if I can peer behind the netlify curtain

callum-mcdata:
Well when a certain someone hasn't done a pull from `current` since he started, no wonder there are some issues 😅 . Just pulled in the most recent changes and I believe that should fix the issue

jasnonaz (approved):
This is fantastic work! Thanks for putting this together and powering through docusaurus issues. Non blocking question for you: Do you think we should have info about the metrics package in here / link to the metrics package. Feels like this is important information for folks who are getting into metrics.

callum-mcdata:
@jasnonaz that is a really good point - definitely should have some links to the metrics package, just as the metrics package should have some links to this doc so people can cross reference. Let me add an additional section linking to the package

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -29,6 +29,8 @@ A metric is a timeseries aggregation over a <Term id="table" /> that supports ze
 
 In v1.0, dbt supports metric definitions as a new node type. Like [exposures](exposures), metrics participate in the dbt DAG and can be expressed in YAML files. By defining metrics in dbt projects, you encode crucial business logic in tested, version-controlled code. Further, you can expose these metrics definitions to downstream tooling, which drives consistency and precision in metric reporting.
```
jtcohen6, line 30: [Non-blocking] Thoughts on updating this language? It still refers to:
- Metrics being brand-new in v1.0 — less true now than it was last December
- Metrics being **experimental** / liable to change in new minor versions — this is still true, but we anticipate removing this "experimental" designation in a few months!

Happy for those changes to come in a subsequent PR, since it's adjacent scope

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -29,6 +29,8 @@ A metric is a timeseries aggregation over a <Term id="table" /> that supports ze
...
 
+For more information on query the metrics that are defined, please reference the [dbt_metrics package.](https://github.com/dbt-labs/dbt_metrics)
```
jtcohen6: 👍  Definitely think it's more than appropriate to prominently link to the `dbt-metrics` package from this page

[Blocking] I think I know what you're saying here, but the wording is a bit tricky. Do you mean:
- "For more information on querying metrics that are defined in your dbt project"
- "For more information on how metrics queries are defined"
?

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -90,6 +92,85 @@ metrics:
...
 </File>
 
+<VersionBlock firstVersion="1.2">
+
+### Available properties
+
+| Field       | Description                                                 | Example                         | Required? |
+|-------------|-------------------------------------------------------------|---------------------------------|-----------|
+| name        | A unique identifier for the metric                          | new_customers                   | yes       |
+| model       | The dbt model that powers this metric                       | dim_customers                   |     yes (no for `expression` metrics)    |
```
jtcohen6: [Blocking] From a quick visual inspection, I think this is the only change in the table between v1.0/v1.1 &rarr; v1.2:
- `model` + `Required?`: `yes (no for expression metrics)`, instead of `yes` (= always) 

Is that right?

I think our `VersionBlock` logic might be fancy enough to make this work:
```suggestion
| model       | The dbt model that powers this metric                       | dim_customers                   | <VersionBlock firstVersion="1.2">yes (no for `expression` metrics)</VersionBlock><VersionBlock lastVersion="1.1">yes</VersionBlock> |
```
I tried that locally and it seemed to do the trick. Then we don't need to duplicate the full table contents for different docs versions.

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -90,6 +92,85 @@ metrics:
...
+
+<VersionBlock firstVersion="1.2">
+
+### Expression Metrics
+In v1.2, support was added for `expression` metrics, which are defined as non-aggregating calculations of 1 or more metrics. By defining these metrics, you are able to create metrics like:
+- ratios
+- subtractions 
+- any arbitrary calculation
+
+As long as the two+ base metrics (the metrics that comprise the `expression` metric) share the specified `time_grains` and `dimensions`, those attributes can be used in any downstream metrics macro.
```
jtcohen6, line 132: [Non-blocking]

> As long as the two+ base metrics (the metrics that comprise the `expression` metric) share the specified `time_grains` and `dimensions`

Ah - do we have validation for this in `dbt-core`? In the `dbt-metrics` package?

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -90,6 +92,85 @@ metrics:
...
+ - name: dim_customers
+   ...
+
+metrics:
+  - name: average_revenue_per_customer
+    label: Average Revenue Per Customer
+    description: "The average revenue received per customer"
+
+    type: expression
+    sql: "{{metric('total_revenue')}} / {{metric('count_of_customers')}}"
```
jtcohen6, line 153: [Non-blocking] Is this the syntax we'd recommend, over syntax like:
```yaml
    sql: "{{ metric('total_revenue') ~ "/" ~ metric('count_of_customers') }}"
```
(Does that alt syntax even work?)

jtcohen6 (changes requested):
@callum-mcdata Thanks for taking this on! I left a few comments, and called out which ones are [blocking] or [non-blocking] for merging this PR.

Could you also add a note to [the "New and changed documentation" section of the v1.2 migration guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/guides/migration/versions/upgrading-to-v1.2.md), describing the change at a high level (one sentence) and linking to this page?

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: Wow, yea the language here is ... not great 😅 . I will update to:

`For more information on querying the metrics defined in your dbt project`

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: We have "validation" for this in the `dbt_metrics` package. The macro constructs a list of acceptable dimensions from all metrics used and then the provided dimensions are compared against that list. The macro errors out if it is an invalid dimension.

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: I tested the alt syntax and it didn't appear to work. Even testing a few different options around quotation placement returned unicode errors. 

To me that is fine anyway because I wouldn't want users to have to worry about concatenation within the sql field. Self contained makes most sense to me and personally feels the most ergonomic.

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: Resolved!

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: I agree - opened #1690 to remember this suggestion and track it in a future PR!

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -90,6 +92,85 @@ metrics:
...
 </File>
 
+<VersionBlock firstVersion="1.2">
+
+### Available properties
+
+| Field       | Description                                                 | Example                         | Required? |
+|-------------|-------------------------------------------------------------|---------------------------------|-----------|
+| name        | A unique identifier for the metric                          | new_customers                   | yes       |
+| model       | The dbt model that powers this metric                       | dim_customers                   | <VersionBlock firstVersion="1.2">yes (no for `expression` metrics)</VersionBlock><VersionBlock lastVersion="1.1">yes</VersionBlock> |
```
jtcohen6, line 100: Idea in making this change is that we can have just one table of "Available properties," rather than duplicating it here + below (beneath `<VersionBlock lastVersion="1.1">`)

callum-mcdata:
@jtcohen6 thanks for all the helpful comments! I've resolved the blocking issues and addressed the non-blocking issues (opened an issue for one of them).  The upgrading page has also been updated with the following line:

- Updated `metrics.md` to reflect the addition of the `metric` function and support for `expression` metrics.

Let me know if there is anything blocking for this PR!

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: Accepted changes!

website/docs/guides/migration/versions/upgrading-to-v1.2.md
```diff
@@ -22,5 +22,6 @@ We added a collection of ["cross-database macros"](cross-database-macros) to dbt
 ## New and changed documentation
 
 _Under construction_
+- Updated `metrics.md` to reflect the addition of the `metric` function and support for `expression` metrics.
```
jtcohen6: Two things:
- The intent here is not to keep a pure changelog, but rather to give users a "highlight reel" of what's changed, and links to read more
- You'll need to resolve a merge conflict before merging (most PRs for v1.2 target updates to this page)
```suggestion
- [**Metrics**](building-a-dbt-project/metrics) now support an `expression` type (metrics-on-metrics), as well as a `metric()` function to use when referencing metrics from within models or `expression`-type metrics.
```

website/docs/docs/building-a-dbt-project/metrics.md
jtcohen6: Not quite. Here's a commit with what I mean: https://github.com/dbt-labs/docs.getdbt.com/commit/54b9b818db0121d2e8fea92690d169933585cdaf

website/docs/docs/building-a-dbt-project/metrics.md
```diff
@@ -90,6 +92,85 @@ metrics:
...
+| time_grains | One or more "grains" at which the metric can be evaluated   | [day, week, month]              | yes       |
+| dimensions  | A list of dimensions to group or filter the metric by       | [plan, country]                 | no        |
+| filters     | A list of filters to apply before calculating the metric    | See below                       | no        |
+| meta        | Arbitrary key/value store                                   | {team: Finance}                 | no        |
+
+</VersionBlock>
+
+<VersionBlock firstVersion="1.2">
+
+### Available types
```
jtcohen6, line 112: Any reason why we can't show this table (or at least most of it) to users of v1.0 + v1.1 as well? I'd propose removing the `VersionBlock` wrapping around the whole table, and then putting something like this for the last line:
```
| expression | <VersionBlock firstVersion="1.2">This metric type is defined as any **non-aggregating** calculation of 1 or more metrics</VersionBlock><VersionBlock lastVersion="1.1">Not yet available — added in v1.2</VersionBlock> |
```

jtcohen6 (approved):
LGTM pending these two changes!
- [X] Update migration guide note (to resolve merge conflict)
- [X] DRY up "Available properties" table

**CM Edit**: *Tasks accomplished on 7/13/22*

website/docs/docs/building-a-dbt-project/metrics.md
callum-mcdata: I'm on board with this idea - I think it was something missing from the pre-1.2 versions anyway so if I can backfill it that's a bonus in my book.

callum-mcdata:
@jtcohen6 The issues raised above have been resolved. For some reason though, I swear I read that one of the comments was around updating the **discussions** section but I'm unable to find that now. Was I dreaming and imagining that or has it just hidden itself in this comment thread?

website/docs/docs/building-a-dbt-project/metrics.md
jtcohen6: @callum-mcdata Still need to delete the duplicated version of this table below as well:

Screenshot 2022-07-14 at 01 50 40

callum-mcdata: Odd - when I spun it up on my local there was only one table 🤔  I wonder if I messed up on the merge conflict from current. Let me fix now!

callum-mcdata: Yep that was it! Fixed it:

image

jtcohen6 (approved):
Let's ship it!

callum-mcdata:
@annafil @runleonarun I'll let y'all press the big green button as you are the codeowners for this repo. Or if you'd prefer me to press it, let me know!

jtcohen6:
@callum-mcdata I think you're good to merge yourself, since you did the leg work here! But I'll press it if you don't want to ;)

callum-mcdata:
GONNA PRESS THAT BUTTON THEN! 🕺

PR #1688, opened 2022-07-12
Update Oracle profile

aosingh:
- New python driver `python-oracledb` related changes which vastly simplifies dbt-oracle installation
- Oracle wallet configuration for ADB (Autonomous Database in Cloud) is explained in a better manner for 1-way TLS and m-TLS
- Overall content organization

## To-do before merge
- [x] Release `dbt-oracle==1.1.1 `on PyPI which adds support for the Oracle's newly released thin driver

### DIFF
--- a/website/docs/reference/warehouse-profiles/oracle-profile.md
+++ b/website/docs/reference/warehouse-profiles/oracle-profile.md
@@ -6,7 +6,7 @@ title: "Oracle Profile"
 
 **Maintained by:** Oracle    
 **Source:** [Github](https://github.com/oracle/dbt-oracle)    
-**Core version:** v1.0.7     
+**Core version:** v1.1.1     
 **dbt Cloud:** Not Supported    
 **dbt Slack channel** [#db-oracle](https://getdbt.slack.com/archives/C01PWH4TXLY)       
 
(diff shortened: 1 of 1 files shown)


dataders (approved):
>  - [ ] Release dbt-oracle==1.1.1 on PyPI which adds support for the Oracle's newly released thin driver
tag me here in a comment when 1.1.1 has been released!

one very extra thing you could do (if you think it's worthwhile) is version this document such that if folks are looking at this page for dbt versions prior to 1.1, they'll see the old information about the JDBC drive. but for versions 1.1 and onwards, they'll see these new options. Not a requirement, but something to think about. I think I'd decide on this depending on how many folks are using `dbt-oracle<=1.1.0`... Check out our [Docs Versioning Guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md) if you really want to get your hands dirty

aosingh:
Thanks @dataders. I can see why docs versioning is useful. 

Although, in this PR, could we skip versioning due to the following reasons ?

- The overall content organization and wallet related docs changes will benefit both users ( using `dbt-oracle >=1.1` and `dbt-oracle<1.1` )

- I thought about using version blocks but `Oracle Client Libraries` installation is required for the new driver's `thick` mode and the old `cx-oracle` driver.  Thus it may happen that both old and new users might install Oracle Client libraries. Hence I have made the distinction in between `thin` and `thick` modes using tabs. 

- For users upgrading to dbt-oracle==1.1.1 without reading the docs, dbt flows will not break. By default dbt-oracle uses the old driver `cx_oracle`. They can continue using dbt as-is. Although, this documentation will help them migrate to the new thin driver. 

Let me know if you have any feedback or questions. I will keep versioning in mind for any future documentation updates. 

Also, we have almost finalized dbt-oracle==1.1.1 and will release it soon on PyPI. I will tag you here.

dataders:
> Although, in this PR, could we skip versioning

yep zero problem!

aosingh:
Hi @dataders 

We have released `dbt-oracle==1.1.1`. This can be merged now. 

Thanks!

PR #1692, opened 2022-07-12
Document TestDocsGenerate for 1.2 upgrade guide

dataders:
## Description & motivation

Document this new basic test and how it works so that adapter maintainers can use it

## Prerelease docs

## Checklist


website/docs/guides/migration/versions/upgrading-to-v1.2.md
```diff
@@ -101,9 +101,18 @@ Via a community contribution from the folks at Layer.ai, [dbt-core#5348](https:/
 
 this is a big deal for folks who are inheriting adapters, e.g. as dbt-synapse does with dbt-sqlserver, and for the family of adapters inherit from dbt-spark today.
 
+### New basic test: `TestDocsGenerate`
+
+[dbt-core#5058](https://github.com/dbt-labs/dbt-core/pull/5058) is another step along [the path of converting all our functional tests](https://github.com/dbt-labs/dbt-core/issues/4788) to the new framework in order to empower  adapter maintainers and other contributors to make use of the same tests that the core team uses for their own adapters. Effectively, this test is validates an adapter's ability to correctly generate the catalog that serves as the static backend of a project docs site.
+PLACEHOLDER ABOUT `adapter_stats()` and `base_generate_catalog()`
```
dataders: @gshank and/or @jtcohen6 I'd love your input here. I spent 10 minutes and couldn't wrap my head around everything. It seems in order for this test to be userful there has to be interesting adapter-specific information that's added to the catalog?

jtcohen6: @dataders When generating the `catalog.json` artifact (during `docs generate`), dbt expects each catalog entry (model, seed, snapshot, or source) to include three sets of information:
1. Common [table metadata](https://github.com/dbt-labs/dbt-core/blob/f988f76fccc1878aaf8d8631c05be3e9104b3b9a/core/dbt/contracts/results.py#L420-L427): relation type (view or table), schema, database, name/identifier, relation-level comment (description), owner
2. A set of `columns`, where each entry also features [common metadata](https://github.com/dbt-labs/dbt-core/blob/f988f76fccc1878aaf8d8631c05be3e9104b3b9a/core/dbt/contracts/results.py#L409-L414): name, data type, index (ordinal position), comment (description)
3. An _optional_ set of relation-level statistics, which completely varies by adapter/database. This could include table size (rows + bytes), last modified timestamp, ...

My understanding is that the tests built into the default tests, `BaseDocsGenerate` and `BaseDocsGenReferences`, test for 1 + 2 and [skip over 3](https://github.com/dbt-labs/dbt-core/blob/f988f76fccc1878aaf8d8631c05be3e9104b3b9a/tests/adapter/dbt/tests/adapter/basic/test_docs_generate.py#L435), because Postgres does not provide any optional statistics.

If an adapter plugin provides no optional supplementary statistics, the out-of-the-box test is expected to work. If it does provide an optional set of statistics, the right move is to:
- Add a method that defines stats for this adapter [e.g. dbt-bigquerys](https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/expected_stats.py)
- Reimplement the `expected_catalog` fixture, [passing the above into `model_stats` and `seed_stats`](https://github.com/dbt-labs/dbt-bigquery/blob/0212fd621ede4c24929a008de718a7e45bc32cec/tests/functional/adapter/test_basic.py#L68-L81)

@gshank Questions for you when you're back:
- It looks like we got this test running for Redshift, Postgres, BigQuery. What's the status of the PRs for Snowflake + Spark?
- Is there a reason why the test checks only for model + seed stats, and not for sources + snapshots as well? I'd also expect stats to vary based on whether the model is materialized as a view or a table. I'm guessing that the answer to both is, because the new test is based on the contents of the old test

dataders:
```suggestion
If your adapter does not add extra relation-level metadata (e.g. table size (rows + bytes), last modified timestamp) which is the case by default, then you can follow the same inherit and `pass` pattern to enable your version of `BaseDocsGenerate` and `BaseDocsGenReferences`. However, if you are supplementing the catalog with more metadata, you'll have to:
- Add a method that defines stats for this adapter [e.g. dbt-bigquerys](https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/expected_stats.py)
- Reimplement the `expected_catalog` fixture, [passing the above into `model_stats` and `seed_stats`](https://github.com/dbt-labs/dbt-bigquery/blob/0212fd621ede4c24929a008de718a7e45bc32cec/tests/functional/adapter/test_basic.py#L68-L81)

```

website/docs/guides/migration/versions/upgrading-to-v1.2.md
```diff
@@ -101,9 +101,18 @@ Via a community contribution from the folks at Layer.ai, [dbt-core#5348](https:/
 
 this is a big deal for folks who are inheriting adapters, e.g. as dbt-synapse does with dbt-sqlserver, and for the family of adapters inherit from dbt-spark today.
 
+### New basic test: `TestDocsGenerate`
```
dataders: @jtcohen6 are there technically two new tests they should implement? 
```suggestion
### New basic tests to implement in adapterland: `BaseDocsGenerate` and `BaseDocsGenReferences`
```

nathaniel-may (approved):
for persnicketiness.

PR #1691, opened 2022-07-12
1.2.0 upgrade guide for adapter maintainers

dataders:
## Description & motivation

Continues on #1634. 

## Prerelease docs
If this change is related to functionality in a prerelease version of dbt (delete if not applicable):
- [x] I've added versioning components, as described in ["Versioning Docs"](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md)
- [ ] I've added a note to the prerelease version's [Migration Guide](https://github.com/dbt-labs/docs.getdbt.com/tree/current/website/docs/guides/migration/versions)

PR #1696, opened 2022-07-13
Move v1.2 adapter upgrade guide to separate page

jtcohen6:
## Description & motivation

@dataders added a ton of great content to the migration guide relevant for adapter maintainers in #1691. However, this content is ONLY relevant to maintainers of adapter plugins (a few dozen people), and the primary audience for the upgrade/migration guide is all dbt users (thousands).

This PR moves this content into its own dedicated page, so that it takes up less real estate in the general-purpose migration guide. **Question:** Is a docs.getdbt.com page the right place for this content? Would it be better as a GitHub discussion (easier to edit/update, see questions from other maintainers)?

In a separate commit, I also removed the section about `get_csv_sql()`. While there's nothing inherently wrong with the addition of this macro in dbt-labs/dbt-core#5207, this did NOT accomplish the thing we hoped it would in `dbt-snowflake`, for some very tricky reasons: https://github.com/dbt-labs/dbt-adapters/issues/740


dataders (changes requested):
@jtcohen6 my idea here is to move the adapter migration guide to a GitHub discussion rather than a separate section of migration guides... perhaps we close this instead?

jtcohen6:
closing!

dataders:
> @jtcohen6 my idea here is to move the adapter migration guide to a GitHub discussion rather than a separate section of migration guides... perhaps we close this instead?

here is where the adapter guide will live. note to self to link out to this in the end-user upgrade guide, perhaps as well in the doc's contributing section?
https://github.com/dbt-labs/dbt-core/discussions/5468

jtcohen6:
> note to self to link out to this in the end-user upgrade guide

✅

PR #1703, opened 2022-07-13
Update next branch to current and remove redirect

JKarlavige:
This merges `current` into `next`, with the one addition of removing the `next.docs` redirect from the `_redirects` file. 

To verify this has the latest updates that are live in `current`, we can review this page: https://deploy-preview-1703--docs-getdbt-com.netlify.app/faqs/Docs/sharing-documentation, which has the correct statement: ` you can have up to 5 read-only users` which was added into `current` today


runleonarun (approved):
LGTM!

PR #1705, opened 2022-07-14
correct and simplify info for Snowflake's `client_session_keep_alive` param

dataders:
## Description & motivation

Someone from Snowflake told us that this was incorrect, AFAIK, they are right. Regardless, end users needn't know how the driver keeps connections alive for longer than four hours, only that it does if you set `client_session_keep_alive` to be `True`.


amychen1776 (approved):
:shipit:

PR #1710, opened 2022-07-14
verify updates to next are pulling

JKarlavige:
test - this is to verify next.docs.getdbt.com is pulling from the next branch, and not current.


lfracalossiFA (approved):

JKarlavige:
Verified this updated the site at `next.docs.getdbt.com`, all looks good here. Reverting
image

PR #1711, opened 2022-07-14
Update and Re-indent Building Adapter page for 1.2

dataders:
## Description & motivation

- added an overview section of the different areas that need work
- change indentation level
- updated the "(source)" URLs of all the macros to point to their new locations
- added an explanation of standardize_grants_dict() method
- instead of adding the GRANTS macros to the list of "required macros" I just made a new section and linked to them

Questions for @jtcohen6 (the first is the only required question for this PR, the other two are what I've been thinking
1. Should the `grant` macros be in "Required Macros" list? 
2. What does it mean if a macro is in the "Required Macros" list but doesn't have `(required)` after it? AFAIK, it means they have `default__` implementations but perhaps we're strongly suggest that they make their own version? I ask because isn't so exhausted
image
3. True or False: This Building an Adapters page should have any core-versioning dynamics because we'd recommend to anyone who wants to make an adapter that they should start with the latest version of dbt-core


dataders:
closing in favor of #1727

PR #1734, opened 2022-07-18
Fix broken URL in upgrade guide

joellabes:


joellabes:
@runleonarun I'm trying to fix this broken link 
image
which is trying to point to https://docs.getdbt.com/guides/migration/versions/reference/dbt-jinja-functions instead of https://docs.getdbt.com/reference/dbt-jinja-functions. 

It's weird though because the [preview](https://deploy-preview-1734--docs-getdbt-com.netlify.app/guides/migration/versions/upgrading-to-v1.2#new-and-changed-documentation) is still pointing to the wrong place, even though it's formatted the same as the `/modules#itertools` link in the same line 😵  What am i doing wrong?

runleonarun:
Hey @joellabes I think I fixed this in another PR. I believe it has to do with the page being auto generated and so you can no longer leverage the `id` when it's not unique and shares a name with another file. So we'd need to disambiguate with adding the entire path `/reference/dbt-jinja-functions`. I'm not sure why the other links, for example, `/dbt-jinja-functions/those_other-pages` work. Maybe because it IS unique?  

We might be able to create unique ids for generated pages but I haven't looked into that yet.

PR #1753, opened 2022-07-20
Create modify-owner-column.md

mirnawong1:
added faq for docs per zendesk confusion here: https://dbtcloud.zendesk.com/agent/tickets/31945

## Prerelease docs

## Checklist


website/docs/faqs/Docs/modify-owner-column.md
```diff
@@ -0,0 +1,15 @@
+---
+title: How do I populate the owner column in the generated docs?
+description: "Modify owner column"
+sidebar_label: 'Can I populate owner column in docs?'
+id: modify-owner-column
+---
+
+
+Unfortunately, you're unable to modify the owner column your generated documentation.
```
matthewshaver: I would avoid negative words like "unfortunately" when telling people we can't do something because of the tone it sets.

Maybe something like:
```suggestion
Due to the nature of the field, you won't be able to change the owner column in your generated documentation. 
```

runleonarun (approved):
Approving so we can merge!

PR #1754, opened 2022-07-20
[Beta] Python models

### LINKED ISSUE #1664: Beta docs: Python models
### Contributions

There are only a bajillion places throughout our docs site where we refer to "model SQL." The work of updating all of those is immense, and out of scope for this issue.

The simplest way to go, for starters, is by creating a new page. That page should note things like:
- Current intentions and limitations
- Example of a working Python model
- Recommended syntax for data frame transformations — we're still developing these opinions!
- Support for common packages (including Pandas)

Once written, we will want to link to that page from _somewhere_. I think the right places will be:
- https://docs.getdbt.com/docs/building-a-dbt-project/building-models
- v1.3 migration guide

I'm honestly not sure!
- A new section in ["Building Models"](https://docs.getdbt.com/docs/building-a-dbt-project/building-models)? A new FAQ?
- Is there a better page?

Between now and October, the section will only be visible to users who select v1.3 from the version dropdown, which will also display a big prerelease banner at the top of page. So, we have until October to figure out how we want to integrate this functionality into the docs site at large :)

### Additional information

Discussion: https://github.com/dbt-labs/dbt-core/discussions/5261

Initial draft PRs:
- https://github.com/dbt-labs/dbt-core/pull/5421
- https://github.com/dbt-labs/dbt-snowflake/pull/182
- https://github.com/dbt-labs/dbt-spark/pull/377
- https://github.com/dbt-labs/dbt-bigquery/pull/209

jtcohen6:
resolves https://github.com/dbt-labs/docs.getdbt.com/issues/1664

## Description & motivation

The first set of "dbt Python model" functionality will be included in v1.3.0-b1 (planned release: next week or the week after). We'll be asking folks to beta-test, and we need to give them docs so that they can.

- Add initial docs for Python models
- Add v1.3 to the version picker (although it doesn't seem to be showing up?)
- Initialize v1.3 migration guide

### TODO before merge
- [x] Prose / code snippets I called out as `TODO`
- [x] Review based on code implementation that actually ends up in v1.3.0-b1
- [x] In places where the syntax differs, should we include warehouse/engine-specific code options for Snowpark + PySpark (+ Pandas)? — I limited the differences as much as possible

### Future — shouldn't block merge for beta
- Demo / walk-through that @lostmygithubaccount is putting together. Should this go in the docs, or a DevHub post + GitHub repo?
- Updates to many, many, _many_ more pages throughout the docs that make reference to dbt models being SQL only

## Prerelease docs
- [x] I've added versioning components, as described in ["Versioning Docs"](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/versioningdocs.md)
- [x] I've added a note to the prerelease version's [Migration Guide](https://github.com/dbt-labs/docs.getdbt.com/tree/current/website/docs/guides/migration/versions)

## Checklist


jtcohen6:
This needs another pass-through, but I'd like to open it up for feedback now. Aiming to get these docs live & ready for beta testers, as soon as there's a beta prerelease to start using.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+    # limit data in dev
+    if target_name == "dev":
+        orders_df = orders_df.limit(500)
+```
+
+</File>
+
+### Materializations
+
+Python models support two materializations:
```
jeremyholtzman, line 213: @jtcohen6 is it worth adding a call out as to why it can't materialize as a view and ephemeral?

On a similar train of thought, can you configure these as snapshots? I assume not?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+## What is a Python model in dbt?
+
+A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
+
+Unlike SQL models, which return a `select` statement, each Python model returns a single <Term id="dataframe">data frame</Term>. When you run a Python model, the result of that data frame will be saved as a table in your data warehouse.
+
+Each Python model lives in a `.py` file in your `models/` folder. It defines a function named **`model()`**, which has two parameters:
+- **`dbt`**: A class compiled by dbt Core, unique to each model, that enables you to run your Python code in the context of your dbt project and DAG.
+- **`session`**: A class representing the connection to the backing engine, which allows you to interact with the data platform. The session is needed to read in tables as data frames, and to writ data frames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
```
jeremyholtzman: typo: `writ` to `write`

jeremyholtzman: As someone who hasn't tried this out yet, when I run a specific python model with `dbt run --select my_python_model` - do I need to do anything to pass in the parameters for `dbt` and `session`? Or is that done automatically?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
+---
+title: "Python models"
+---
+
+:::info Beta
+
+This feature is arriving in dbt Core v1.3, which is currently in beta. We encourage you to:
```
jeremyholtzman: is it worth highlighting that it's not available in cloud yet (unless I'm missing something)?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+    
+    # setting configuration
+    dbt.config(materialized="table")
+```
+
+</File>
+
+The `config()` function accepts _only_ literal values (strings, booleans, and numeric types). It is not possible to pass another function or more complex data structure. The reason: dbt statically analyzes the arguments to `config()` while parsing your model, without actually executing any of your Python code.
+
+Python models have limited access to project context, including inputs such as `var`, `env_var`, and `target`. If you want to use those values to power conditional logic in your model, we recommend passing them through a dedicated `.yml` file config instead:
```
joellabes: Does this mean "there's not much access to project context, you only have var, env_var and target"? That's how I first read it, but then reading the code sample I suspect it's the opposite: "there isn’t much access to project context, you don't even get var, env_var or target"

In addition to clarifying this sentence, it might be useful to have a list of what things are/aren't available to Python (if that's a dozenish items - less so if it's hundreds)

website/docs/docs/building-a-dbt-project/building-models/python-models.md
KiraFuruichi: Is it also worth making a callout that they're only available in BQ, Snowflake, and Databricks earlier here?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
+---
+title: "Python models"
+---
+
+:::info Beta
+
+This feature is arriving in dbt Core v1.3, which is currently in beta. We encourage you to:
+- Read [the discussion](https://github.com/dbt-labs/dbt-core/discussions/5261)
+- Join the **#beta-feedback-python-models** channel in the dbt Community Slack
```
KiraFuruichi:
```suggestion
- Join the **#beta-feedback-python-models** channel in the [dbt Community Slack](https://www.getdbt.com/community/join-the-community/)
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+- returns: a data frame (Snowpark, PySpark, or Pandas)
+
+dbt Python models have access to almost all of the same configuration options as SQL models. You can test them, document them, add `tags` and `meta` properties to them, persist their descriptions as database comments, grant access on their results to other users, and so on. You can select them by their name, their file path, their configurations, whether they are upstream or downstream of another model, or whether they have been modified compared to a previous project state.
+
+### Referencing other models
+
+Your Python model will want to read data from other models (SQL or Python). Do this using the `dbt.ref()` function. The same idea applies for raw source tables, via `dbt.source()`. Those functions return data frames, pointing to the upstream source, model, seed, or snapshot.
```
KiraFuruichi:
```suggestion
Your Python model will want to read data from other models (SQL or Python) or sources. Do this using the `dbt.ref()` function. The same idea applies for raw source tables, via `dbt.source()`. Those functions return data frames, pointing to the upstream source, model, seed, or snapshot. 
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Defining functions
+
+In addition to defining a `model` function, the Python model can import other functions, or define its own functions. Here's an example, on Snowpark, defining a custom `add_one` function:
+
+<File name='models/my_python_model.py'>
+
+```python
+def add_one(x):
+    return x + 1
```
joellabes, line 297: Are functions only:
- user-definable inside of the same file
- importable from a public PyPI package
? 

Or can you do something chaotic like `import add_one from my_python_model` in `my_second_python_model.py`?

Related but broader question: I assume there are no Python macros? Either Jinja to template python or a global space for `add_one` to be defined once

website/docs/terms/dataframe.md
```diff
@@ -0,0 +1,24 @@
+---
+id: dataframe
+title: Data Frame
```
KiraFuruichi:
```suggestion
title: Data frame
```
Adjusting capitalization to be consistent with other glossary pages :)

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+</File>
+
+#### UDFs
+
+You can use the `@udf` decorator or `udf` function to define an "anonymous" function, and call it within your `model` function's dataframe transformation. This is a common pattern for applying more complex functions as data frame operations, especially if those functions require inputs from third-party packages.
+- [Snowpark Python: Creating UDFs](https://docs.snowflake.com/en/developer-guide/snowpark/python/creating-udfs.html)
+- ["PySpark functions: udf"](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.udf.html)
+
+:::info ❔ Our questions
```
KiraFuruichi: Should we make the question mark a red one? I couldn't see the white/gray one at first and a red one will probably catch the eye more.

KiraFuruichi:
This is so extremely exciting :) Jeremy H beat me to some of the things I initially noticed (explaining why view + ephemeral materializations are not supported, spelling typo), but I left some minor aesthetic comments as well.

I think bigger picture (which I don't think should be a blocker **at all** to releasing this tomorrow) that I was thinking about as I was reading this was:
- General use cases for Python dbt models: I think this doc covers really well what dbt Python models can do in comparison to what current dbt models do. Are there certain things we don't want people trying out yet? Are there certain scenarios we can provide (or link to) for folks/beta testers to find use cases for Python models? I'm sure if they're trying them out in the beta, they already have some ideas of their own, which is why I don't think this a blocker or anything.
- Python model location: I know it's mentioned here that Python models should live in `/models` directory, which makes total sense to me, but was also left with a sense of, "But where in the `/models` directory?" I wonder down the line how we could help folks figure out where Python models should live, especially as they start being used for DS/ML. Right now, it makes a ton of senses for Python models to fall under the normal dimensional modeling/modular data modeling structure teams have in-place (since [assumption] they are likely to be purely transformative models like current dbt models). This is also somewhat related to the comment in the UDF section around a `function` endpoint as you consider Python functions to potentially be reusable across multiple dbt models.
- So freaking excited to see a `data frame` glossary page! In the next few weeks I'm going to (actually do my job and) expand on that glossary page to (selfishly) build in some SEO and follow the format we have going for other glossary pages (if this is cool with you).

website/docs/docs/building-a-dbt-project/building-models/python-models.md
joellabes: Probably overkill for this single page, but we can control how admonitions are rendered: https://docusaurus.io/docs/next/markdown-features/admonitions#customizing-rendering-behavior

website/docs/docs/building-a-dbt-project/building-models/python-models.md
ChenyuLInx: Call out of convention is great!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
+
+Unlike SQL models, which return a `select` statement, each Python model returns a single <Term id="dataframe">data frame</Term>. When you run a Python model, the result of that data frame will be saved as a table in your data warehouse.
+
+Each Python model lives in a `.py` file in your `models/` folder. It defines a function named **`model()`**, which has two parameters:
+- **`dbt`**: A class compiled by dbt Core, unique to each model, that enables you to run your Python code in the context of your dbt project and DAG.
+- **`session`**: A class representing the connection to the backing engine, which allows you to interact with the data platform. The session is needed to read in tables as data frames, and to writ data frames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
+- returns: a data frame (Snowpark, PySpark, or Pandas)
```
ChenyuLInx: We only support materialize pandas df in snowflake right now

website/docs/docs/building-a-dbt-project/building-models/python-models.md
ChenyuLInx: Currently you cannot do that. It is possible to support(not very pretty way). @jtcohen6 maybe that's something to call out that we don't support it.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+:::info ❔ Our questions
+When developing a new dbt Python model, should we recommend Pandas-style syntax for rapid iteration, and then refactor? Which open source libraries provide compelling abstractions across different data engines and vendor-specific APIs? Should dbt attempt to play a longer-term role in standardizing across them?
+:::
+
+### Limitations
+
+Python models have capabilities that SQL models do not. They also have some drawbacks compared to SQL models:
+- **Time and cost.** Python models are slower to run than SQL models, and the cloud resources that run them can be more expensive. Running Python requires more general-purpose compute, and in some cases that compute may live on a separate service or architecture from your SQL models. However, we believe that deploying Python models via dbt is **dramatically** faster and cheaper than spinning up separate tooling and infrastructure to orchestrate Python transformations in production.
```
ChenyuLInx: In general I think the cost for the actual execution of the model is going to be similar, but we will save the cost to have separate orchestration and have better integration with the rest of your data that created by dbt project

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+**Docs:** ["Developer Guide: Snowpark Python"](https://docs.snowflake.com/en/developer-guide/snowpark/python/index.html)
+
+</div>
+
+<div warehouse="Databricks">
+
+**Additional setup:** The `user` field in the `dbt-spark` profile, usually optional, is required for Python modeling.
+
+**Note:** Python models will be created and run as notebooks in your Databricks workspace. The notebooks will be created within the personal workspace of the `user` running dbt.
```
ChenyuLInx: Should we add that we don't delete the notebook so you can use it for interactive development of model and bring the code back?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,378 @@
...
+
+</div>
+
+<div warehouse="BigQuery">
+
+The `dbt-bigquery` uses a service called Dataproc to submit your Python models as PySpark jobs. That Python/PySpark code will read from your tables and views in BigQuery, and saves its final result back to BigQuery.
+
+**Additional setup:**
+- Create a dedicated Dataproc cluster: https://cloud.google.com/dataproc/docs/guides/create-cluster
+- Create a dedicated Cloud Storage bucket: https://cloud.google.com/storage/docs/creating-buckets
```
ChenyuLInx: Maybe dedicated is a bit dramatic if user already have those setup they can just use it

website/docs/terms/dataframe.md
b-per: The consensus online seems actually to be with `DataFrame`. I checked Pandas, Snowpark and Spark and they all use it in one word, with the `F` uppercase

website/docs/docs/building-a-dbt-project/building-models.md
```diff
@@ -17,6 +17,14 @@ If you're new to dbt, we recommend that you check out our [Getting Started guide
 
 :::
 
+<VersionBlock firstVersion="1.3">
+
+In v1.3, dbt Core is adding support for **Python models**.
+
+dbt's Python capabilities are an extension of its capabilities with SQL models. We recommend that you read this page first, and then read: ["dbt Models (Python)"](python-models)
```
b-per: Isn't the page called "Python Models" instead of "dbt Models (Python)"?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Good point re: ephemeral, and clarifying no support for non-model resource types (snapshots, tests).

Views are tricky. I don't want to get _too_ into the weeds here. Views actually are supported by [Snowpark](https://docs.snowflake.com/en/developer-guide/snowpark/python/working-with-dataframes.html#creating-a-view-from-a-dataframe), but there isn't an analogue in PySpark. Up to us if we want to add it in a platform-specific way.

website/docs/docs/building-a-dbt-project/building-models.md
b-per: Might be worth mentioning also that what says `.sql` below would need to be replaced by `.py` for Python. No need to go into too much detail in that page but just calling it out.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Yup, good point of clarification — these arguments are populated by dbt when you run the model. Nothing you need to pass in, you just need to define the function that accepts them.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: > is it worth highlighting that it's not available in cloud yet (unless I'm missing something)?

It will be available in dbt Cloud, but with some functional gaps that we're working to close over the next ~month

> Is it also worth making a callout that they're only available in BQ, Snowflake, and Databricks earlier here?

Good call! Cody had the same thought and added a link

website/docs/terms/dataframe.md
```diff
@@ -0,0 +1,24 @@
...
+displayText: dataframe  
+hoverSnippet: A data frame is a two-dimensional data structure (rows and columns). It's the most common way of representing and interacting with large datasets in Python.
+---
+:::important This page could use some love
+This term would benefit from additional depth and examples. Have knowledge to contribute? [Create a discussion in the docs.getdbt.com GitHub repository](https://github.com/dbt-labs/docs.getdbt.com/discussions) to begin the process of becoming a glossary contributor!
+:::
+
+<!--- Useful reference: https://databricks.com/glossary/what-are-dataframes --->
+
+A data frame is a way of storing and manipulating data in Python. (It's also used in other languages popular for data processing, such as R and Scala.)
```
b-per: Would it make sense to do "manipulating data in Python" => "manipulating **tabular** data in Python"

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,379 @@
...
+
+A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
+
+Unlike SQL models, which return a `select` statement, each Python model returns a single <Term id="dataframe">data frame</Term>. When you run a Python model, the result of that data frame will be saved as a table in your data warehouse.
+
+Each Python model lives in a `.py` file in your `models/` folder. It defines a function named **`model()`**, which has two parameters:
+- **`dbt`**: A class compiled by dbt Core, unique to each model, that enables you to run your Python code in the context of your dbt project and DAG.
+- **`session`**: A class representing the connection to the backing engine, which allows you to interact with the data platform. The session is needed to read in tables as data frames, and to write data frames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
+- returns: a data frame (Snowpark, PySpark, or Pandas)
```
b-per: instead of "- returns" that makesit part of the parameters list I'd write "<newline> and returns ..."

website/docs/docs/building-a-dbt-project/building-models/python-models.md
b-per: In addition to Chenyu's comment, shouldn't we list which df style is supported on which platform? e.g. what dfs work with BQ?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: I played around with a few options, and I think I'm going to go with `:::note` + red question mark for now

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,379 @@
...
+        orders_df = orders_df.limit(500)
+```
+
+</File>
+
+### Materializations
+
+Python models support two materializations:
+- `table`
+- `incremental`
```
b-per, line 215: Are Python models supporting the same incremental strategies as their SQL counterparts?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,379 @@
...
+  - name: my_python_model
+    config:
+      packages:
+        - "numpy==1.23.1"
+        - scikit-learn
+```
+
+</File>
+
+#### UDFs
```
b-per, line 426: Do we have an example? I personally am not sure about:
- how to create it
- how to use it

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,379 @@
...
+
+When developing a Python model, you will find yourself asking these questions:
+
+**Why Pandas?** The most common API for data frames. Makes it easy to explore sampled data and develop transformations locally. For small datasets, you can "promote" your code as-is into dbt models, and run in production.
+
+**Why _not_ Pandas?** Performance. Pandas runs "single-node" transformations, which cannot benefit from the parallelism and distributed compute offered by modern data warehouses. This quickly becomes a problem as you operate on larger datasets. Some data platforms do support optimizations for code written using Pandas' data frame API, preventing the need for major refactors. For example, ["Pandas on PySpark"](https://spark.apache.org/docs/latest/api/python/getting_started/quickstart_ps.html) offers support for 95% of Pandas functionality using the same API.
```
b-per: "offers support for 95% of Pandas functionality using the same API" => "offers support for 95% of Pandas functionality using the same API while still leveraging parallel processing"

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,379 @@
...
+
+**Installing packages:** Snowpark supports a number of popular packages via Anaconda. The full list is at https://repo.anaconda.com/pkgs/snowflake/. Packages are installed at the time your model is being run. Different models can have different package dependencies. If you are using third-party packages, for better performance, Snowflake recommends using a dedicated virtual warehouse, rather than one that's likely to have many concurrent users.
+
+**Docs:** ["Developer Guide: Snowpark Python"](https://docs.snowflake.com/en/developer-guide/snowpark/python/index.html)
+
+</div>
+
+<div warehouse="Databricks">
+
+**Additional setup:** The `user` field in the `dbt-spark` profile, usually optional, is required for Python modeling.
```
b-per: Could we link the profile docs page here and for BQ?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
b-per: I would also put a link to the bottom of the page where we explain the tradeoffs and differences between those

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: > there isn’t much access to project context, you don't even get var, env_var or target

Correct. In the current implementation, you get almost nothing, _unless_ you pass it in explicitly as `config`. That does get you access to `var`, `env_var`, `target` if you'd like. But no config-setting via macros, no `graph` introspection.

website/docs/docs/building-a-dbt-project/building-models.md
```diff
@@ -17,6 +17,14 @@ If you're new to dbt, we recommend that you check out our [Getting Started guide
 
 :::
 
+<VersionBlock firstVersion="1.3">
+
+In v1.3, dbt Core is adding support for **Python models**.
```
b-per, line 22: There is no indication in the current docs whether it is possible to use Python models only in Core or in Cloud as well. Is it intentional? If not we could mention to people what version of dbt they need to pick in dbt Cloud.

b-per (changes requested):
Well done! I think it is already giving a lot of info (enough for people to start). 

I just suggested a few changes.
The most controversial being that I'd be in  favor of using `DataFrame` everywhere instead of `data frame`

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: I really consider this an open question for us!

I'm disinclined to do this one, given the chaos it could lead to, and over-reliance on the local file system:
> Or can you do something chaotic like `import add_one from my_python_model` in `my_second_python_model.py`?

I think we have two good options to pursue:
- Some platforms support registering these functions as "named" / persistent UDFs. This would be the case for "dbt should know about `function`s." These UDFs have some drawbacks; inconsistent support; and the main use case seems to be defining a Python UDF that then gets called from within a SQL model. That seems valuable for exposing a prediction function output of a trained ML model, less so for a utility like `add_one`.
- Potentially more promising: most of these platforms allow you to upload your _own_ home-spun "Python package," as a `.zip`, ==. You could then `import` it and use its functions, as if it were a public PyPI package.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Yup - going to qualify that I mean cost from the perspective of the human maintaining this stuff

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Good call!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: looks like Cody had the same thought - let's tone down the drama

website/docs/docs/building-a-dbt-project/building-models.md
jtcohen6: I tried a few different permutations of this, but I wasn't happy with any. I'm going to stick with minimal note for now. We have a lot of future work to do, to figure out the right ways to integrate Python capabilities / references into the existing docs that say SQL, SQL, `.sql`

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: yessir!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: That makes two of us :) I can try to get a simple working example of this

website/docs/docs/building-a-dbt-project/building-models.md
jtcohen6: I'll add a note about this to the top of the "Python models" page

jtcohen6:
Thank you for the quick reviews and excellent feedback @lostmygithubaccount @ChenyuLInx @KiraFuruichi @joellabes @jeremyholtzman @b-per!! Thrilled that this turned into a real team effort.

> So freaking excited to see a data frame glossary page! In the next few weeks I'm going to (actually do my job and) expand on that glossary page to (selfishly) build in some SEO and follow the format we have going for other glossary pages (if this is cool with you).

You have the conn, Furuichi

> The most controversial being that I'd be in favor of using `DataFrame` everywhere instead of `data frame`

Good by me! Made this change.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+Currently, Python models are supported for users of `dbt-snowflake`, `dbt-spark` (Databricks), and `dbt-bigquery`. Both Databricks and BigQuery use PySpark as the processing framework. Snowflake uses its own framework, Snowpark, which has many similarities to PySpark.
+
+:::caution
+
+The implementation for Python models on GCP (BigQuery + Dataproc) is the roughest of the three. Running PySpark on Dataproc requires more manual setup and configuration. Clusters require sufficient resources or auto-scaling policies to handle concurrent Python model runs.
+
+We have made the code available for the beta, but we are reserving the right to leave it out of the final v1.3 release if the experience is too unfriendly to end users. If you're a GCP expert, we're very open to hearing your thoughts in Slack or the GitHub discussions!
+
+:::
```
jtcohen6: @dataders This reads a bit too harsh right now. My goal here is to set clear expectations for any BQ users around both (a) the setup experience, (b) the possibility that this functionality is cut from v1.3 (final). Definitely open to your wordsmithing

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
+---
+title: "Python models"
+---
+
+:::info Beta
+
+This feature is arriving in dbt Core v1.3, which is currently in beta. Note that only [specific data warehouses](#specific-data-warehouses) support dbt Python models, and that support for some dbt Cloud capabilities will be in progress during the beta period.
```
dataders: not sure  what "some dbt Cloud capabilities" means here. do you mean to say
> these beta capabilities in Python will be available in a very limited capacity, but improved over the course of the beta

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- Join the **#beta-feedback-python-models** channel in the [dbt Community Slack](https://www.getdbt.com/community/join-the-community/)
+
+Below, you'll see sections entitled "❓ **Our questions**." We're working to develop our opinionated recommendations ahead of the final release this October—and you can help! Comment in the GitHub discussions, leave thoughts in Slack, talk about it with colleagues and friends.
+:::
+
+## What is a Python model in dbt?
+
+A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
+
+Unlike SQL models, which return a `select` statement, each Python model returns a single <Term id="dataframe">DataFrame</Term>. When you run a Python model, the result of that DataFrame will be saved as a table in your data warehouse.
```
dataders: Been chewing on this line for a while now
```
A SQL model's format has always been something that can compile to jinja and be executed against the data platform and materialized as specified. dbt is used to declare the end state of the model, and the data platform's query engine does the work. What's different about Python models is that the `.py` file serves to define a Python function that outputs a `dataframe` object. When running a Python model, this dataframe is created, then materialized as specified. There are some parallels to the declarative nature of SQL vs. the imperative nature of Python

The "output" of a  model file is a `select` statement that is bundled into a corresponding `create table` or `create view` statement and sent to the datawarehouse. However, the mechanics 
Unlike SQL models, which return a `select` statement, each Python model returns a single <Term id="dataframe">DataFrame</Term>. When you run a Python model, the result of that DataFrame will be saved as a table in your data warehouse.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+dbt Python models have access to almost all of the same configuration options as SQL models. You can test them, document them, add `tags` and `meta` properties to them, grant access on their results to other users, and so on. You can select them by their name, their file path, their configurations, whether they are upstream or downstream of another model, or whether they have been modified compared to a previous project state.
+
+### Defining a Python model
+
+Each Python model lives in a `.py` file in your `models/` folder. It defines a function named **`model()`**, which has two parameters:
+- **`dbt`**: A class compiled by dbt Core, unique to each model, that enables you to run your Python code in the context of your dbt project and DAG.
+- **`session`**: A class representing the connection to the backing engine, which allows you to interact with the data platform. The session is needed to read in tables as DataFrames, and to write DataFrames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
```
dataders: does the end user ever have to modify these parameters to be different than `dbt` and `session`? If not, why expose them to users at all?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+<File name='models/my_python_model.py'>
+
+```python
+def model(dbt, session):
+  
+    # DataFrame representing an upstream model
+    upstream_model = dbt.ref("upstream_model_name")
+    
+    # DataFrame representing an upstream source
+    upstream_source = dbt.source("upstream_source_name", "table_name")
```
dataders, line 120: are these lazily-evaluated references? or does `dbt.source()` read the reference into memory at time of definition?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
nghi-ly:
```suggestion
- **`session`**: A class representing the connection to the backend engine, which allows you to interact with the data platform. The session is needed to read in tables as DataFrames, and to write DataFrames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
```

nghi-ly: not sure if this was a typo so just flagging

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+### Defining a Python model
+
+Each Python model lives in a `.py` file in your `models/` folder. It defines a function named **`model()`**, which has two parameters:
+- **`dbt`**: A class compiled by dbt Core, unique to each model, that enables you to run your Python code in the context of your dbt project and DAG.
+- **`session`**: A class representing the connection to the backing engine, which allows you to interact with the data platform. The session is needed to read in tables as DataFrames, and to write DataFrames back to tables. In PySpark, by convention, the `SparkSession` is named `spark`, and available globally. For consistency across platforms, we always pass it into the `model` function as an explicit argument named `session`.
+
+The `model()` function must return a single DataFrame. On Snowpark (Snowflake), this can be a Snowpark or Pandas DataFrame. On PySpark (Databricks + BigQuery), this should be a PySpark DataFrame (converted back from Pandas if needed). For more about choosing between Pandas and native DataFrames, see ["DataFrame API + syntax"](#dataframe-api--syntax)
```
nghi-ly:
```suggestion
The `model()` function must return a single DataFrame. On Snowpark (Snowflake), this can be a Snowpark or Pandas DataFrame. On PySpark (Databricks + BigQuery), this should be a PySpark DataFrame (converted back from Pandas if needed). For more about choosing between Pandas and native DataFrames, see ["DataFrame API + syntax"](#dataframe-api--syntax).
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    
+    return final_df
+```
+
+</File>
+
+When you `dbt run --select python_model`, dbt will prepare and pass in both arguments (`dbt` and `session`). All you have to do is define the `model()` function that accepts them.
+
+### Referencing other models
+Your Python model will want to read data from other models (SQL or Python) or sources. Do this using the `dbt.ref()` function. The same idea applies for raw source tables, via `dbt.source()`. Those functions return DataFrames, pointing to the upstream source, model, seed, or snapshot.
```
nghi-ly:
```suggestion
Use the `dbt.ref()` function to let your Python model read data from other models (SQL or Python) or sources. The same idea applies for raw source tables from `dbt.source()`. Those functions return DataFrames, pointing to the upstream source, model, seed, or snapshot. 
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+```python
+def model(dbt, session):
+    
+    # setting configuration
+    dbt.config(materialized="table")
+```
+
+</File>
+
+The `config()` function accepts _only_ literal values (strings, booleans, and numeric types). It is not possible to pass another function or more complex data structure. The reason: dbt statically analyzes the arguments to `config()` while parsing your model, without actually executing any of your Python code.
```
nghi-ly:
```suggestion
The `config()` function accepts _only_ literal values (strings, booleans, and numeric types). It's not possible to pass another function or a more complex data structure. The reason is because dbt statically analyzes the arguments to `config()` while parsing your model, without actually executing any of your Python code.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+</File>
+
+### Configuring Python models
+
+Just like SQL models, there are three ways to configure Python models:
+1. In `dbt_project.yml`, where you can configure many models at once
+2. In a dedicated `.yml` file, within the `models/` directory
+3. Within the model's `.py` file, using the `dbt.config()` function
+
+The `dbt.config()` function allows you to set configurations for your model within your Python file:
```
dataders:
```suggestion
Using the `dbt.config()` method within your `model()` definition allows you to set configurations within your Python file, similar to how you would for a `.sql` model:
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    dbt.config(materialized="table")
+```
+
+</File>
+
+The `config()` function accepts _only_ literal values (strings, booleans, and numeric types). It is not possible to pass another function or more complex data structure. The reason: dbt statically analyzes the arguments to `config()` while parsing your model, without actually executing any of your Python code.
+
+#### Accessing project context
+
+dbt Python models do not use Jinja to render compiled code. Compared to SQL models, Python models have very limited access to global project context. That context is made available via the `dbt` class, passed in as an argument to the `model()` function.
```
nghi-ly:
```suggestion
dbt Python models don't use Jinja to render compiled code. Compared to SQL models, Python models have very limited access to global project context. That context is made available from the `dbt` class, passed in as an argument to the `model()` function.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
dataders:
```suggestion
There's a hard limit as to how fancy you can get with the  `.config()` method because it accepts _only_ literal values (strings, booleans, and numeric types). It is not possible to pass another function or more complex data structure. The reason: dbt statically analyzes the arguments to `config()` while parsing your model, without actually executing any of your Python code.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+</File>
+
+### Materializations
+
+Python models support two materializations:
+- `table`
+- `incremental`
+
+Incremental Python models support all the same [incremental strategies](configuring-incremental-models#about-incremental_strategy) as their SQL counterparts, for the adapter on which they're running.
```
nghi-ly:
```suggestion
Incremental Python models support all the same [incremental strategies](configuring-incremental-models#about-incremental_strategy) as their SQL counterparts, for the adapter they're running on.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+At present, Python models cannot be materialized as `view` or `ephemeral`. Python is not supported for non-model resource types (e.g. tests, snapshots).
```
nghi-ly:
```suggestion
At present, Python models can't be materialized as `view` or `ephemeral`. Python isn't supported for non-model resource types (e.g. tests, snapshots).
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    return df
+```
+
+</File>
+
+</div>
+
+</WHCode>
+
+**Note:** Incremental models are not yet supported on BigQuery.
```
nghi-ly:
```suggestion
**Note:** Incremental models aren't currently supported on BigQuery.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
dataders:
```suggestion
Your Python model will want to read data from other models (SQL or Python) or sources. Do this using `dbt`'s `ref()` method. The same idea applies for raw source tables, via `dbt.source()`. Both methods return DataFrames, pointing to the upstream source, model, seed, or snapshot. 

```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    temps_df = dbt.ref("temperatures")
+    
+    # warm things up just a little
+    df = temps_df.withColumn("degree_plus_one", add_one(temps_df["degree"]))
+    return df
+```
+
+</File>
+
+At present, Python functions defined in one dbt model cannot be imported and reused in other models. See the section below, ["Code reuse"](#code-reuse), for the patterns we're considering.
```
nghi-ly:
```suggestion
At present, Python functions defined in one dbt model can't be imported and reused in other models. See the ["Code reuse"](#code-reuse) section for the patterns we're considering.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+```
+
+</File>
+
+### Configuring Python models
+
+Just like SQL models, there are three ways to configure Python models:
+1. In `dbt_project.yml`, where you can configure many models at once
+2. In a dedicated `.yml` file, within the `models/` directory
+3. Within the model's `.py` file, using the `dbt.config()` function
```
dataders:
```suggestion
3. Within the model's `.py` file, using the `dbt.config()` method
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    config:
+      materialized: table
+      target_name: "{{ target.name }}"\
+      specific_var: "{{ var('SPECIFIC_VAR') }}"
+      specific_env_var: "{{ env_var('SPECIFIC_ENV_VAR') }}"
+```
+
+</File>
+
+Then, within the model's Python code, use the `dbt.config.get()` function to _access_ values of configurations that have been set:
```
dataders, line 192:
```
Then, within the model's Python code, use the `dbt.config.get()` method to _access_ values of configurations that have been set:
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+    return df
+```
+
+</File>
+
+At present, Python functions defined in one dbt model cannot be imported and reused in other models. See the section below, ["Code reuse"](#code-reuse), for the patterns we're considering.
+
+### Using PyPI packages
+
+You can also define functions that depend on installed packages. In this example, we use the `holidays` package to determine if a given date fell on a holiday in France. For simplicity and consistency across platforms, the code below uses the Pandas API. The exact syntax, and the need to refactor for multi-node processing, still varies.
```
nghi-ly:
```suggestion
You can also define functions that depend on installed packages. In this example, we use the `holidays` package to determine if a given date is a holiday in France. For simplicity and consistency across platforms, the code below uses the `pandas` API. The exact syntax and the need to refactor for multi-node processing still varies.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+Then, within the model's Python code, use the `dbt.config.get()` function to _access_ values of configurations that have been set:
+
+<File name='models/my_python_model.py'>
+
+```python
+def model(dbt, session):
+    target_name = dbt.config.get("target_name")
+    specific_var = dbt.config.get("specific_var")
+    specific_env_var = dbt.config.get("specific_env_var")
```
dataders, line 200: @lostmygithubaccount this brings back many azureml sdk argparse memories. I'm trying to remember the hack I had to just unpack all the config into a dictionary... at the risk of saying something crazy, perhaps arg parse could help us make things more simple? idk
https://stackoverflow.com/a/22570801/3842610

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+</File>
+
+</div>
+
+</WHCode>
+
+#### Configuring packages
+
+We encourage you to explicitly configure required packages and versions, so that dbt can track them in project metadata. This configuration is required for the implementation on some platforms. If you need specific versions of packages, specify them.
```
nghi-ly:
```suggestion
We encourage you to explicitly configure required packages and versions so dbt can track them in project metadata. This configuration is required for the implementation on some platforms. If you need specific versions of packages, specify them.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- Should dbt have a role in abstracting over UDFs? Should dbt support a new type of DAG node, `function`? Would the primary use case be code reuse across Python models, or defining Python-language functions that can be called from SQL models?
+- How can dbt help users when uploading or initializing private Python assets? Is this a new form of `dbt deps`?
+- How can dbt support users who want to test custom functions? If defined as UDFs: "unit testing" in the database? If "pure" functions in packages: encourage adoption of `pytest`?
+:::
+
+### DataFrame API + syntax
+
+Over the past decade, most people writing data transformations in Python have adopted <Term id="dataframe">DataFrame</Term> as their common abstraction. dbt follows this convention by returning `ref()` and `source()` as DataFrames, and it expects all Python models to return a DataFrame.
+
+A DataFrame is a two dimensional data structure (rows and columns). It supports convenient methods for transforming that data, creating new columns from calculations performed on existing columns. It also offers convenient methods for previewing data while developing locally or in a notebook.
```
nghi-ly:
```suggestion
A DataFrame is a two-dimensional data structure (rows and columns). It supports convenient methods for transforming that data, creating new columns from calculations performed on existing columns. It also offers convenient methods for previewing data while developing locally or in a notebook.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
dataders:
```suggestion
:::caution

Here be dragons; this implementation for Python models on GCP (BigQuery + Dataproc) needs the most love of the three before the official release.

For example, running PySpark on Dataproc requires more manual setup and configuration. Clusters require sufficient resources or auto-scaling policies to handle concurrent Python model runs.

We have made the code available for the beta, but we are reserving the right to leave it out of the final v1.3 release if the experience is too unfriendly to end users. If you're a GCP expert, we'd love your help; please share your thoughts in Slack or the GitHub discussions!

:::
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+:::note ❓ Our questions
+- When developing a new dbt Python model, should we recommend Pandas-style syntax for rapid iteration, and then refactor?
+- Which open source libraries provide compelling abstractions across different data engines and vendor-specific APIs?
+- Should dbt attempt to play a longer-term role in standardizing across them?
+:::
+
+### Limitations
+
+Python models have capabilities that SQL models do not. They also have some drawbacks compared to SQL models:
+- **Time and cost.** Python models are slower to run than SQL models, and the cloud resources that run them can be more expensive. Running Python requires more general-purpose compute, and in some cases that compute may live on a separate service or architecture from your SQL models. **However:** We believe that deploying Python models via dbt—with unified lineage, testing, and documentation—is, from a human standpoint, **dramatically** faster and cheaper. By comparison, spinning up separate infrastructure to orchestrate Python transformations in production, and separate tooling to integrate with dbt, is much more time-consuming and expensive.
```
nghi-ly:
```suggestion
- **Time and cost.** Python models are slower to run than SQL models, and the cloud resources that run them can be more expensive. Running Python requires more general-purpose compute and, in some cases, that compute might live on a separate service or architecture from your SQL models. **However:** We believe that deploying Python models via dbt—with unified lineage, testing, and documentation—is, from a human standpoint, **dramatically** faster and cheaper. By comparison, spinning up separate infrastructure to orchestrate Python transformations in production, and separate tooling to integrate with dbt, is much more time-consuming and expensive.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- **These capabilities are very new.** As data warehouses develop new features, we expect them to offer cheaper, faster, and more intuitive mechanisms for deploying Python transformations. We reserve the right to change the underlying implementation for executing Python models. Our commitment to you is around the code in your model `.py` files, following the guidance above.
+
+As a general rule, if there's a transformation you could write equally well in SQL or in Python, we believe that well-written SQL is preferable: it's more accessible to a greater number of colleagues, and it's easier to write code that's performant at scale. If there's a transformation you _cannot_ write in SQL, or where 10 lines of elegant and well-annotated Python could save you 1000 lines of hard-to-read Jinja-SQL, Python is the way to go.
```
nghi-ly:
```suggestion
As a general rule, if there's a transformation you could write equally well in SQL or in Python, we believe that well-written SQL is preferable: it's more accessible to a greater number of colleagues, and it's easier to write code that's performant at scale. If there's a transformation you _can't_ write in SQL, or where 10 lines of elegant and well-annotated Python could save you 1000 lines of hard-to-read Jinja-SQL, Python is the way to go.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+:::
+
+<WHCode>
+
+<div warehouse="Snowflake">
+
+**Additional setup:** None needed. Snowpark Python is in Public Preview - Open, and enabled by default for all accounts.
+
+**Installing packages:** Snowpark supports a number of popular packages via Anaconda. The full list is at https://repo.anaconda.com/pkgs/snowflake/. Packages are installed at the time your model is being run. Different models can have different package dependencies. If you are using third-party packages, for better performance, Snowflake recommends using a dedicated virtual warehouse, rather than one that's likely to have many concurrent users.
```
nghi-ly:
```suggestion
**Installing packages:** Snowpark supports a number of popular packages via Anaconda. The full list is at https://repo.anaconda.com/pkgs/snowflake/. Packages are installed at the time your model is being run. Different models can have different package dependencies. If you are using third-party packages, Snowflake recommends using a dedicated virtual warehouse for better performance, rather than one that's likely to have many concurrent users.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+**Docs:** ["Developer Guide: Snowpark Python"](https://docs.snowflake.com/en/developer-guide/snowpark/python/index.html)
+
+</div>
+
+<div warehouse="Databricks">
+
+**Additional setup:** The `user` field in your [Spark connection profile](spark-profile), usually optional, is required for running Python models.
```
nghi-ly:
```suggestion
**Additional setup:** The `user` field in your [Spark connection profile](spark-profile) (which is usually optional) is required for running Python models.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+**Docs:**
+- [PySpark dataframe syntax](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.html)
+- [Databricks: Introduction to DataFrames - Python](https://docs.databricks.com/spark/latest/dataframes-datasets/introduction-to-dataframes-python.html)
+
+</div>
+
+<div warehouse="BigQuery">
+
+The `dbt-bigquery` adapter uses a service called Dataproc to submit your Python models as PySpark jobs. That Python/PySpark code will read from your tables and views in BigQuery, and saves its final result back to BigQuery.
```
nghi-ly:
```suggestion
The `dbt-bigquery` adapter uses a service called Dataproc to submit your Python models as PySpark jobs. That Python/PySpark code will read from your tables and views in BigQuery, and save its final result back to BigQuery.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+**Docs:** ["Developer Guide: Snowpark Python"](https://docs.snowflake.com/en/developer-guide/snowpark/python/index.html)
+
+</div>
+
+<div warehouse="Databricks">
+
+**Additional setup:** The `user` field in your [Spark connection profile](spark-profile), usually optional, is required for running Python models.
+
+**Note:** Python models will be created and run as notebooks in your Databricks workspace. The notebook will be created within the personal workspace of the `user` running dbt, and named after the model it is executing. dbt will update the notebooks on subsequent runs of the same model, but it will not delete it—so you can use the notebook for quicker interactive development. Just remember to update the code in your dbt model, based on your in-notebook iteration, before the next `dbt run`.
```
dataders: I feel there's an opportunity to be more explicit about the proposed workflow.
```suggestion
**Note:** Python models will be created and run as notebooks in your Databricks workspace. The notebook will be created within the personal workspace of the `user` running dbt, and named after the model it is executing. dbt will update the notebooks on subsequent runs of the same model, but it will not delete it. We're thinking of this as a feature rather than a bug. The idea being that you can

  1. scaffold a simple version of the model and `run` it,
  2. go to Databricks to interactively develop it to the desired state, then
  3. copy your changes back into your dbt `.py` model and call `dbt run` again
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+</div>
+
+<div warehouse="BigQuery">
+
+The `dbt-bigquery` adapter uses a service called Dataproc to submit your Python models as PySpark jobs. That Python/PySpark code will read from your tables and views in BigQuery, and saves its final result back to BigQuery.
+
+**Additional setup:**
+- Create or use an existing [Cloud Storage bucket](https://cloud.google.com/storage/docs/creating-buckets)
+- Create or use an existing [Dataproc cluster](https://cloud.google.com/dataproc/docs/guides/create-cluster), with the [Spark BigQuery connector initialization action](https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/connectors#bigquery-connectors). (Google recommends copying the action into your own Cloud Storage bucket, rather than using the example version shown in the screenshot below.)
```
nghi-ly:
```suggestion
- Create or use an existing [Dataproc cluster](https://cloud.google.com/dataproc/docs/guides/create-cluster) with the [Spark BigQuery connector initialization action](https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/connectors#bigquery-connectors). (Google recommends copying the action into your own Cloud Storage bucket, rather than using the example version shown in the screenshot below.)
```

website/docs/guides/migration/versions/05-upgrading-to-v1.3.md
```diff
@@ -0,0 +1,31 @@
...
+
+For users of [state-based selection](understanding-state): This release includes logic providing backwards and forwards compatibility for older manifest versions. While running dbt Core v1.3, it should be possible to use `state:modified --state ...` selection against a manifest produced by dbt Core v1.0+.
+
+## For maintainers of adapter plugins
+
+_GitHub discussion forthcoming_
+
+## New and changed documentation
+
+- **[Python models](building-models-with-python)** are natively supported in `dbt-core` for the first time, on data warehouses that support Python runtimes.
```
nghi-ly, line 32:
```suggestion
- **[Python models](building-models-with-python)** are natively supported in `dbt-core` for the first time on data warehouses that support Python runtimes.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+</div>
+
+<div warehouse="Databricks">
+
+**Additional setup:** The `user` field in your [Spark connection profile](spark-profile), usually optional, is required for running Python models.
+
+**Note:** Python models will be created and run as notebooks in your Databricks workspace. The notebook will be created within the personal workspace of the `user` running dbt, and named after the model it is executing. dbt will update the notebooks on subsequent runs of the same model, but it will not delete it—so you can use the notebook for quicker interactive development. Just remember to update the code in your dbt model, based on your in-notebook iteration, before the next `dbt run`.
+
+**Installing packages:** We recommend configuring packages on the interactive cluster which you will be using to run your Python models.
```
dataders, line 573: are we recommending to initially install all the packages you might use before creating python models in lieu of specifying requirements as part of a Python model's specifification? As in, it's not  easy/possible/practical to configure model-specific python environments, so just do it once manually outside of dbt?

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- Create or use an existing [Dataproc cluster](https://cloud.google.com/dataproc/docs/guides/create-cluster), with the [Spark BigQuery connector initialization action](https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/connectors#bigquery-connectors). (Google recommends copying the action into your own Cloud Storage bucket, rather than using the example version shown in the screenshot below.)
+
+<Lightbox src="/img/docs/building-a-dbt-project/building-models/python-models/dataproc-connector-initialization.png" title="Add the Spark BigQuery connector as an initialization action"/>
+
+Add these attributes to your [BigQuery profile](bigquery-profile):
+- `gcs_bucket`: Storage bucket to which dbt will submit PySpark code, and the Dataproc cluster will stream back execution logs.
+- `dataproc_cluster_name`: Name of Dataproc cluster to use for running Python model (executing PySpark job)
+- `dataproc_region`: GCP region of that Dataproc cluster (e.g. `us-central1`)
+
+The user or service account running dbt needs the following permissions, in addition to those needed for BigQuery ([docs](https://cloud.google.com/dataproc/docs/concepts/iam/iam)):
```
dataders: saying that all users need to permisssions makes the work clearer than if it's just for a single user...
```suggestion
Any user or service account that will run dbt projects that include `.py` files will need the following permissions in addition to those needed for BigQuery
```

website/docs/terms/dataframe.md
```diff
@@ -0,0 +1,24 @@
...
+---
+:::important This page could use some love
+This term would benefit from additional depth and examples. Have knowledge to contribute? [Create a discussion in the docs.getdbt.com GitHub repository](https://github.com/dbt-labs/docs.getdbt.com/discussions) to begin the process of becoming a glossary contributor!
+:::
+
+<!--- Useful reference: https://databricks.com/glossary/what-are-dataframes --->
+
+A DataFrame is a way of storing and manipulating tabular data in Python. (It's also used in other languages popular for data processing, such as R and Scala.)
+
+It's possible to string together a number of DataFrame transformations. For example, if `df` represents a DataFrame containing one row per person living in the Eastern United States over the last decade, we can calculate the number of people living in Philly each year:
```
nghi-ly: consider using `Philadelphia` unstead of `Philly` to be less US centric, less colloquial

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- dataproc.clusters.use
+- dataproc.jobs.create
+- dataproc.jobs.get
+- dataproc.operations.get
+- storage.buckets.get
+- storage.objects.create
+- storage.objects.delete
```
dataders: @dataders are you trying to enfore monospace API permissions?!? do you think this is a game for children!?!
```suggestion
- `dataproc.clusters.use`
- `dataproc.jobs.create`
- `dataproc.jobs.get`
- `dataproc.operations.get`
- `storage.buckets.get`
- `storage.objects.create`
- `storage.objects.delete`
```

nghi-ly (approved):
beta docs look great, @jtcohen6 ! small nits. non-blocking.

dataders (commented):
reading this felt like visiting my childhood home as an adult after many years and a new family lives there. Just when I thought I got out of the python data transformation world -- they pull me back in!

some considerations:
1. "DataFrame" or "dataframe"? relatedly, do all references of the word need to link out to the glossary?
2. I'm not sure what the WH-specific "Installing Packages" advice is regarding. As an antecedent step to defining model level requirements?
3. i'm 92% that `dbt.ref()` and the like are better categorized as methods not functions... but don't care that much..

website/docs/terms/dataframe.md
dataders:
```suggestion
It's possible to string together a number of DataFrame transformations. For example, if `df` represents a DataFrame containing one row per person living in the Eastern United States over the last decade, we can calculate the number of people living in Philadelphia each year:
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
runleonarun:
```suggestion
dbt Core v1.3, which is currently in beta, includes Python models. Note that only [specific data warehouses](#specific-data-warehouses) support dbt Python models, and that support for some dbt Cloud capabilities will be in progress during the beta period.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
+---
+title: "Python models"
+---
+
+:::info Beta
+
+This feature is arriving in dbt Core v1.3, which is currently in beta. Note that only [specific data warehouses](#specific-data-warehouses) support dbt Python models, and that support for some dbt Cloud capabilities will be in progress during the beta period.
+
+We encourage you to:
+- [Read the original discussion](https://github.com/dbt-labs/dbt-core/discussions/5261)
```
runleonarun:
```suggestion
- [Read the original discussion](https://github.com/dbt-labs/dbt-core/discussions/5261) that shaped this feature.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- [Weigh in on our developing best practices](https://github.com/dbt-labs/docs.getdbt.com/discussions/1811)
```
runleonarun:
```suggestion
- Contribute to [best practices for developing Python models in dbt ](https://github.com/dbt-labs/docs.getdbt.com/discussions/1811).
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+- Join the **#beta-feedback-python-models** channel in the [dbt Community Slack](https://www.getdbt.com/community/join-the-community/)
```
runleonarun:
```suggestion
- Join the **#beta-feedback-python-models** channel in the [dbt Community Slack](https://www.getdbt.com/community/join-the-community/).
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+Below, you'll see sections entitled "❓ **Our questions**." We're working to develop our opinionated recommendations ahead of the final release this October—and you can help! Comment in the GitHub discussions, leave thoughts in Slack, talk about it with colleagues and friends.
```
runleonarun:
```suggestion
Below, you'll see sections entitled "❓ **Our questions**." We're working to develop our opinionated recommendations ahead of the final release this October—and you can help! Comment in the GitHub discussions, leave thoughts in Slack, discuss dbt Python models with colleagues and friends.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+:::
+
+## What is a Python model in dbt?
```
runleonarun: I think we can broaden this a bit to excite people about Python models with the "why" it's so dang cool!
```suggestion
## About Python models in dbt
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
```diff
@@ -0,0 +1,560 @@
...
+
+A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
```
runleonarun: It would be nice to provide a tl;dr for the linked discussion/issue so we can spark interest for people who don't like to go off and read more. Can you add just another sentence or two? Possibly an example? I can help if you need it, @jtcohen6, but I wasn't sure what exactly to say. 
```suggestion
dbt Python models will help you solve use cases that you can't solve with SQL, and bring you closer to dbt when the data warehouse is your principal tool. 

A dbt Python model is a function that reads in dbt sources or models, applies a series of transformations, and returns a transformed dataset. Just like SQL models!
```

runleonarun (commented):
Hey @jtcohen6 Just some suggestions. Mostly would like to see a but of context and examples for the beta so people who see this first, want to try it out!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Going with this for now:

> Support in dbt Cloud will be limited to start, and improve over the course of the beta.

I'm really talking about:
- Syntax highlighting in the IDE
- Support for profile configurations on Databricks + BQ
- Some breaking changes for the Metadata API

We're still thinking hard about the right behavior for "Preview Data" in the IDE, given how slow Python models are to run, compared to SQL models.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: Lazy evaluation if you're using a DataFrame API worth its salt — so yes! As soon as you convert to Pandas, though, that will read everything into memory.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: The end user needs to access classmethods of `dbt`, and some properties of `session` as well. We think, better to document those and make it clearer where they're coming from, rather than pulling them out of the aether.

We could have gone for a "script"-style approach, where `dbt` and `session` appear to be available "globally":
```python
def some_fn:
    ...
    
some_df = dbt.ref("some_model")
final_df = some_df.apply(some_fn)

final_df
```
Where really we just take users' code, and turn it into:
```python
def model(dbt, session):
    def some_fn:
        ...
    
    some_df = dbt.ref("some_model")
    final_df = some_df.apply(some_fn)

    return final_df  # we add this return?
```
It's still an option we could consider, if people like that a lot better! (@ChenyuLInx)

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: The way Chenyu implemented this, dbt will statically analyze the literals being passed into `config.get()`, and then supply _just_ those config values in the compiled code. Is that more magic or less magic?

jtcohen6:
Thanks for the thorough reviews @nghi-ly @dataders @runleonarun!!

I'm working through your feedback, haven't gotten all the way through, but pushing up some intermediate progress.

---

@dataders 

> "DataFrame" or "dataframe"?  Benoit says “DataFrame,” and I think that’s the most technically correct. It looks a bit stuffy to me, but I’m okay with the formality for now — it’s a new concept for many.

> relatedly, do all references of the word need to link out to the glossary?

Good question! How does the docs team feel about this? Every instance, just the first one on a given page, something in between?

> I’m 92% that dbt.ref() and the like are better categorized as methods not functions... but don't care that much..


You’re 100% right. Deliciously, in our implementation, dbt.ref() is a classmethod… pointing to a function named ref()! At the risk of some more inconsistency, I’ve done my best to replace instances of “function” with “method,” when that’s the one I mean.

---

@runleonarun 

> Mostly would like to see a but of context and examples for the beta so people who see this first, want to try it out!

At the risk of too much text—this page will definitely want to be split up before final release!—I've added a more exciting intro, as well as a placeholder example, before we get into the nitty-gritty implementation details.

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: just going to put the whole darn thing in backticks!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: This totally varies based on the backend / implementation. On Databricks, it _is_ possible to tell a cluster to run `pip install <some_package>` in a first notebook cell. That can take quite a bit of time, though. Their recommendation (and it makes sense to me) is to configure the needed packages on the cluster ahead of time.

By configuring packages in that way, it does mean that two models cannot have conflicting package version requirements. I think that's a recipe for disaster, anyway...

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: I was trying to be a little cute here. Not sure if I'm achieving it, might be better to cut altogether:
```suggestion
Below, you'll see sections entitled "❓ **Our questions**." We're working to develop our opinionated recommendations ahead of the final release this October—and you can help! Comment in the GitHub discussions; leave thoughts in Slack; bring up dbt + Python in casual conversation with colleagues and friends.
```

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: yes, good framing!

website/docs/docs/building-a-dbt-project/building-models/python-models.md
jtcohen6: I ended up taking a different tack here, incorporating Leona's feedback asking for a hook + higher-level intro

website/docs/docs/building-a-dbt-project/building-models/python-models.md
joellabes: I agree that you shouldn't be able to do arbitrary imports from functions defined in other models in the `models/` directory.

> most of these platforms allow you to upload your own home-spun "Python package," as a .zip, ==. You could then import it and use its functions, as if it were a public PyPI package

Where would the code that gets zipped and uploaded be source controlled? This feels very analogous to the before-state of https://discourse.getdbt.com/t/using-dbt-to-manage-user-defined-functions/18 - code is magically available but no one really knows where it came from. 

If we can resolve that, I do like the idea of an internal functions package. 

Would finding `.py` functions defined in the `macros` directory (or add `functions/` to your `macro-paths`) and embedding them all into an internal package be good enough at breaking reliance on the filesystem? This assumes that you're disinclined to enable imports from inside a model because it would be bad practice, not that it would be hard to implement technically.

jtcohen6: > Where would the code that gets zipped and uploaded be source controlled?

Yup — 100% needs to live within the dbt project, and be version-controlled along with it.

I think that "upload" step will require different behind-the-scenes implementations across different data warehouses. We could even opt for UDFs as the backing implementation, if a given warehouse doesn't support arbitrary package upload. There's risk of a leaky abstraction here, but it feels like a place where dbt can help in a big way, and where we won't see the full value of Python models without some answer here.

Questions we need to answer:
- Is this "upload" step part of `dbt deps`? What if these reusable methods live in the root project, rather than a dbt package?
- How to support the transition from local &rarr; cloud development (if we want to preserve that workflow)?
- Those Python models should ideally be (a) pure functions with (b) thorough unit testing. Can dbt help here? Or should we encourage standard best practices Python development (`pytest` etc)?

@lostmygithubaccount @ChenyuLInx This feels like one of our highest-priority spikes during the v1.3 beta &rarr; final period.

b-per (approved):

runleonarun (approved):
Looks great!

PR #1822, opened 2022-08-02
fix formatting in [ Adapter-specific configs > Snowflake configurations ]

leoebfolsom:
Line breaks were causing the page to render funny with a newline and large space before the word "materialization." I think this change should fix it.

It's a little hard for me to tell for sure if this is doing what I want it to because github isn't rendering the `<Term>` tags in the preview.

The code ticks were also rendering improperly, and this seems to fix it.

## Description & motivation

This is what I'm trying to fix (see large line break and improperly rendered code ticks):

Screen Shot 2022-08-01 at 5 04 46 PM


dataders (approved):
sweet -- thanks

PR #1826, opened 2022-08-03
fix duplicated word in generic test doc

chris-s-friedman:
## Description & motivation

Documentation for why tests may be in `macros/` has the word "work" duplicated. Fixes this typo


runleonarun (approved):
Thanks for fixing this @chris-s-friedman!

PR #1827, opened 2022-08-03
Specify that Metadata API is Multi Tenant only

boxysean:
## Description & motivation
Requested by a Single Tenant customer, which reflects current reality


runleonarun (approved):
Look good @boxysean!

PR #1840, opened 2022-08-08
Fix a typo

farfromunique:
`create view`, `create view` => `create view`, `create table`

## Description & motivation
I was reading through the docs, and saw this typo, so I fixed it.


joellabes (approved):
Good spotting @farfromunique! Thanks for fixing this 🌟

PR #1845, opened 2022-08-09
Set `<text | json | default>` for `config.log_format` of `profiles.yml`.

nikitabarskov:
## Description & motivation

Update documentation for `config.log_format` to match the actual expected value.


nghi-ly (approved):
Great catch! Thanks for the fix! Appreciate it!

PR #1848, opened 2022-08-10
Snowpark setup third-party terms

jtcohen6:
Mention Snowpark / Anaconda third-party terms within "Additional Setup."

If we always require `pandas` to be installed (https://github.com/dbt-labs/dbt-snowflake/pull/229), then using dbt Python models _at all_ requires accepting third-party terms as a prerequisite.


jtcohen6:
This is a small change to beta docs that clarifies a point of confusion for a beta tester. I'm going to go ahead and merge as is, given the size of the current backlog

PR #1849, opened 2022-08-10
Update exclude.md with comments

AlexFrid:
## Description & motivation
- Added clarifying comments in line with other doc pages.
- Removed dbt test from the snapshot block to.

The reason for this is to:
- add clarity i.e. what does `--select my_package.*+ --exclude my_package.a_big_model+` mean in plain English
- remove confusion i.e. why is the dbt test command under the snapshots header?


nghi-ly (approved):
tysm for adding the descriptive comments! super helpful!

PR #1851, opened 2022-08-10
Update methods docs to include fqn definition

AlexFrid:
## Description & motivation
While through the dbt docs in preparation for the certification I noticed that I can’t find the definition of the “fqn” method anywhere in the docs.

The "fqn" method is however used in various places without an explicit explanation of what it means or what it does.
You can only implicitly see it through [examples like here](https://docs.getdbt.com/reference/node-selection/yaml-selectors#indirect-selection)

The purpose of this PR is to explicitly define what "fqn" means and how it fits in with the other methods, adding as an equivalent to the "file" method


foundinblank:
Just chiming in that I was wondering about `fqn` too in the context of the YAML Selectors page, and checked if anyone else had an open PR for this - came across this! +1 to this change.

nghi-ly (approved):
thanks for the adding this, @AlexFrid ! 🚢

PR #1862, opened 2022-08-15
Update tags to include tests

AlexFrid:
## Description & motivation

Adding tests to the list of commands you can use tags with. 

image


dataders (approved):
great callout!

PR #1866, opened 2022-08-16
Document adapter.quote()

joellabes:
## Description & motivation
A hidden function on the adapter object that I needed to reference but couldn't find.


dataders (approved):
great catch! :shipit:

PR #1889, opened 2022-08-22
Blog post: Narrative modeling

### LINKED ISSUE #1725: Narrative Modeling: How Structure can Tell a Story
for Developer Blog
# Intro

“When ***business questions*** come, they come not single spies, but in battalions.” - noted data professional William Shakespeare

This is, perhaps, a more collaborative way to describe the analytics engineer / stakeholder relationship than “if you give a mouse a cookie.’ 

After all, business questions *should* multiply, following the path from “What happened?” to “Why did that happen?” to “How can we predict what happens in the future?” and finally “How can we make something else happen instead?”

What does this look like in practice? Let’s take an example:

*How many orders came in last week?*

With a straightforward question, formal modeling principles may not be important. You have the source data you need and the information requested. Name it something that works and move on. It’s that simple.

It’s *not* that simple. 

*How many of the placed orders were fulfilled within three days?*

*How many items were in each order?*

*How many of the orders were returning customers?*

Business curiosity is the quintessential hydra (just look at a dbt DAG graph), and the analytics engineer should know better than to think a single head is the end of it. 

With an exponential growth in scope, however, comes a question of consistency. Modeling at scale is often easier with a structural framework. Specifically, principled naming conventions and database design help analytics engineers keep flow through stylistic decisions to focus on the request at hand, while providing other team members the ability to support follow-up questions on models they may not have built in the first place. 

# The Leading Solution

So what conventions should we use in our projects? 

History and gravity have pulled many a team towards dimensional modeling, building business molecules out of atomic fcts and dims. This approach benefits from well-deserved “if it ain’t broke” adoption as well as a commanding market share of data team practices. 

Within this framework, our business question would likely be answered from a table called `fct_orders`, because an order is, in a sense, an event. It might have a `customer_id` field that ties it to another model, `dim_customers`. 

These are actually common examples of a fact/dim structure, and if we have this tried-and-true approach, it may not make sense to seek another one. Any alternative would have to have similarly reliable conventions while offering some additional benefit to be worth the departure from the norm. 

But what if that additional benefit was clarity?

# Introducing Narrative Modeling

In the past, I’ve used a framework I call *narrative modeling*, naming and structuring data models so that they tell the story of the business. Unlike dimensional modeling, which was born out of a need to solve a technical problem (data storage), the principles of narrative modeling are people-oriented:

- Navigating an analytical database should be intuitive and educational
- Navigating the underlying code for an analytical database should be intuitive and stakeholder-friendly
- Adding to an analytical database should be intuitive and scaleable

Out of those principles came three primary conventions: event tables, details tables, and entity schemas.

## Event Tables

First, the backbone of a good narrative is the plot: what happened? In our example, *How many orders came in last week?*

To answer this question, we still need one row per order placed, with timestamps so that we can filter down to specific time periods (e.g. last week). But what should we name it?

Every principle of narrative modeling included the work “intuitive”. With that in mind, let’s call this table `order_placed`.

In narrative modeling, we’d call `order_placed`  an *event table*. Event tables have the naming convention of `subject_verb` . 

If this looks similar to `fct_` tables, it is! However, the benefit of `order_placed`over `fct_orders` appears when you need an additional order-level event. For instance, *How many orders were fulfilled last week?*

- In dimensional modeling, you already used `fct_orders`, so you likely would need to rename that model. This can result in naming conventions like `fct_order_placement` and `fct_order_fulfillment` which can feel like reporting on two separate entities rather than two things that can happen to orders.
- In narrative modeling, the sky’s the limit as we stand up `order_fulfilled` alongside `order_placed` with room for `order_cancelled` and all matter of other steps along an order’s flow.

Now we can model out the narrative thrust of an order, but events can be a little dry without a fleshed-out protagonist. How would we describe our hero, order CVRX1114?

## Details Tables

In narrative modeling, details tables are where a user can expect to find more descriptive information about a particular business entity. 

In the example of orders, an `order_details` table might contain fields like:

- customer_id
- items_ordered
- payment_method

Initially, it might seem as though these should just be fields on `order_placed` and, for those examples, it’s a reasonable point. Consider instead the following:

- items_fulfilled
- days_to_fulfillment

Once can imagine a stakeholder wanting to know all five of these data points, but the context for them comes from multiple events in the order’s flow. A`_details` table allows us to assemble a broad bank of knowledge about a given entity (in this case, an order) in one place. If data refresh is performant enough, this could even be the place to surface true snapshot fields like `status` or`current_location`.

So we have everything that happened to our order and everything we’d want to know about it. Now let’s put it all together…literally.

## Entity Schemas

I mentioned above that navigating an analytical database should be intuitive and educational. Given the building blocks of event and details tables described above, the intuitive part could be considered complete just by naming convention. If all of our data models were in a single schema, then the following tables would likely sort together:

- `order_cancelled`
- `order_details`
- `order_fulfilled`
- `order_placed`

Keeping related information grouped seems sufficient for usability. Then again, I’ve seen enough analytical databases where the database and schema have the same name (e.g. `DATA_MARTS.DATA_MARTS.*`) to see an opportunity for building a database that describes the business as soon as you open it. 

Throughout our example, `order` has been an entity we want to know a lot about. Any undertaking that creates business questions will have dozens if not hundreds of entities that people care about. Even in our example, we’ve glanced off other entities, such as `item` and `customer`. 

Grouping the models for each entity under a schema named for that entity builds out a database that describes the scope of the business in simple terms, while also allowing stakeholders using the database to navigate more purposefully, expanding schemas rather than scrolling tables in a single mega-schema.

- `analytics` database
    - `customer` schema
        - `customer_details`
        - `customer_registered`
    - `item` schema
        - `item_details`
        - `item_fulfilled`
    - `order` schema
        - `order_cancelled`
        - `order_details`
        - `order_fulfilled`
        - `order_placed`

This structuring principle can also help address the boogeyman of business entity resolution, overlapping names. If you’re a business that orders parts regularly and then fulfills customer orders, the question *How many orders came in last week?* can get a lot more confusing. At a fast-moving company, a new hire could be asked the question and answer in the wrong context completely, because the tenured stakeholder can no longer imagine mixing them up. 

If instead, the database had schemas for`parts_order` and`customer_order`, that same new hire hits the database, sees those schemas, and thinks “Oh, there are two types…I should probably ask which.” That distinction can be a lot harder to spot in a single analytics schema.

Intuitive *and* educational. 

# Benefits In Practice

- Stakeholders find it easier to participate in data modeling
    - They don’t need to learn the idiosyncrasies of fct/dim structuring (e.g. idea that only some business datapoints are called facts).
    - The discrete models are framed as events or entities in the business, which means the stakeholder can compare the model’s logic to their subject matter expertise.
- Intangible business steps are easier to model
    - Knowledge gaps are captured accurately. For instance, if the best way you know that a shipment was received by a customer is that a truck driver scanned it out of the system, you can model `shipment_scanned_out` as an explicit model, followed by an implicit`shipment_received` model referencing it. This stores in code the company’s current point-of-view that the scanning action is the best information available
    - Certain business decisions directly drive data transformations. If an entire package costs $50.00 to ship and it has multiple items inside, shipping cost could be attributed to each item via weight or product value. In either case, teams can capture this attribution as `item_apportioned_shipping_cost`.
    
    # Should Your Team Use It?
    
    Narrative modeling primarily values comprehension as an outcome of data modeling, which can be a high priority for…
    
    - …companies with a high stakeholder-to-data-team-member ratio.
    - …companies with SQL-savvy stakeholders
    - …companies looking to on-board new team members at a high rate (as the project in this case is a sketch of the business itself)
    
    Narrative modeling might not be the right fit, however, if storage spend, even in cloud warehouses, needs to be closely tracked. After all, fct/dim models emerged from a need to optimize data storage. 
    

# In Closing

There’s a classic movie called Desk Set in which Katharine Hepburn runs the reference desk of a major TV network and Spencer Tracy comes in to install a research computer. In gladiatorial bouts of banter, the movie investigates the concept of knowledge and, in particular, how human it should be. At one point, Tracy’s Richard Sumner runs Hepburn’s Bunny Watson through a chunky “train left the station” word problem and I think about her response (abbreviated below) all the time. 

**Richard Sumner:** Now, a train started out at Grand Central, with seventeen passengers aboard and a crew of nine. At 125th Street, four got off and nine got on. At White Plains, three got off and one got on. At Chappaqua, nine got off and four got on. And at each successive stop thereafter, nobody got off, nobody got on till the train reached its next-to-the-last stop, where five people got off and one got on. Then it reached the terminal.

**Richard Sumner:** How many people got off at Chappaqua?

**Bunny Watson:** Nine.

**Richard Sumner:** Uh, would you mind telling me how you arrived at that conclusion?

**Bunny Watson:** Spooky, isn't it? Did you notice that there are also nine letters in Chappaqua?

**Richard Sumner:** Are you in the habit of associating words with the number of letters in them?

**Bunny Watson:** I associate many things with many things.

**Richard Sumner:** I see. Hmm.

**Bunny Watson:** Aren't you going to ask me how many people got off at White Plains? Three.

**Richard Sumner:** But there are ten letters in White Plains.

**Bunny Watson:** No. Eleven.

**Richard Sumner:** [beat] But only three got off there.

**Bunny Watson:** You see, I've only ever been to White Plains three times in my whole life.

Much like Mr. Sumner, we would be hard-pressed to teach a computer to answer questions the way Bunny does. Her accumulated knowledge and how she accesses it is deeply human. Why, then, do we so often take the accumulated knowledge of subject matter experts across our business and abstract it AWAY from its context to capture it in a data model? Put another way, why do we store business answers in a way that, over time, forgets the questions? 

Narrative modeling can help structure the data with questions and answers stored side-by-side, allowing us to be reference librarians with a killer card catalog. We can model out the expertise of the finance team or the marketing team or the product team in their own words, which keeps the context for not only the next person to join *those* teams, but also the next person to join our own. In this way, in the face of battalions of business questions, we can become a host unto ourselves.

ian-fahey:
## Description & motivation

This adds a new blog post for the Narrative Modeling article. The Notion version of this is available [here](https://www.notion.so/dbtlabs/Narrative-Modeling-069eb895a0714572b61e9027e75f7511)

## Checklist

Closes #1725


ian-fahey:
@KiraFuruichi I think I did as much as I can.

Things I couldn't get to work were my in-line images and I guess any of the failed checks above...

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -0,0 +1,236 @@
+---
+title: "Narrative modeling: how structure can tell a story
```
KiraFuruichi:
```suggestion
title: "Narrative modeling: How structure can tell a story"
```

KiraFuruichi: I think this is causing some of the issues!

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Narrative modeling: How structure can tell a story"
 description: "Analytics engineers work in the space between technical and business understanding, but tried-and-true modeling patterns often default to the technical. In this article, Ian Fahey makes a case for a pattern of modeling that follows the narrative of the business it represents."
-slug: narrative-modeling
+slug: narrative-modeling"
```
KiraFuruichi: ah there should be no quote here! just `slug: narrative-modeling`

ian-fahey: Updated that...is the issue that I have so many quotes in the piece? Do I need to escape them somehow?

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -0,0 +1,236 @@
+---
+title: "Narrative modeling: How structure can tell a story"
+description: "Analytics engineers work in the space between technical and business understanding, but tried-and-true modeling patterns often default to the technical. In this article, Ian Fahey makes a case for a pattern of modeling that follows the narrative of the business it represents."
+slug: narrative-modeling
+authors: [ian_fahey]
+tags:
```
KiraFuruichi: ah this is also a missing tag here!
```suggestion
tags: [analytics craft]
```

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -0,0 +1,236 @@
...
+date: 2022-08-22
+is_featured: true
+---
+
+The larger a data ecosystem gets, the more its users and stakeholders expect consistency. As the ratio of data models to team members (to say nothing of stakeholders to team members) skyrockets, an agreed-upon modeling pattern often acts as scaffolding around that growth. 
+
+The biggest tool in the toolbox today, dimensional modeling, offers enough consistency to make it the dominant approach in the space, but what might be possible if we shut that toolbox, took a break from our workbench, and instead strolled over to our bookshelf?
+
+In other words, what if we told a story?
+
```
KiraFuruichi:
```suggestion
<!--truncate-->
```

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -0,0 +1,236 @@
...
+- `customer_id`
+- `items_ordered`
+- `payment_method`
+
+Initially, it might seem as though these should just be fields on `order_placed` and, for those examples, it’s a reasonable point[^1]. Consider instead the following:
+
+- `items_fulfilled`
+- `days_to_fulfillment`
+
+One can imagine a stakeholder wanting to know all five of these data points, but the context for them comes from multiple events in the order’s flow. A`_details` table allows us to assemble a broad bank of knowledge about a given entity (in this case, an order) in one place. If data refresh is performant enough, this could even be the place to surface true snapshot fields like `status` or`current_location`.
```
KiraFuruichi:
```suggestion
One can imagine a stakeholder wanting to know all five of these data points, but the context for them comes from multiple events in the order’s flow. A `_details` table allows us to assemble a broad bank of knowledge about a given entity (in this case, an order) in one place. If data refresh is performant enough, this could even be the place to surface true snapshot fields like `status` or`current_location`.
```

website/blog/2022-08-22-narrative-modeling.md
```diff
@@ -0,0 +1,236 @@
...
+
+**Richard Sumner:** Uh, would you mind telling me how you arrived at that conclusion?
+
+**Bunny Watson:** Spooky, isn't it? Did you notice that there are also nine letters in Chappaqua?
+
+**Richard Sumner:** Are you in the habit of associating words with the number of letters in them?
+
+**Bunny Watson:** I associate many things with many things.
+
+Richard Sumner: I see. Hmm.
```
KiraFuruichi: all of Richard/Bunny are un bolded after this line. Is that on purpose?

ian-fahey: Not intentional. Fixed!

KiraFuruichi (approved):
This is getting the ✅ from me! Fantastic, thoughtful, and incredibly well-written first developer blog post Ian! I'm incredibly excited for folks to read through this and use it in pair with the "How we Structure our dbt Projects" Guide.

Feel free to merge this in whenever you want, just ensure it's not behind `current`! I have promotion for this post in Twitter and LinkedIn scheduled for later next week, but if you feel like it, feel free to share it in the #i-made-this channel in the Community Slack. I'll also be sharing it in #devhub channel later this week 😄

ian-fahey:
Checked that this is still commits AHEAD of current! Merging now!

PR #1891, opened 2022-08-22
Fix typo and update for more specific permission needed to run dbt command

tovganesh:


dataders (approved):
thanks for the careful improvement. this will definitely benefit end users

PR #1901, opened 2022-08-23
update pagination_next and pagination_prev

thorn14:
Make the pagination match the steps in the tutorial rather than the order of docs by referencing id

## Description & motivation
Updating the pagination links to follow the order of the getting-started guide rather than follow the integration order.


runleonarun (approved):
This looks great! 😍 I had no idea we could leverage the frontmatter like that!

PR #1912, opened 2022-08-25
Updated the apache spark adapter page to add Apache Hudi support

vingov:
## Description & motivation
Recently, I added the support for apache hudi table format to the dbt-spark adapter. This is the PR: https://github.com/dbt-labs/dbt-spark/pull/210

This feature has already been released, there are few community members asking for docs, hence I have updated the Apache spark adapter page with these changes.

There is already a blog on how to use this feature on the apache hudi community site: 
https://hudi.apache.org/blog/2022/07/11/build-open-lakehouse-using-apache-hudi-and-dbt

In addition to this change, I plan to create a separate page to add more details on the apache hudi + dbt integrations and all the various options this feature supports.

The reason, I updated this apache spark adapter page is to give another option for users who wish to use merge/update operations without the dependency of Databricks run time, with just an open source vanilla instance of apache spark.

## Prerelease docs
not applicable

## Checklist
not applicable

### DIFF
--- a/website/docs/reference/resource-configs/spark-configs.md
+++ b/website/docs/reference/resource-configs/spark-configs.md
@@ -12,13 +12,13 @@ To-do:
 
 When materializing a model as `table`, you may include several optional configs that are specific to the dbt-spark plugin, in addition to the standard [model configs](model-configs).
 
-| Option  | Description                                        | Required?               | Example                  |
-|---------|----------------------------------------------------|-------------------------|--------------------------|
-| file_format | The file format to use when creating tables (`parquet`, `delta`, `csv`, `json`, `text`, `jdbc`, `orc`, `hive` or `libsvm`). | Optional | `parquet`|
-| location_root  | The created table uses the specified directory to store its data. The table alias is appended to it. | Optional                | `/mnt/root`              |
-| partition_by  | Partition the created table by the specified columns. A directory is created for each partition. | Optional                | `date_day`              |
-| clustered_by  | Each partition in the created table will be split into a fixed number of buckets by the specified columns. | Optional               | `country_code`              |
-| buckets  | The number of buckets to create while clustering | Required if `clustered_by` is specified                | `8`              |
+| Option  | Description                                                                                                                        | Required?               | Example                  |
+|---------|------------------------------------------------------------------------------------------------------------------------------------|-------------------------|--------------------------|
+| file_format | The file format to use when creating tables (`parquet`, `delta`, `hudi`, `csv`, `json`, `text`, `jdbc`, `orc`, `hive` or `libsvm`). | Optional | `parquet`|
+| location_root  | The created table uses the specified directory to store its data. The table alias is appended to it.                               | Optional                | `/mnt/root`              |
+| partition_by  | Partition the created table by the specified columns. A directory is created for each partition.                                   | Optional                | `date_day`              |
+| clustered_by  | Each partition in the created table will be split into a fixed number of buckets by the specified columns.                         | Optional               | `country_code`              |
+| buckets  | The number of buckets to create while clustering                                                                                   | Required if `clustered_by` is specified                | `8`              |
 
 ## Incremental models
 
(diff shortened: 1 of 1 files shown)


vinothchandar (approved):
If its helps,the hudi specs are here. https://hudi.apache.org/tech-specs

website/docs/reference/resource-configs/spark-configs.md
```diff
@@ -28,12 +28,12 @@ When materializing a model as `table`, you may include several optional configs
 
 </Changelog>
 
-dbt seeks to offer useful, intuitive modeling abstractions by means of its built-in configurations and <Term id="materialization">materializations</Term>. Because there is so much variance between Apache Spark clusters out in the world—not to mention the powerful features offered to Databricks users by the Delta file format and custom runtime—making sense of all the available options is an undertaking in its own right.
+dbt seeks to offer useful, intuitive modeling abstractions by means of its built-in configurations and <Term id="materialization">materializations</Term>. Because there is so much variance between Apache Spark clusters out in the world—not to mention the powerful features offered to Databricks users by the Delta file format and custom runtime or Hudi file format with Apache Spark runtime —making sense of all the available options is an undertaking in its own right.
```
dataders: i feel this sentence is now trying to accomplish to many things. I understand why you want to mention Hudi alongside Delta, but maybe we can break it into two to three standalone sentences?

website/docs/reference/resource-configs/spark-configs.md
```diff
@@ -191,19 +191,22 @@ insert overwrite table analytics.spark_incremental
 
 
 **Usage notes:** The `merge` incremental strategy requires:
-- `file_format: delta`
-- Databricks Runtime 5.1 and above
+- `file_format: delta or hudi`
+- Databricks Runtime 5.1 and above for delta file format
+- Apache Spark for hudi file format
```
dataders, line 198: naive question do Databricks Spark clusters support Hudi?

website/docs/reference/resource-configs/spark-configs.md
```diff
@@ -191,19 +191,22 @@ insert overwrite table analytics.spark_incremental
...
 
 dbt will run an [atomic `merge` statement](https://docs.databricks.com/spark/latest/spark-sql/language-manual/merge-into.html) which looks nearly identical to the default merge behavior on Snowflake and BigQuery. If a `unique_key` is specified (recommended), dbt will update old records with values from new records that match on the key column. If a `unique_key` is not specified, dbt will forgo match criteria and simply insert all new records (similar to `append` strategy).
 
 <Tabs
-  defaultValue="source"
+  defaultValue="delta_source"
```
dataders: I understand why you added new tabs (and props for running with this docusaurus specific convention!). But since the difference is small, I propose generalizing the existing source and run code examples to accommodate both with the following changes:
1. for the source code tab:
    1. so in the source code mention both formats
        ```yml
        file_format='delta', # or 'hudi'
        ```
    2. change source file name to `file_incremental.sql` (or `lake_`? idk)
1. for the run code tab update the filename and model name to match whatever you decided previously

website/docs/reference/resource-configs/spark-configs.md
```diff
@@ -320,3 +391,28 @@ snapshots:
 ```
 
 </File>
+
+
+## Apache Hudi configurations
```
dataders: Same note as above, I feel as if this section can be generalized to avoid duplication.

dataders (changes requested):
Thanks for the PR! My comments are mostly about synthesizing docs about the two files formats rather than duplicating. One more question: are there any important distinctions in usage & behavior that users of dbt-spark should take into account? If everything's functionally the same, all the more reason to document the file format feature only once. That would be the only addition I could suggest.

website/docs/reference/resource-configs/spark-configs.md
vingov: Yep, good point, updated into two sentences.

website/docs/reference/resource-configs/spark-configs.md
vingov: We haven't tested Hudi on Databricks Spark cluster, theoretically, it should work unless there are any specific differences from vanilla spark runtime. To be on the safe side, we are recommending users use only Spark runtime.

website/docs/reference/resource-configs/spark-configs.md
vingov: Yes, even though I initially thought of it and then ignored it to be more precise, it makes sense to use single source code with hudi on comment since the rest of the code remains the same.

website/docs/reference/resource-configs/spark-configs.md
vingov: good point, updated it.

vingov:
@dataders - Thanks for the quick review, I've addressed all your comments, ptal thanks!

dataders (approved):
thanks @vingov 💪🏻

PR #1913, opened 2022-08-26
Fix the blog of unit testing dbt packages

yu-iskw:
Let me fix a minor thing in the blog article of unit testing dbt packages.


jasnonaz (dismissed):
Thanks for catching!

dbeatty10:
Taking a look at this, there's actually one more typo to fix here as well -- I'm guessing the intent was `postgres__to_literal` instead of `postgress__to_literal`:

image

I'm going to try making that fix and then merge it afterwards.

yu-iskw:
@dbeatty10 Thank you for following up.

dbeatty10 (approved):

PR #1931, opened 2022-08-31
Update `snowflake-profile.md`

one-data-cookie:
# What
Provide more realistic examples for `account` in `snowflake-profile.md`

# Why
When trying to connect with `abc123.eu-west-1.aws`, connection fails. The required format for AWS EU West 1, i.e. EU (Ireland), is `abc123.eu-west-1`, so without `.aws`. Unlike `abc123.eu-west-2.aws`, however, which is a required format for Europe (London).

Keeping both examples to avoid confusion by end users. dbt Cloud doesn't return any error when `account` is incorrect whilst connecting Snowflake database.

## Prerelease docs

## Checklist


dataders (approved):
great catch. i agree there should be two references. if you're interested in a future PR, I think that this info would be much better presented in a table format.

PR #1942, opened 2022-09-06
update installation method for dbt-athena

nialloriordan:
## Description & motivation

The dbt Athena adapter now publishes to pypi and no longer requires pip installation from git repo

pypi: https://pypi.org/project/dbt-athena-adapter/
updated installation from dbt-athena git repo: https://github.com/Tomme/dbt-athena#installation


dataders (approved):
thanks @nialloriordan

PR #1969, opened 2022-09-09
Create events page

JKarlavige:
This adds an events page which includes the AddEvent embed.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -852,3 +852,8 @@ footer #ot-sdk-btn.optanon-show-settings:hover{
   color: var(--ifm-footer-color);
   text-decoration: underline;
 }
+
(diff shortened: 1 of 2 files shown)


krevitt (changes requested):
Looks great!  I added copy (H1 + paragraph) in a commit.

How do we add a meta desc + modify the title in this .js file format?  If you point me to an example I can add it in.

JKarlavige:
> Looks great! I added copy (H1 + paragraph) in a commit.
> 
> How do we add a meta desc + modify the title in this .js file format? If you point me to an example I can add it in.

@krevitt Added the tags to update the title & description. I added them with placeholder content, so once the content is updated those are g2g

krevitt (approved):
Lets go!

PR #1999, opened 2022-09-13
Add missing semicolon in example

msardana94:
## Description & motivation
Missing semicolon throws an error similar to `syntax error line 3 at position 6 unexpected 'create'` since `sql_header` clause just gets prepended to `create` block.
I am not sure if there's a need to explicitly mention a note about the same in docs.


dataders (approved):
thanks @msardana94! though worth noting that, on some data platforms, SQL strings are auto-parsed and one need not demarcate statement termination with `;`. But they never hurt, so this change improves things!

PR #2004, opened 2022-09-13
Discourse integration

JKarlavige:
## Description & motivation

Asana task: https://app.asana.com/0/1200099998847559/1202844016314445/f

This adds a new `DiscourseFeed` component to be used throughout the site. This component uses the Discourse API to display topics on docs pages. A new `/forum` page is created to display different sections of Discourse topics via the new component.

This also includes an example of how the component can be used on any docs pages.

I will be doing the website work within this PR, and will open a separate PR to handle the Algolia search side of this task. I expect the Algolia work to take longer so no reason to hold up this work.

## Preview
Forum page: https://deploy-preview-2004--docs-getdbt-com.netlify.app/forum

## TODOs
- [x] Remove example use of Discourse component from Jinja docs page


website/docs/faqs/Jinja/which-jinja-docs.md
```diff
@@ -10,3 +10,6 @@ If you are stuck with a Jinja issue, it can get confusing where to check for mor
 1. [Jinja's Template Designer Docs](https://jinja.palletsprojects.com/page/templates/): This is the best reference for most of the Jinja you'll use
 2. [Our Jinja function reference](/docs/building-a-dbt-project/jinja-macros#related-reference-docs): This documents any additional functionality we've added to Jinja in dbt.
 3. [Agate's table docs](https://agate.readthedocs.io/page/api/table.html): If you're operating on the result of a query, dbt will pass it back to you as an agate table. This means that the methods you call on the <Term id="table" /> belong to the Agate library rather than Jinja or dbt.
+
+<!-- EXAMPLE: DELETE BEFORE TAKE-LIVE -->
+<DiscourseFeed title='Discourse Jinja topics' tags='jinja' styles={{marginTop: 40}} />
```
joellabes: Could the top margin be baked into the default template? I suspect that it's something we'll always want.

website/functions/get-discourse-posts.js
```diff
@@ -0,0 +1,100 @@
...
+    
+    if(!topics)
+      throw new Error('Unable to get results from api request.')
+
+    // Return posts 
+    return await returnResponse(200, topics)
+  } catch(err) {
+    // Log and return the error
+    console.log('err', err)
+    return await returnResponse(500, { error: 'Unable to get events from Discourse.'})
```
joellabes: I assume this is copy-paste from the events feed
```suggestion
    return await returnResponse(500, { error: 'Unable to get topics from Discourse.'})
```

website/functions/get-discourse-posts.js
```diff
@@ -0,0 +1,100 @@
...
+  if(query) {
+    const encodedQuery = encodeURIComponent(query)
+    return encodedQuery
+  }
+}
+
+function validateItem({ key, value }) {
+  // predefined Discourse values
+  // https://docs.discourse.org/#tag/Search/operation/search
+  const inStringValues = ['title', 'likes', 'personal', 'messages', 'seen', 'unseen', 'posted', 'created, watching', 'tracking', 'bookmarks', 'assigned', 'unassigned', 'first', 'pinned', 'wiki']
```
joellabes:
```suggestion
  const inStringValues = ['title', 'likes', 'personal', 'messages', 'seen', 'unseen', 'posted', 'created', 'watching', 'tracking', 'bookmarks', 'assigned', 'unassigned', 'first', 'pinned', 'wiki']
```

joellabes: I don't know which user this API key is associated with, but it's probably unnecessary to support things like personal, seen, unseen, created as they're user-specific. I guess they're not doing any harm? 

If we wanted to cut it down, I think the only ones that are generic are `title, first, pinned, wiki`.

website/functions/get-discourse-posts.js
```diff
@@ -0,0 +1,100 @@
...
+  for(const [key, value] of Object.entries(JSON.parse(body))) {
+    // validate categories
+    // if valid, add to query string
+    if(validateItem({ key, value })) {
+      if(key === 'category') {
+        query += `#${value} `
+      } else if(key === 'inString') {
+        query += `in:${value}`
+      } else {
+        query += `${key}:${value} `
```
joellabes: Does MDX allow duplicate keys? You're allowed to do something like [`status:open status:solved`](https://discourse.getdbt.com/search?q=status%3Aopen%20status%3Asolved), but this feels like it would break: 

`<DiscourseFeed category='help' status='solved' status='open' />`

It's not likely that we'd do this very often, so having a string instead of an array as the main input type is more ergonomic. Can you pass arrays into MDX if needed (`status=['solved', 'open']`)? Or would we have to comma-separate them and you split them out?

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <div className={feedStyles.discoursePosts} style={{minHeight: setMinHeight, ...styles}}>
+      {title && (
+        <h2>{title}</h2>
+      )}
+      {loading ? (
+        <img src="/img/loader-icon.svg" alt="Loading" className={feedStyles.loadingIcon} />
+      ) : isError || !posts?.length > 0 ? (
+        <p>Unable to load Discourse posts at this time.</p>
```
joellabes:
```suggestion
        <p>Unable to load forum posts at this time.</p>
```

I think we'll de-emphasise the platform we're using and lean into it just being the Forum (in the UI at least; the code can still be specific)

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+  min_posts = undefined,
+  max_posts = undefined,
+  min_views = undefined,
+  max_views = undefined,
+  tags = undefined, 
+  term = undefined,
+  category = undefined,
+  title = undefined,
+  link_text = "See latest topics",
+  link_href = "https://discourse.getdbt.com/",
```
joellabes: As the default, it would be good for the CTA to be `Ask the Community` which goes to `https://discourse.getdbt.com/new-topic?category=XXXX&tags=comma,separated`

Note that in search you can either use `,` in tags (for any of) and `+` (for all of), so you'd need to replace +s with ,s here too

https://meta.discourse.org/t/how-to-create-a-new-topic-link-with-pre-filled-information/28074

Then we could override it on the forum homepage to be the Read More ones you currently have. (Side note: are we allowed to set them to null/undefined to cause no button to be shown?)

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
+import React, { useState, useEffect } from 'react'
+import axios from 'axios'
+import feedStyles from './styles.module.css';
+
+export default function DiscourseFeed({
+  order = 'latest_topic',
+  status = undefined,
+  after = undefined,
```
joellabes: Can we have relative periods here? I was thinking we'd do something like last 30 days for the help ones, and last 90 for discussions.

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+      )}
+      {loading ? (
+        <img src="/img/loader-icon.svg" alt="Loading" className={feedStyles.loadingIcon} />
+      ) : isError || !posts?.length > 0 ? (
+        <p>Unable to load Discourse posts at this time.</p>
+      ) : (
+        <ul>
+          {posts.map(post => (
+            <li>
+              <PostWrapper post={post}>{post.title}</PostWrapper>
```
joellabes: As they currently look, there's not a lot of social proof that the answers you click through to are going to be accurate. Would it be overkill to put a ✅ with `Solved` hover text in the front of solved topics? (I think it's called `has_accepted_answer` or something in the response body)

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+              {/* <span> {post.username && `- by ${post.username}`}</span> */}
+              {post?.username || post?.posts_count && (
+                <>
+                  {' '}-
+                  <span>
+                    {post?.username && `by ${post.username}${post?.posts_count && ','}`}
```
joellabes: If you wanted to get the username without a second API call, you can get it from the `posts` part of the search response. 
image

You'll want to only take it if the `post_number` is 1. When we're searching for topic-wide metadata (e.g. tag, category, solved status) this is fine - that'll always be the one you're given. It's only if you search for specific text that you might get a different post - e.g. this one, where you get the first post in a topic that contains the search text. We don't intend to do that for any of our embeds AFAIK. 
image

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+                    {' '}
+                    {post?.posts_count && `${post.posts_count} comments`}
```
joellabes: Arguably this could be `{posts_count - 1}` replies. Not attached to this. 

Side note: `1 comments` is my biggest UI pet peeve. (Very attached to this 😅)

joellabes (changes requested):
This is looking so great @JKarlavige 🤩😍

A handful of comments, hopefully nothing too arduous!

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
...
+  after = undefined,
+  before = undefined,
+  inString = undefined, 
+  min_posts = undefined,
+  max_posts = undefined,
+  min_views = undefined,
+  max_views = undefined,
+  tags = undefined, 
+  term = undefined,
+  category = undefined,
```
joellabes: Can we default this to `'help'` so we don't have to specify it everywhere? That's going to be the overwhelming majority of embeds.

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,122 @@
+import React, { useState, useEffect } from 'react'
+import axios from 'axios'
+import feedStyles from './styles.module.css';
+
+export default function DiscourseFeed({
+  order = 'latest_topic',
+  status = undefined,
```
joellabes: Likewise here, let's have status:solved as the default and override it where needed. 

OR, it might be easier to have a focused `DiscourseHelpFeed` which ultimately creates a `DiscourseFeed` but we don't have to put defaults into the main constructor. Is that practical?

joellabes (commented):
Was just writing a comment in Asana about how we only have to specify the tags we want in help mode, and then thought "hang on a minute, is that accurate?"

website/functions/get-discourse-posts.js
JKarlavige: Yeah that's a good point, i'll adjust this to support multiple values for the same property.

website/functions/get-discourse-posts.js
JKarlavige: Yeah it doesn't hurt having those on our end as it only ensures the values will work with Discourse. But if you would like to prevent those other values from being used, we can remove them from the validation here. Will leave that up to you!

website/src/components/discourse/index.js
JKarlavige: Updated the default text & link for the CTA. If the `category` or `tags` are passed in, it will append them to the link url by default. If a custom `link_href` is passed in, it will override the default.

For the tags, since it's a string with comma (or +) separated values, passing a string such as `<DiscourseFeed  tags='jinja,testing' />` will work as i'm passing the tags string directly into the query.

Also added a separate `hide_cta` prop to control whether the cta should be hidden.

website/src/components/discourse/index.js
JKarlavige: Love this

website/src/components/discourse/index.js
JKarlavige: Does this refer to the `after` property? If a different category is passed in other than `help` or `discussions` should we exclude the `after` date, or should we set a default number of days back?

I pushed an update to set a relative `after` date if the category is `help` or `discussions`, otherwise it excludes the `after` date altogether for now.

website/src/components/discourse/index.js
JKarlavige: Set the default status to `solved`. We can create a more specific component which uses the original component with specific props if you'd like. If you feel like it would be easier for the team to work with we can get it setup. I would just want to go over all the settings that more specific component would need.

website/src/components/discourse/index.js
JKarlavige: ahhhh good find! Will get this in there

website/functions/get-discourse-posts.js
JKarlavige: Updated this to support arrays for the `status` property. The docs page example shows how multiple statuses can be passed in: https://github.com/dbt-labs/docs.getdbt.com/blob/2590c0049fcdd15eb0005f10cd4c625f799d2528/website/docs/faqs/Jinja/which-jinja-docs.md

`status` can now support string: `status='open'` or arrays: `status={['open', 'public']}` when being passed into the component

JKarlavige:
@joellabes marked comments as resolved for which i pushed updates for. I have a few open questions/clarifications for the others!

website/functions/get-discourse-posts.js
joellabes: I think let's remove them so that it's clear to future users which ones they can realistically use

website/src/components/discourse/index.js
joellabes: Yes I meant it for after - will have a look at the new code

lfracalossiFA:
Hi @JKarlavige,
Is the "-" between the thread title and the byline by design? I don't love it. But I also don't know if this is something that was requested!

Screen Shot 2022-09-19 at 4 40 00 PM

lfracalossiFA:
The fonts also seem small to me on the list items on /forum. (I think it's the overall docs font face; I feel like it skews small). Any objection to bumping those up to 18px? I'd also suggest making the title bold, eg:

Screen Shot 2022-09-19 at 4 47 38 PM

website/src/components/discourse/styles.module.css
```diff
@@ -0,0 +1,45 @@
...
+.discourseTopics ul {
+  list-style: none;
+  margin: 0 0 1rem;
+  padding: 0;
+  width: 100%;
+}
+.discourseTopics ul li {
+  padding: 1rem 0;
+  border-bottom: 2px solid var(--color-off-white);
+}
```
lfracalossiFA, line 26: I mentioned this in the discussion section of this PR, but I think we should bump up the font here to 18px, and make the `a` bold. 

(Maybe just for /forum? Not sure it would be appropriate in the in-page component, because it might distract from the doc content itself).

website/src/components/discourse/styles.module.css
```diff
@@ -0,0 +1,45 @@
...
+.discourseTopics ul li:last-of-type {
+  border-bottom: none;
+}
+
+/* Cta */
+.discourseCta {
+  font-size: 1.1rem;
+  margin-top: auto;
+}
```
lfracalossiFA: I feel like the font size and weight are inconsistent with buttons like this elsewhere on the site? (Compare to homepage). But maybe there is a different button you based this off?

lfracalossiFA:
It looks like the meta description is "Description here." Any way to customize that? 

Screen Shot 2022-09-19 at 5 18 22 PM

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    fetchData()
+  }, [])
+
+  // Set initial min-height
+  // This is to avoid layout shifts
+  // which affects Lighthouse performance scores
+  const setMinHeight = isError
+    ? 'auto'
+    : 414
```
lfracalossiFA, line 101: 👍

website/src/pages/forum.js
```diff
@@ -0,0 +1,25 @@
...
+    <Layout>
+      <Head>
+        <title>dbt Discourse</title>
+        <meta name="description" content="Description here" />
+      </Head>
+      <section className='discourse-forum-page'>
+        <div className='container'>   
+          <h1>dbt Discourse</h1>
+          <DiscourseFeed title='Open topics' category='help' status='unsolved'  link_text='See open topics' />
+          <DiscourseFeed title='Cool Problem Solving' category='show-and-tell' after='2022-06-14' status='unsolved' link_text='See more topics' link_href='https://discourse.getdbt.com/c/show-and-tell/22' />
```
lfracalossiFA: I saw the conversation about the `after` attribute with Joel above -- will we want a date like this to be hard-coded here? Or is this just an example?

lfracalossiFA (dismissed):
Overall, this is really looking good -- will add great value to our pages. I have a few questions as well as requests re: styling/UI. (I didn't see a comp anywhere, so feel free to ignore my suggestions if they contradict any stakeholder requests).

website/src/components/discourse/index.js
joellabes: I had originally thought it might be another prop on the DiscourseFeed, but I don't mind how you've done it here! 

Everything that isn't `help` can be 90 days - we might wind up reconsidering this at some point but messing around with the params is the sort of thing I can change unassisted 🎉

website/src/components/discourse/index.js
```diff
@@ -34,41 +84,64 @@ export default function DiscourseFeed(styles = {}) {
...
+      ) : isError || !topics?.length > 0 ? (
+        <p>Unable to load forum topics at this time.</p>
+      ) : (
+        <ul>
+          {topics.map(topic => (
+            <li key={topic.id}>
+              {topic?.has_accepted_answer && (
+                <span className={feedStyles.solvedTopic} title="Solved">✅ </span>
+              )}
+              <TopicWrapper topic={topic}>{topic.title}</TopicWrapper>
```
joellabes, line 124: Also on the social proof front, it'd be nice to pull through the `like_count` from the posts array. I think this will have more purpose on the Discussion ones as I doubt people are going to click like on the OP of a support request, but if they do we can show them there too! 

I don't know how you render them - it might get messy to have `How do I do computers? by joellabes, 1 reply, 3 likes` but if we don't have design time then that might just have to do for now.

website/src/pages/forum.js
joellabes: I doubt we'd ever hardcode a date in here, and instead would just use the relative calculation 👍

website/src/pages/forum.js
```diff
@@ -0,0 +1,25 @@
...
+          <DiscourseFeed title='Interesting Discussions' category='discussions'  status='unsolved' link_text='See discussions' link_href='https://discourse.getdbt.com/c/discussions/21' />
```
joellabes: discussions and show and tell don't have Mark as Solved enabled, so the status is redundant here

joellabes (changes requested):
Yeah this is all basically done from my perspective! Handful of tweaks but then we're good to go

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,147 @@
...
+  max_posts = undefined,
+  min_views = undefined,
+  max_views = undefined,
+  tags = undefined, 
+  term = undefined,
+  category = 'help',
+  title = undefined,
+  link_text = "Ask the Community",
+  link_href = `https://discourse.getdbt.com/new-topic${category ? `?category=${category}` : ''}${tags ? (!category ? `?tags=${tags}` : `&tags=${tags}`) : ''}`,
+  hide_cta = false,
```
joellabes: Isn't it more standard to have positive flags, so it'd be `show_cta = true`? Non-blocking

website/src/components/discourse/index.js
joellabes: Yeah I think it would be good to make a focused one. The default props: 
```javascript
order = 'latest_topic',
status = 'solved',
category = 'help',
link_text = 'Ask the Community',
link_href = // the same URL builder you've already got and I don't want to copy-paste
hide_cta = 'false'
```

Then on the main `DiscourseFeed`, everything would be undefined except for `hide_cta` which would be true. (Or, if you [changed it to be `show_cta`](https://github.com/dbt-labs/docs.getdbt.com/pull/2004#discussion_r974851260), then _everything_ could be undefined because `undefined` is falsy right?)

website/src/pages/forum.js
lfracalossiFA: @joellabes That's why I'm wondering about the `after='2022-06-14'` on Cool Problem Solving...

website/src/components/discourse/styles.module.css
JKarlavige: Hmmmm, this button is the same one used on the homepage cards, i'm using the same `button button--primary` classes. The only thing I did was reduce the font-size a bit as i felt it was too large. Is there anything specific you think we should change with the buttons for this component?

website/src/components/discourse/styles.module.css
JKarlavige: I bumped the font to `18px` and make it bold. I'm not crazy about having bold text in one place where the component is used, and not bold in another spot. We'd have to add another prop to control that, and there's already 18 potential props that can be used. 😅 Maybe `font-weight: 600` is a good in-between? I'll have that in my next push

JKarlavige:
> It looks like the meta description is "Description here." Any way to customize that?
> 
> Screen Shot 2022-09-19 at 5 18 22 PM

cc @joellabes, I added title and description meta tags to the /forum page. What should the meta title and description be for the page?

website/src/pages/forum.js
JKarlavige: Removed the `after` property for the Cool Problem Solving component.

website/src/components/discourse/styles.module.css
lfracalossiFA: > Is there anything specific you think we should change with the buttons for this component?

Hm, I guess not. From a UX perspective it's just as usable.

website/src/components/discourse/styles.module.css
lfracalossiFA: Works for me, thanks!

website/src/components/discourse/index.js
JKarlavige: Here's an example of how it looks with the like count added. Not many have the like count so not sure if any would show up in the preview, so adding it here. I had to set the `after` prop to `2022-01-01` on the first component within the /forum page to get it to show up.

`<DiscourseFeed title='Open topics' category='help' status='unsolved'  link_text='See open topics' link_href='https://discourse.getdbt.com/c/help/19' show_cta={true} after='2022-01-01' />`

image

JKarlavige:
@joellabes All updates in place and ready for re-review. @lfracalossiFA i'll hold off on requesting a re-review from you until the meta fields are in place.

website/src/components/discourse/index.js
joellabes: Works for me!

website/src/pages/forum.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import Layout from '@theme/Layout';
 import Head from '@docusaurus/Head';
-import DiscourseFeed from '@site/src/components/discourse';
+import { DiscourseFeed } from '@site/src/components/discourse';
 
 function Events() {
```
joellabes, line 6: Is this a copy-paste hangover too? And does it matter?

joellabes (approved):
😻 😻 😻

website/src/pages/forum.js
```diff
@@ -0,0 +1,25 @@
...
+import Layout from '@theme/Layout';
+import Head from '@docusaurus/Head';
+import { DiscourseFeed } from '@site/src/components/discourse';
+
+function Events() {
+  return (
+    <Layout>
+      <Head>
+        <title>dbt Discourse</title>
+        <meta name="description" content="Description here" />
```
joellabes: @JKarlavige How about "Recent interesting discussions from across the dbt Forum" for now

website/src/components/discourse/index.js
```diff
@@ -0,0 +1,194 @@
...
+    topic_count={topic_count}
+    styles={styles}
+  />
+}
+
+// Helpers
+function TopicWrapper({ topic, children }) {
+  if(topic?.slug) {
+    return (
+      <a href={`https://discourse.getdbt.com/t/${topic.slug}`} title={topic.title} target="_blank">{children}</a>
```
joellabes: The slug is human-readable, but the actual important part of the link is the topic ID. 

https://discourse.getdbt.com/t/compare-two-models-schemas will redirect to https://discourse.getdbt.com/t/compare-two-models-schemas/3868

https://discourse.getdbt.com/t/fjhsbjhfbshjf/3868 will also get to the right place. 

I don't think this is _likely_ to break since we're using search results from a few seconds ago and the slug probably won't have been changed by a topic rename, but it's worth doing it the proper-er way.

joellabes (changes requested):

website/src/components/discourse/index.js
joellabes: (and _THEN_ I promise I'm done with feedback 😅)

website/src/pages/forum.js
JKarlavige: Since we added a second `DiscourseHelpFeed` component (which i added directly in the original component file), I had to change the way the components were exported. It was originally set to export 1 default component: `export default`, but since the file now has 2 components I changed to named exports `export const <variable-name>`. This allows the imports to be destructured with `import { DiscourseFeed, DiscourseHelpFeed }` as needed

website/src/components/discourse/index.js
JKarlavige: good call, just updated!

JKarlavige:
All updates in place and re-requested review. Documentation for using this component has been put together and is available here: https://www.notion.so/dbtlabs/Discourse-Component-Documentation-7057e62ee25b4536b84e676ca1c0e892

joellabes (approved):

JKarlavige:
- Tests added and pass successfully
- Updated to current
- Removed example from Jinja docs page
- Added /forum page to Learn dropdown in nav (i lied, we reverted the top nav PR until restructuring is complete.)
- This is ready for lift-off 🚀 

image

JKarlavige:
@joellabes note - this will be absent from the navigation until the new top nav is ready to be taken live.

PR #2031, opened 2022-09-19
DevHub top navigation updates

JKarlavige:
## Description & motivation

This updates the top navigation per the Figma mockup.

## To-do before merge
- [x] Sort out where "Courses" should point to
- [x] Sort out Community nav items
- [x] Sort out where Create a Free Account should point to

## Notes
We will move forward with taking the updated nav live without the community tab. When the community pages are built out, they will be added to the nav at that time. Same goes for the new /forum Discourse page.

## Preview
https://deploy-preview-2031--docs-getdbt-com.netlify.app/


lfracalossiFA (approved):
lgtm!

PR #2035, opened 2022-09-20
dbt-iomete: Updated documentation for iomete adapter

namigaliyev:
## Description & motivation

Added new connection parameter `account_number` and updated host format, based on new format of iomete lakehouse JDBC connection url.

Related to [release](https://github.com/iomete/dbt-iomete/releases/tag/v1.0.1)

### DIFF
--- a/website/docs/reference/warehouse-profiles/iomete-profile.md
+++ b/website/docs/reference/warehouse-profiles/iomete-profile.md
@@ -4,11 +4,11 @@ title: "iomete Profile"
 
 ## Overview of dbt-iomete
 
-**Maintained by:** iomete
-**Author:** Namig Aliyev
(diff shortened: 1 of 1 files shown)


dataders (approved):
thanks for keeping this up-to-date! It almost makes me think that there should be a `region_name` field instead of a `host` field? because the `.iomete.com` is probably the same for everyone?

namigaliyev:
> thanks for keeping this up-to-date! It almost makes me think that there should be a `region_name` field instead of a `host` field? because the `.iomete.com` is probably the same for everyone?

Good point, `.iomete.com` is the same for everyone and `region` would be more reasonable than `host` field.  A new pull request is on its way. 🙂

PR #2038, opened 2022-09-20
Update packages

JKarlavige:
Spinoff of #972, which was reverted due to the `package-lock.json` being updated, causing issues within existing PRs where the Netlify builds were failing.


nghi-ly (approved):
lgtm!

PR #2090, opened 2022-09-27
Update test-selection-examples.md

savannahcarnahan:
Updated syntax from legacy test_type name (data) to current nomenclature (singular)

## Description & motivation

I was confused by the test_type:data designation under Test Selection: Examples. I was able to find a pull request documenting the legacy test_type tags [here](https://github.com/dbt-labs/dbt-core/pull/3880). I used this to update the example for the current version.


runleonarun (approved):
Great call out, and thank you for contributing to our docs @savannahcarnahan! I agree that we can leave data as it will be backwards compatible.

PR #2101, opened 2022-09-29
One FAQ entry not showing on the Sources page

b-per:
## Description & motivation

The name of the FAQ file misses an `s`.
Would there be a way to do those checks automatically?


b-per:
Fixed in the Preview:
image

mirnawong1 (approved):

mirnawong1:
thank you for that flag @b-per and also asking @JKarlavige whether it is possible to check automatically:

> Would there be a way to do those checks automatically?

JKarlavige:
@mirnawong1 At the moment there is not a way to check these automatically. But testing these components is on the agenda for next quarter!

PR #2111, opened 2022-10-01
Fix typo in RedShift SQL query in tutorial-initiate-project.md

Odessit007:
## Description & motivation
The snippet suggests this SQL query for RedShift:
```
select * from jaffle_shop_customers
```
which is wrong since there must be a dot instead of underscore before `customers`.


nghi-ly:
Totally agree with you, @Odessit007 ! 

In reviewing your PR, I discovered this was fixed in PR #2475. Will go ahead and close this.

Thanks so much for contributing to the dbt docs!

PR #2119, opened 2022-10-03
Changing Version for Metrics Package

callum-mcdata:
## Description & motivation
We decided to go with a different version number for the metrics package to better reflect its 1:1 tie with the dbt-core version.

### DIFF
--- a/website/docs/docs/building-a-dbt-project/metrics.md
+++ b/website/docs/docs/building-a-dbt-project/metrics.md
@@ -318,7 +318,7 @@ Use the following [metrics package](https://hub.getdbt.com/dbt-labs/metrics/late
 ```yml
 packages:
   - package: dbt-labs/metrics
-    version: [">=0.4.0", "<0.5.0"]
+    version: [">=1.3.0", "<1.4.0"]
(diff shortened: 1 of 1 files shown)


mirnawong1 (approved):
great stuff, tested it and good to go! @callum-mcdata can I just clarify two questions below:

1. why is the range not like the one below?

  - package: dbt-labs/metrics
    version: [">=1.3.0", "<2.0.0"]

The[ hub info](https://hub.getdbt.com/dbt-labs/metrics/latest/) has the dbt version range of 1.30 - 2.0.0 and the actual package only has 1.3.0 - should that be changed too?

2. should the new version number be reflected in the semantic layer docs?

callum-mcdata:
@mirnawong1 this is part of our tighter coupling of dbt metrics and dbt core. We'll be slowing down to their minor version release cycle so that each version of the metrics package corresponds to the core release. As such, we want to make sure that this version of the package is only used with 1.3 in case we need to add breaking changes to 1.4.

I released an older version that had that range and it looks like it got picked up by the hub. Let me dig into that

PR #2124, opened 2022-10-03
Rework long discussions

joellabes:
## Description & motivation
Reflect change from GHD to Discourse for forum

- [x] The page has been removed from `website/sidebars.js`
- [x] An entry has been added to `_redirects`
- [x] [Run link testing](https://github.com/dbt-labs/docs.getdbt.com#running-the-cypress-tests-locally) to update the links that point to the deleted page ⚠️ I did this manually instead of using the Cypress test


joellabes:
@gwenwindflower @jasnonaz i think it makes sense for me to merge this into your branch and then go into `next` together instead of trying to reconcile two overlapping change sets, wdyt?

gwenwindflower:
yep @joellabes let's do that!

jasnonaz (approved):
let there be forum 🥸

PR #2125, opened 2022-10-04
Add and render post blurbs

joellabes:
## Description & motivation
It would be nice to include the blurb of the post where available. This should do that!

### DIFF
--- a/website/src/components/discourse/index.js
+++ b/website/src/components/discourse/index.js
@@ -138,7 +138,7 @@ export const DiscourseFeed = ({
                 <>
                   {' '}
                   <p>
-                    {topic?.blurb}
+                    {topic.blurb}
                   </p>
                 </>
               )}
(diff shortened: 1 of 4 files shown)


joellabes:
@JKarlavige just a quick one - I've added the blurbs in as a separate paragraph.

website/src/components/discourse/index.js
```diff
@@ -134,6 +134,14 @@ export const DiscourseFeed = ({
                   </span>
                 </>
               )}
+              {(topic?.blurb) && (
+                <>
+                  {' '}
+                  <p>
+                    {topic.blurb}
+                  </p>
```
JKarlavige: Paragraphs have a default margin-bottom throughout the site, so this creates an uneven spacing for each topic row.

I don't think it looks too bad as is, but if you wanted to adjust this, you can add the following css to `website/src/components/discourse/styles.module.css`:
```
.discourseTopics p {
  margin-bottom: 0;
}
```

image

JKarlavige (approved):
Nice work building out the code for this, looks great! Added one suggestion as far as margin goes on the new paragraph tag, but this looks solid. Approved! :shipit:

PR #2132, opened 2022-10-04
More metric updates

callum-mcdata:
## Description & motivation
Fixing the `window` property for metrics


mirnawong1 (approved):
approved and good to go @callum-mcdata !

PR #2147, opened 2022-10-06
add note about cross-db macros in dbt-sqlserver

sdebruyn:
## Description & motivation
Add note about unsupported cross-db macros in dbt-sqlserver


dataders (approved):
:shipit:

PR #2155, opened 2022-10-06
Update supported-data-platforms.md

Damuel4000:
Grammar correction on Supported Data Platforms introductory paragraph.

## Description & motivation

This change removes a word from a sentence which caused a grammatical error. Though the meaning was clear without this change, this change creates easier understanding on the part of the reader.


nghi-ly:
@DamuelOnline , I totally agree with you! 

In reviewing your PR, I found that this was addressed in PR #2271 . I'll go ahead and close this.

Thanks so much for contributing to the dbt docs!

PR #2166, opened 2022-10-07
Don't include the most recent data when doing an incremental update

AndrewLane:
## Description & motivation

Trying to get the logic of an incremental model just right


mirnawong1:
thanks for opening this up @AndrewLane - great suggestion! generally speaking (not relative to the python code specifically), it probably depends on the user’s data. If there are no duplicates, like each record is its own timestamp, then > is probably fine. If you are expecting duplicates like it is a date field and there are multiple records for each date, then you would use >=  and hopefully a unique key not to bring in duplicates.

PR #2167, opened 2022-10-07
Fix outdated links to Azure git docs

gsass:
## Description & motivation

Some links were referencing the old (`current` branch) locations of some Azure DevOps docs. These changes fix that.


ericbaumann:
This seems fine. I clicked around the PR deploy and was able to reach the links in question

I did get a few 404s on other links that appeared in the search menu (these followed a similar path to the one you're updating here), is this a problem?  

example https://deploy-preview-2167--docs-getdbt-com.netlify.app/docs/dbt-cloud/cloud-configuring-dbt-cloud/cloud-import-a-project-by-git-url#step-2-import-your-repository-by-name

gsass:
@ericbaumann yuck - yeah that's the same copypasta error. Let me do a global search and see where else we gotta fix this.

ericbaumann:
> @ericbaumann yuck - yeah that's the same copypasta error. Let me do a global search and see where else we gotta fix this.

One thing I'd be wary of in this "global find & replace" is other systems having references to these outdated paths - `cloud-ui` for instance, also seems contain references to docs with these URLs. 

I don't:
1. truly understand the context behind why this path is no longer valid
2. feel that you should automatically assume responsibility for tracking all of these dead links down on this PR

If you do end up deciding to take on the work of updating additional references, i want to make sure we have eyes from someone with more context into how this docs system works because that person is not currently me

gsass:
Agreed @ericbaumann - I really want us to have some link checking as part of our tooling. It's [available in docusaurus](https://docusaurus.io/docs/api/docusaurus-config#onBrokenLinks) 2.0.1 and 2.1.0 via a simple config change. I think we should upgrade Docusaurus since our current version isn't maintained anymore, which will also let us error on broken links.

ericbaumann (approved):

MitchellBarker (approved):
Seems perfect!

runleonarun (approved):
🚢 Thanks for fixing this!

PR #2171, opened 2022-10-08
add quotes to `project_name` in example macro

krishbox:
## Description & motivation

A small change to add quotes to the example macro to make it syntactically correct.


mirnawong1:
hey @krishbox thanks for opening this pr up! The current snippet is actually correct. `project_name` is a Jinja variable so we don’t quote it as if it was a string - does that make sense?

image

mirnawong1:
hey @krishbox thanks again for opening this up and I'll go ahead and close this pr. appreciate your contribution!

PR #2175, opened 2022-10-10
navbar titles now have a hover and fixed the padding issue

zesuma:
## Description & motivation

preview: https://deploy-preview-2175--docs-getdbt-com.netlify.app/

I've combined 3 tickets into this one:

**Nesting and interactive states of folders**
https://app.asana.com/0/1200099998847559/1203010186078425/f

**Font size is different in IDE GA launch user documentation**
https://app.asana.com/0/1200099998847559/1203134802011321/f

**Fonts in toggle snippet are different sizes**
https://app.asana.com/0/1200099998847559/1203118391233000/f

**Changes made:**

1. Sidebar nesting has been fixed. When you hover over the main title of the dropdown menu, the background color is activated.
2. The indentation has standardized indentation. 
3. Font sizes inside toggle snipets and tables is now 18px

PR #2180, opened 2022-10-10
Change Caret to right side

thorn14:
## Description & motivation
With the current side navigation being left aligned for text, it would be easier to reveal the nesting structure of collapsible sections by moving the caret to the right hand side like below.
Screen Shot 2022-10-10 at 12 59 51 PM

This is a change from the current implementation where nested folders can appear to be belonging to other docs as below:
Screen Shot 2022-10-10 at 1 00 23 PM

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -369,29 +369,28 @@ a.navbar__item.navbar__link.btn:hover {
 
 .menu__caret {
   padding: 0;
+  margin-left: 0;
 }
 
 .menu__caret:hover {
   background: transparent;
 }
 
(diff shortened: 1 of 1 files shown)


joellabes:
Is this PR any relation to https://app.asana.com/0/1200099998847559/1203075014761010/f ? Or are they happening independently?

thorn14:
Separate PR to address half of this ticket
https://app.asana.com/0/1200099998847559/1203010186078425/f from this pr
https://github.com/dbt-labs/docs.getdbt.com/pull/2175

PR #2189, opened 2022-10-11
Removing vestigial content and adding example

callum-mcdata:
## Description & motivation
Removing a holdover from old metric documentation that was confusing to some users AND adding a new example for the `expression` field to show users they can use non-aggregating sql functions


mirnawong1 (approved):
LGTM @callum-mcdata!

PR #2208, opened 2022-10-13
Changing the date column name

callum-mcdata:
## Description & motivation
Great find by @joellabes that the date field doesn't 100% make sense. Changing to `event_date` for more clarity.


website/blog/2022-10-12-how-to-design-and-structure-metrics.md
```diff
@@ -54,7 +54,7 @@ In this example, we’ll cover the basics of defining a metric and a fairly stra
 
 In the past, the `marts` tables were often your end stage layer before data was consumed in another tool or system. Now, the mart is the springboard for the creation of our metric. So we'll begin by looking our end-state `marts` model called `order_events` that looks something like the below table, but on the order of millions of rows instead of five. Our finance team uses the below model to better understand revenue but inconsistencies in how its reported have led to requests that the data team centralize the definition in the dbt repo.
```
joellabes:
```suggestion
In the past, the `marts` tables were often your end stage layer before data was consumed in another tool or system. Now, the mart is the springboard for the creation of our metric. So we'll begin by looking our end-state `marts` model called `order_events` that looks something like the below table, but on the order of millions of rows instead of five. Our finance team uses the below model to better understand revenue but inconsistencies in how it's reported have led to requests that the data team centralize the definition in the dbt repo.
```

joellabes (dismissed):
Found a couple of other typos, which I couldn't comment on because they weren't close enough to the diff: 

practicioners -> practitioners on lines 25 and 27
organizations -> organization's on line 76

>In practice, the all-in-one YML method would look like the following:

there's no image or code block here

joellabes (approved):
🌟

PR #2210, opened 2022-10-13
tweak copy to maintain equal lengh of cards

jasnonaz:
Unequal copy lengths were causing cards to feel off balance


gwenwindflower (approved):
:shipit:

PR #2212, opened 2022-10-13
Fixing spacing for the metrics blog

callum-mcdata:
## Description & motivation
Notion convert to Markdown does not get along with spaces vs tabs. Fixing all that - amazing eyes by @b-per !!


dbeatty10 (approved):
I didn't check that the YAML actually works, but _visually_ it looks like what it is meant to correct.

PR #2218, opened 2022-10-13
Feature/discourse comments

JKarlavige:
This re-opens @breezyfasano's original PR #1069 

## Preview

https://deploy-preview-2218--docs-getdbt-com.netlify.app/blog/how-to-build-a-mature-dbt-project-from-scratch

Original description:

## Description & motivation
Adding comments to the dev blog via Discourse embeds

How to Review:
- Preview link: [click on any blog post!](https://deploy-preview-1069--docs-getdbt-com.netlify.app/blog/)
- Below each post content, there should be an embed that says 'start discussion'! That's the discourse embed. Let me know if you all can see it 

A note about Discourse's embed functionality from the Discourse docs:

> One important thing to note with this setup is that users have to navigate to your forum to post replies. This is intentional, as we feel that the posting interface on a Discourse forum is currently much richer than what we could embed via Javascript.

Edit 4:08pm CT:
Here are the comments in action! this does create a discourse topic, but I just commented to test the system works as planned and then deleted the topic it made. Screenshots below.

## On the Website
Screen Shot 2022-02-01 at 4 07 28 PM

## On Discourse Topic
Screen Shot 2022-02-01 at 4 07 21 PM


joellabes:
@JKarlavige looks like the iframe is too smol somtimes https://www.loom.com/share/bf2c2d7dd0274c2f98feb58e7f5a8124

JKarlavige:
> @JKarlavige looks like the iframe is too smol somtimes https://www.loom.com/share/bf2c2d7dd0274c2f98feb58e7f5a8124

Yeah seeing that on my end as well, looks to only occur in Chrome. TBH, i'm not crazy about this iframe, there's a couple of other things i'm noticing.
- It won't adapt to dark mode, so we'll have a white box at the bottom.
- In Firefox, there's no padding, so it looks pretty bad in dark mode.
- The "1 Reply" button doesn't seem to work for me.

The script is generating the height of the iframe, so i'm not sure how we can work around that. We might be able to set a predefined height, but would have to setup vertical scrolling.

What are your thoughts on holding off on this, and building it out with the API in the same way we did with the DiscourseFeed component? Obviously that would push us back to after Coalesce though.

image

jasnonaz:
I think that's the right call - thanks @JKarlavige. Twas worth a shot

JKarlavige:
> I think that's the right call - thanks @JKarlavige. Twas worth a shot

Yeah definitely worth trying it out. Posted an update in the Asana task for us to circle back to this and build it out in a similar way to the DiscourseFeed component.

Will keep this PR open and we can add our additional work directly into here.

JKarlavige:
Closing without merge, this was completed in PR #2577.

PR #2236, opened 2022-10-14
Fix links from npm output

nghi-ly:
## Description & motivation

Fix links in /reference files from npm run build output


runleonarun (approved):
LGTM

PR #2243, opened 2022-10-14
LOG_FORMAT env var should be DBT_LOG_FORMAT

troyharvey:
image

## Description & motivation

After fumbling around with the `LOG_FORMAT` environment variable and failing to get structured json logging enabled I discovered in the GitHub commits that the variable is actually `DBT_LOG_FORMAT=json`


website/docs/reference/global-configs.md
```diff
@@ -214,7 +214,7 @@ FailFast Error in model model_1 (models/model_1.sql)
 
 ### Log Formatting
 
-The `LOG_FORMAT` config specifies how dbt's logs should be formatted. If the value of this config is `json`, dbt will output fully structured logs in <Term id="json" /> format; otherwise, it will output text-formatted logs that are sparser for the CLI and more detailed in `logs/dbt.log`.
+The `DBT_LOG_FORMAT` environment variable specifies how dbt's logs should be formatted. If the value of this config is `json`, dbt will output fully structured logs in <Term id="json" /> format; otherwise, it will output text-formatted logs that are sparser for the CLI and more detailed in `logs/dbt.log`.
```
runleonarun:
```suggestion
The `LOG_FORMAT` config specifies how dbt's logs should be formatted. If the value of this config is `json`, dbt will output fully structured logs in <Term id="json" /> format; otherwise, it will output text-formatted logs that are sparser for the CLI and more detailed in `logs/dbt.log`.
```

runleonarun: 👋🏻 @troyharvey This is actually correct as it is. You can read about [environment variables here](https://docs.getdbt.com/reference/global-configs#environment-variables) where we say to add DBT_ to the config `LOG_FORMAT` to construct them properly. 

That being said, it was obviously not clear, so we will rework this section of the docs to clarify that piece!

runleonarun: @mirnawong1 I wonder if this could be part of the next phase of the reference project?

runleonarun (approved):
@troyharvey thank you improving the dbt product documentation! We will make the nuances of the `DBT_` clearer in a future project.

PR #2245, opened 2022-10-15
dbtignore and migration note

lostmygithubaccount:
## Description & motivation

There is a significant change in v1.3 that could be considered a breaking change -- non-model `.py` files are now parsed and likely to cause errors if they exist in model directories.

This was first reported in the community Slack, and since has been seen a few times. To address it, we added a `.dbtignore` file in v1.3 that can be used to ignore non-model `.py` files (or anything) from being parsed by dbt.

## To-do before merge

I could not figure out how to make the actual filename/sidebar show as ".dbtignore" -- if that's possible, we should do that.

Any cleanup desired by docs team.

Do we call this a breaking change? I'd say it technically is one.

## Prerelease docs
N/A

## Checklist

### DIFF
--- a/website/docs/guides/migration/versions/05-upgrading-to-v1.3.md
+++ b/website/docs/guides/migration/versions/05-upgrading-to-v1.3.md
@@ -9,7 +9,11 @@ title: "Upgrading to v1.3 (latest)"
 
 ## Breaking changes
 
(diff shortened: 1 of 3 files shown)


jtcohen6:
@lostmygithubaccount Looks good! I'm going to borrow this PR (if you don't mind) to also add a "maybe breaking change" note about the `profiles.yml` order, prompted by https://github.com/dbt-labs/dbt-core/issues/6066.

jtcohen6 (approved):

PR #2299, opened 2022-10-24
shorten survey CTA copy

glin7:
## Description & motivation
Trim the fat in survey banner CTA (in notification bar).


runleonarun (approved):
@glin7 Good idea to shorten!

PR #2337, opened 2022-10-31
Clarify that discourse topics are user generated

JKarlavige:
This adjusts the Discourse Feed component styles to better indicate this section is generated from users.

## Preview

https://deploy-preview-2337--docs-getdbt-com.netlify.app/docs/build/jinja-macros


jasnonaz (approved):
This looks awesome!! Thanks for punching this up

PR #2338, opened 2022-10-31
Additional page contexts for blog

JKarlavige:
This adds IDs to blog elements to get their values within GTM. In GTM, we will be passing the following additional blog context into Snowplow:

- author
- category
-  publish_date

## Asana task

https://app.asana.com/0/1200099998847559/1202307891226757/f

## Preview

https://deploy-preview-2338--docs-getdbt-com.netlify.app/blog/sql-surrogate-keys


website/src/theme/BlogPostItem/index.js
```diff
@@ -101,6 +101,38 @@ function BlogPostItem(props) {
...
+      blogDate: (formattedDate && isBlogPostPage) ? formattedDate : undefined
+    }
+
+    if(authors && authors.length > 0 && isBlogPostPage) {
+      authors.map((author, i) => {
+        blogContext.blogAuthor += 
+          `${author.name}${i !== authors.length - 1 ? ', ' : ''}`
+      })
+    } else {
+      blogContext.blogAuthor = undefined
```
lfracalossiFA: I'm assuming it's okay that these values are going into GTM as "undefined?" I'm not sure how GTM treats those values.

lfracalossiFA (dismissed):

website/src/theme/BlogPostItem/index.js
JKarlavige: Yes that sets the variables to `undefined` in GTM, which is intended. This component lives on other pages such as the blog index page. If on any page other than a blog post page, we want to make sure we update the data layer variables to undefined, otherwise GTM will use the value which was last pushed.

For example, if someone views a blog post page, the author, category and date is sent to GTM as variables. If we didn't update these variables to `undefined` when clicking a different page, when the visitor clicks a different page it would still pass in the original values from the blog post page into Snowplow.

This ensures we're updating the variables on every page load, that way those variables will only have values if viewing a blog post page, otherwise it's passed in as `undefined` for all other pages. This is also why the Root component was updated to push `undefined` when viewing any other pages where the `BlogPostItem` component is not used.

lfracalossiFA: Good to know! Does "undefined" just go into Snowplow as a string? I assume the data folks know what to do with this.

JKarlavige: Yeah it would go through like the screenshot below. I followed up in the asana task to verify if that'll work or if we should make any adjustments to improve it on the data side.
image

website/src/theme/BlogPostItem/index.js
```diff
@@ -101,6 +101,36 @@ function BlogPostItem(props) {
     );
   };
 
+  // dbt custom - send blog context to datalayer to send to snowplow
+  useEffect(() => {
+    let blogContext = {
+      event: 'blogContext',
+      blogAuthor: '',
+      blogCategory: '',
+      blogDate: formattedDate ? formattedDate : undefined
```
lfracalossiFA, line 110: Where are we using this?

lfracalossiFA (commented):
Quick question for you

website/src/theme/BlogPostItem/index.js
lfracalossiFA: By "this" I mean `blogDate`.

JKarlavige: These are Data Layer variables in GTM. Those variables are then being passed into snowplow

lfracalossiFA: Okay, I didn't see where blogDate was being used in this file, so I wasn't sure if it was being used somewhere else.

Thanks! Are you just waiting on my approval?

JKarlavige: I believe so! Looks like both Krevitt and Brandon approved in Asana 👍

lfracalossiFA (approved):
lgtm!

PR #2358, opened 2022-11-03
Mention updated_at config for check strategy

mniehoff:
Updated_at config can be used to fill the timestamps when using the check strategy for creating snapshots. See dbt-core issue 3376.

## Description & motivation
Just found this handy feature by googling, not be reading the docs. Should be in the docs imho.

## To-do before merge
Nothing


website/docs/docs/build/snapshots.md
```diff
@@ -373,7 +373,7 @@ Snapshot results (note that `11:30` is not used anywhere):
 
 <br/>
 
-For the `check` strategy, the current timestamp is used to populate each column
+For the `check` strategy, the current timestamp is used to populate each column. If configured, the `updated_at` column is used instead, as with the timestamp strategy.
```
mirnawong1:
```suggestion
For the `check` strategy, the current timestamp is used to populate each column. If configured, the `check` strategy uses the `updated_at` column instead, as with the timestamp strategy.
```

mirnawong1 (approved):
hey @mniehoff apologies for the delay here! thanks for opening this up and making it easier for users. I've gone ahead and approved + merged this for you, really appreciate it! ✨

PR #2374, opened 2022-11-07
Custom search weight component

john-rock:
[Asana Task](https://app.asana.com/0/1200099998847559/1202910189491311/f)

# Current Problem
We have no way to currently set a doc page to rank higher than another page via Algolia search. This is a feature of the paid tier plan within Algolia but since we are on the free plan we currently do not have access to this through the Algolia dashboard. 

# Proposed Solution
This PR adds a custom `<DocSearchWeight />` component that is added to every doc page. This component accepts one prop (`weight`) that accepts a value from a doc items frontmatter called `search_weight` (if no value is set in the frontmatter this defaults to 0).

Since we have _do_ have the ability to set custom ranking properties within the Algolia dashboard theoretically adding this attribute to the page and then setting in the Algolia crawler config to read this value on the page, we should be able to influence which pages are ranked higher within the search results.

❗️Please view the [Notion Doc](https://www.notion.so/dbtlabs/WIP-Custom-Search-Weight-Component-58aaa3aca7ff42938887e6159f63e7da) for more information on usage. ❗️

## Algolia Configuration
This configuration will only have to be completed once after merging if approved. We need to adjust a few things in both the Algolia Crawler and Algolia Dashboard

### Algolia Crawler
Inside the config editor add the following items:
Editor___Crawler_Admin_Console
This will allow the crawler to read the value of our custom `searchWeight`

Editor___Crawler_Admin_Console
This will allow us to use the value as a ranking value within the Algolia Dashboard.

### Algolia Dashboard
Under Search > Configuration > Ranking & Sorting
Add the customSearchWeight attribute as  a custom ranking property and set it to descending.
Index___Algolia


JKarlavige:
This looks like a great solution for handling edge cases where we want to bump one search result ahead of the other. From your Loom video, it looks like this will work as expected, really creative way to solve this!

Like you mentioned it's not easy to test as we don't have a sandbox crawler. But for verifying this works in production, i can't see any issues if we merged the change and updated the crawler config. You'll just need to fire off a new crawl after adjusting the config.

To confirm the expected behavior if we were to merge this as-is to see how it works in production, the `Build your DAG > Sources` page should begin appearing ahead of the rest of the results when searching "Sources" in the search. Is that correct?

john-rock:
> To confirm the expected behavior if we were to merge this as-is to see how it works in production, the `Build your DAG > Sources` page should begin appearing ahead of the rest of the results when searching "Sources" in the search. Is that correct?

That is correct. 

I don't anticipate any negative side effects from making this change but in the event there are, we revert this PR and revert the settings within both the Algolia Crawler and Dashboard.

lfracalossiFA:
@john-rock From reading the Notion doc, I gather there is no upper bound on `searchWeight`? Kind of like z-index?

website/src/theme/DocItem/index.js
```diff
@@ -49,6 +50,10 @@ export default function DocItem(props) {
   // If term has cta property set, show that cta
   const termCTA = frontMatter?.cta && frontMatter.cta
 
+  // dbt Custom
+  // If the page has a searchWeight value, apply that value
+  const searchWeight = frontMatter?.searchWeight && frontMatter.searchWeight
```
lfracalossiFA: Okay, dumb question. What's the difference between these two statements on the right?

lfracalossiFA (dismissed):

JKarlavige (dismissed):
:shipit:

website/src/theme/DocItem/index.js
john-rock: Not a dumb question. For this I followed the pattern we were using for `termCTA` and in this case I believe the && is being used as a short circuit operator between the two statements (e.g. `const searchWeight = A && B` - B is only evaluated if A is truthy).

Looking at it now since we are using optional chaining with `frontMatter?.searchWeight`, I don't think it's necessary to do a short circuit eval. @JKarlavige thoughts on just using `const searchWeight = frontMatter?.searchWeight` here?

john-rock:
> @john-rock From reading the Notion doc, I gather there is no upper bound on `searchWeight`? Kind of like z-index?

Correct - Right now there is no upper bound.

website/src/theme/DocItem/index.js
JKarlavige: Yeah that should work fine, I believe the shorter has the same outcome as the longer statement. If `frontMatter` or `frontMatter.searchWidget` is not set, `const searchWeight` will = `undefined`.

lfracalossiFA: Okay -- so I think I understand. Let me read up a little bit more on what `?` does, because I've seen you folks using it a lot and I assumed it was a TS thing 😆 .

JKarlavige: @lfracalossiFA The way i look at it is it being a shorthanded `if` statement for checking nested properties in objects. For example, the statement is shorthand for:
```
if(frontMatter && frontMatter.searchWidget)
```

Doing `frontMatter?.searchWidget` says if `frontMatter` is not set, return false, otherwise proceed to check if `searchWidget` is set.

nghi-ly (commented):
this looks great! 

a couple of questions: 

- should the YAML key be `search-weight` (with a hyphen) instead for style consistency? i use `sidebar-label` quite a bit. for example: https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/docs/get-started/getting-started/getting-set-up/setting-up-bigquery.md  
- per the Notion doc, the value is between 0 - 100. is it possible to have "no", "low", "medium", "high" instead like presets?

john-rock:
Hey @nghi-ly 

>     * should the YAML key be `search-weight` (with a hyphen) instead for style consistency? i use `sidebar-label` quite a bit. for example: https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/docs/get-started/getting-started/getting-set-up/setting-up-bigquery.md
Absolutely - I can switch it to use a hyphen format instead.

>     * per the Notion doc, the value is between 0 - 100. is it possible to have "no", "low", "medium", "high" instead like presets?
Let me look into this and get back to you. The official ranking factor settings within Algolia allow for a numerical value which can then be sorted asc or desc as well as having a boolean value where we can use `is-featured: true` which will then boost all doc pages that have true. I may be able to intercept the `low`, `medium`, and `high` values before they get placed on the page and assign them a numerical value behind the scenes that relates to their values. Let me experiment and get back to you. Thanks for checking this out!

nghi-ly:
> Let me look into this and get back to you. The official ranking factor settings within Algolia allow for a numerical value which can then be sorted asc or desc as well as having a boolean value where we can use is-featured: true which will then boost all doc pages that have true. I may be able to intercept the low, medium, and high values before they get placed on the page and assign them a numerical value behind the scenes that relates to their values. Let me experiment and get back to you. Thanks for checking this out!

@john-rock : goodness, tYOUsm for putting your eyes on this!! maybe "light" , "medium" , "heavy" makes more sense when talking about "weight". just suggs, totally your call.

john-rock:
Hey @nghi-ly, I was able to update the component with both of your request. The new accepted values for search_weight are `light`, `medium`, and `heavy`. These values are intercepted before page build and have a numerical value applied behind the scenes. I also went ahead and updated this information in the Notion doc. 

Lmk what you think and if we should move forward with this experiment.

JKarlavige (approved):
Verified component getting `search_weight` frontmatter property and converting it to the correct numerical value. lgtm! 

image

nghi-ly (approved):
incredible! ty!! 🚢

PR #2386, opened 2022-11-13
Clarify supported behavior in selector inheritance

jtcohen6:
## Description & motivation
The same misconception has now come up in two `dbt-core` issues:
- https://github.com/dbt-labs/dbt-core/issues/5709
- https://github.com/dbt-labs/dbt-core/issues/6212

Basically, while `selector` inheritance is useful & usable for some things, it still lacks some capabilities that users may otherwise expect / perceive. Let's call out that limitation explicitly.

I'm not in love with the language I'm using here. Just taking a first pass at making the docs change we need to see in the world.


website/docs/reference/node-selection/yaml-selectors.md
```diff
@@ -288,4 +288,7 @@ selectors:
               value: buzz
 ```
 
+**Note:** While selector inheritance allows the logic from another selector to be _reused_, it does not allow the logic from that selector to be _modified_ by means of `parents`, `children`, `indirect_selection`, etc. The `selector` method simply returns the complete set of nodes returned by the named selector.
```
mirnawong1:
```suggestion
**Note:** While selector inheritance allows the logic from another selector to be _reused_, it doesn't allow the logic from that selector to be _modified_ by means of `parents`, `children`, `indirect_selection`, and so on. 

The `selector` method returns the complete set of nodes returned by the named selector.
```

mirnawong1 (dismissed):
thanks for clarification and opening up the PR @jtcohen6 made some tiny tweaks and I've approved it for you. Let me know if you'd like me to merge it as well!

mirnawong1:
hey @jtcohen6 wanted to flag this - are you happy with the suggestions?

mirnawong1 (approved):

PR #2389, opened 2022-11-14
Wistia videos overlapping H1 in docs

JKarlavige:
## Description & motivation

Wistia videos are overlapping the docs h1 in the airflow set up guide. It looks like these specific videos have an aspect ratio which isn't 16:9, causing the default padding to not push the rest of the content down far enough on this page.

It looks like there is an existing optional `paddingTweak` prop setup on the `WistiaVideo` component, likely to fix videos that experience this specific issue. I added the `paddingTweak` prop for videos on this page which have the issue. 

## Preview
https://deploy-preview-2389--docs-getdbt-com.netlify.app/guides/orchestration/airflow-and-dbt-cloud/2-setting-up-airflow-and-dbt-cloud

Example page where issue is not occurring.
https://deploy-preview-2389--docs-getdbt-com.netlify.app/docs/get-started/learning-more/refactoring-legacy-sql


mirnawong1 (approved):
thank you for tagging me @JKarlavige and it looks good to me!

PR #2391, opened 2022-11-14
Update column_types.md

ernestoongaro:
## Description & motivation
Spent some time to figure this out and thought it might help others!


website/docs/reference/resource-configs/column_types.md
```diff
@@ -79,3 +79,6 @@ seeds:
 
 ## Recommendation
 Use this configuration only when required, i.e. when the type inference is not working as expected. Otherwise you can omit this configuration.
+
+## Troubleshooting
+Note that column type configuration is case sensitive, so a column called Country_Name in your Seed, should be referenced by Country_Name, not country_name. This is irregardless of quoting configuration.
```
mirnawong1:
```suggestion
Note: The `column_types` configuration is case-sensitive, regardless of quoting configuration. If you specify a column as `Country_Name` in your Seed, you should reference it as `Country_Name`, and not `country_name`.  
```

mirnawong1:
hey @ernestoongaro thanks so much for looking out for others and opening this PR up! I've made some tweaks for you to have a look. i also think your new content would be better placed on the top as under "description" rather than "troubleshooting" so the user can learn this from the get go - what do you think? if so, i can make the edits for you!

ernestoongaro:
> hey @ernestoongaro thanks so much for looking out for others and opening this PR up! I've made some tweaks for you to have a look. i also think your new content would be better placed on the top as under "description" rather than "troubleshooting" so the user can learn this from the get go - what do you think? if so, i can make the edits for you!

Thanks @mirnawong1 ! I think this is a rare thing, so I don't want to take up mental real estate at the top, so happy to keep it here (or if there's another place where seeds are documented with a larger troubleshooting section)

mirnawong1:
no worries @ernestoongaro and let me know if you're ok with the content suggestions

UPDATE - oops my bad, i saw you committed them already! thank you! I'll merge it for you once the test have passed!

mirnawong1 (approved):

PR #2413, opened 2022-11-17
Fixing Label to be required

callum-mcdata:
ignore my misspelled branch name plz

## Description & motivation
Label is a required field for metrics. If we decide to change this in the future we can do that within core but for now we need to make sure the docs reflect reality.


mirnawong1 (approved):
lgtm and thanks for opening this up @callum-mcdata ! the branch spelling has been ignored 😅

PR #2415, opened 2022-11-17
Remove advice to avoid multi outputs for int models

b-per:
The advice of avoiding multiple outputs from a single intermediate has been debated internally and seems to be confusing to some community users as well.

For now, we can remove this section to avoid confusion and potentially add additional items of what int models should NOT be in the future.

See [this slack](https://dbt-labs.slack.com/archives/C017GDLAF7D/p1668611429698739?thread_ts=1668604131.310129&cid=C017GDLAF7D) as well


mirnawong1:
thanks for opening this up @b-per and sorry for the delay here as we catch up. I'll go ahead and approve this pr for you to merge! ✨

mirnawong1 (approved):

PR #2418, opened 2022-11-18
Example of custom macros in get_where_subquery

jtcohen6:
## Description & motivation

We get this question _a lot_:
- https://github.com/dbt-labs/dbt-core/issues/3580#issuecomment-929125142
- https://github.com/dbt-labs/dbt-core/issues/4103#issuecomment-948567484
- https://github.com/dbt-labs/dbt-core/issues/5259#issuecomment-1129310833
- https://github.com/dbt-labs/dbt-core/issues/5940
- https://github.com/dbt-labs/dbt-core/discussions/6236

Folks want to be able to stick custom macros in the `where` config for tests. While that isn't directly possible today — and that limitation is worth documenting explicitly — there's a halfway-decent workaround that's worth documenting.


website/docs/reference/resource-configs/where.md
```diff
@@ -123,6 +123,8 @@ tests:
 
 ### Custom logic
 
+The rendering context for the `where` config is the same as for all configurations defined in `.yml` files. You have access to `{{ var() }}` and `{{ env_var() }}`, but you **do not** have access to custom macros for setting this config. If you do want to use access custom macros to template out the `where` filter for certain tests, there is a workaround.
```
dbeatty10:
```suggestion
The rendering context for the `where` config is the same as for all configurations defined in `.yml` files. You have access to `{{ var() }}` and `{{ env_var() }}`, but you **do not** have access to custom macros for setting this config. If you do want to use custom macros to template out the `where` filter for certain tests, there is a workaround.
```

website/docs/reference/resource-configs/where.md
```diff
@@ -132,3 +134,46 @@ dbt replaces `{{ model }}` in generic test definitions with `{{ get_where_subque
 You can override this behavior by:
 - Defining a custom `get_where_subquery` in your root project
 - Defining a custom `<adapter>__get_where_subquery` [dispatch candidate](dispatch) in your package or adapter plugin
+
+Within this macro definition, you can reference whatever custom macros you want, based on static inputs from the configuration. At simplest, this enables you to DRY up code that you'd otherwise need to repeat across many different `.yml` files. Because the `get_where_subquery` macro is resolved at runtime, your custom macros can also include [fetching the results of introspective database queries](https://docs.getdbt.com/reference/dbt-jinja-functions/run_query).
+
+**Example:** Filter your test to the past three days of data, using dbt's cross-platform `{{ dateadd() }}` utility macro.
```
dbeatty10: We've been trying to provide an explicit `dbt` namespace when showing code examples for cross-platform macros.

Two different ideas for slight updates.

Idea 1:

Just link to the docs for this macro rather than Jinja-fying it:
```suggestion
**Example:** Filter your test to the past three days of data, using dbt's cross-platform [`dateadd()`](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros#dateadd) utility macro.
```

Idea 2:
Leave the Jinja curlies, but add an explicit `dbt` namespace:
```suggestion
**Example:** Filter your test to the past three days of data, using dbt's cross-platform `{{ dbt.dateadd() }}` utility macro.
```

website/docs/reference/resource-configs/where.md
```diff
@@ -132,3 +134,46 @@ dbt replaces `{{ model }}` in generic test definitions with `{{ get_where_subque
...
+</File>
+
+<File name='macros/custom_get_where_subquery.sql'>
+
+```sql
+{% macro get_where_subquery(relation) -%}
+    {% set where = config.get('where', '') %}
+    {% if "__three_days_ago__" in where %}
+        {# replace placeholder string with result of custom macro #}
+        {% set three_days_ago = dateadd('day', -3, current_timestamp()) %}
```
dbeatty10: Adding in explicit namespace for `dateadd` (maybe `current_timestamp` should have it too?):
```suggestion
        {% set three_days_ago = dbt.dateadd('day', -3, current_timestamp()) %}
```

dbeatty10 (changes requested):
This is... `__wild__`! Thank you for documenting this possibility!

It's an example of very `__lightweight__`, roll-your-own [string interpolation](https://en.wikipedia.org/wiki/String_interpolation) using `__magic__` strings rather than start/end delimiters. Now I'm curious how far an `__enterprising__` person could get with defining and parsing their own delimiters!

(Feel free to go on your own [Mad Libs®](https://en.wikipedia.org/wiki/Mad_Libs) adventure and replace my comments above using the adjectives of your choice.)

---

Two things to resolve (both of which have inline suggestions):
1. Choose either "use" or "access" as the MadLibs verb to proceed "custom macros" (but probably not both)
2. Choose whether or not to namespace the `dateadd` and `current_timestamp` macros

dbeatty10 (approved):
Just committed some suggestions from my original review to get this merged.

If it turns out we actually wanted different text, there is nothing stopping us from continuing to make updates.

PR #2422, opened 2022-11-19
Update needed version of git to 2.26.0 for

brandon-segal:
## Description & motivation
Update the needed version of gt 2.26.0 to support sparse checkout. [git as a bug in the 2.25.0 version](https://stackoverflow.com/questions/72223738/failed-to-initialize-sparse-checkout) which was [later addressed in the 2.26.0](https://github.com/git/git/commit/47dbf10d8a5ce8c9b441a16b7698c7d70585dff0). This bug prevents users from running `dbt deps` when they have a git dependency within a subdirectory accessed via https.


mirnawong1 (approved):
hey @brandon-segal fab catch and congrats on opening up your first pr with us ✨! thanks for your patience and I've gone ahead and approved + merged this for you. 

appreciate your contribution!

PR #2435, opened 2022-11-21
Replace module-path with packages-install-path

joellabes:
## Description & motivation
In dbt v1 we changed from module-path to packages-install-path. The docs were correct but the example was still using the old version. 

I haven't versioned this content because 0.21 is about to roll out of even critical support. 

## Prerelease docs


nghi-ly (approved):
tysm for fixing this, @joellabes !

PR #2440, opened 2022-11-22
Update packages.md: bump dbt-utils version

MichelleArk:
## Description & motivation

Currently the example dbt-utils package is set to `revision: 0.1.21` in the packages documentation. If a user interested in testing out packages were to copy and paste the example directly (as I did!), they are met with an opaque error from the latest major version of dbt:

```
❯ dbt deps
16:26:38  Running with dbt=1.3.0
16:26:40  Encountered an error:
Runtime Error
  Invalid config version: 1, expected 2
```

This error is resolved with an more up-to-date version of dbt-utils in the example. 

## Prerelease docs
If this change is related to functionality in a prerelease version of dbt - N/A

## Checklist
If you added new pages - N/A

If you removed existing pages - N/A

### DIFF
--- a/website/docs/docs/build/packages.md
+++ b/website/docs/docs/build/packages.md
@@ -35,7 +35,7 @@ packages:
     version: 0.7.0
 
   - git: "https://github.com/dbt-labs/dbt-utils.git"
-    revision: 0.1.21
+    revision: 0.9.2
 
   - local: /opt/dbt/redshift
 ```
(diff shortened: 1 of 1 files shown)


dbeatty10 (approved):
Thanks for contributing this update @MichelleArk -- will plan on merging this today if possible.

[0.1.21](https://github.com/dbt-labs/dbt-utils/releases/tag/0.1.21) is certainly a blast from the past! Although we'd expect most folks to install dbt_utils via the `version` syntax, we definitely want all code examples on our website to actually work!

Note: @joellabes is working on releasing 1.0.0 soon-ish, so tagging him here as a nudge to consider creating a similar PR as this one to bump the version in the docs another time post release. It doesn't need to be in lock-step with every release, but should probably be modern-ish.

Should also consider this example at the same time:
```yaml
  - package: dbt-labs/snowplow
    version: 0.7.0
```

joellabes:
Yeah I like this! thank you 🎉 will loop back post 1.0

PR #2433, opened 2022-11-21
Create SQL glossary

johnblust:
## Description & motivation
Adds SQL glossary index to docs site and content that is ready to publish. This PR is still a WIP and is not ready for review

PR #2462, opened 2022-11-30
Combine glossary directories

johnblust:
## Description & motivation
Testing whether we can roll both of the glossaries under the same sidebar index.

@JKarlavige One thing that's problematic about this format IMO: the first folder is automatically open (in this case, the AE glossary). This means the user has to scroll the sidebar all the way down to see that there is a second glossary directory available (the SQL glossary). Is there a way to circumvent this?

@krevitt Do you know how to check for crawl depth on the SQL Glossary sidebar index before it goes live? It seems like the crawl depth would be 3 (homepage > /glossary > /avg), but not sure whether we should count the "Aggregate Functions" nested folder click in this equation. Either way, we should be okay, but the fewer clicks from a crawl depth perspective, the better.

@KiraFuruichi Is this what you had in mind? Anything you would change about it?


JKarlavige:
@johnblust If you're looking to have all the glossaries in one place, it might be best to create a "glossary index" category, and have all the glossaries nested within. See the screenshot and code below:

image

```
Glossary: [
     {
      type: "category",
      label: "Glossaries",
      link: {
        type: "generated-index",
        title: "All Our Glossaries",
        description:
          "Whole bunch of info on how cool these glossaries are and why you should read every last word.",
        slug: "/glossary/intro",
      },
      items: [
        {
          type: "category",
          label: "Analytics Engineering Glossary",
          link: {
            type: "generated-index",
            title: "Analytics Engineering Glossary",
            description:
              "The Analytics Engineering Glossary is a living collection of terms & concepts commonly used in the data industry. You can use and contribute to this resource to educate yourself, your team, and your stakeholders.",
            slug: "/glossary",
          },
          items: [
            {
              type: "autogenerated",
              dirName: "terms",
            },
          ],
        },
        {
          type: "category",
          label: "SQL Glossary",
          link: {
            type: "generated-index",
            title: "SQL Aggregate Functions",
            description: "Test",
            slug: "/sql-glossary",
          },
          items: [
            {
              type: "autogenerated",
              dirName: "sql-glossary",
            },
          ],
        },
      ],
    },
  ],
```

krevitt:
Hmm I'm not sure we'd want to intermingle these in the same left sidebar, what do you think @KiraFuruichi?  

I'm assuming w/ docusaurus that we have an all-or-nothing choice, and that having an index requires also accepting the sidebar intermingling.

If that's the case, I'd vote to keep them separate + link to each glossary separately in the learn menu (also keeps our crawl depth down on those pages).

KiraFuruichi:
I personally think it makes sense to keep them in the left sidebar if we continue developing multiple glossaries, so folks don't have to navigate to different areas on-site for glossary content. I do, however, understand the impact on crawl depth. @emilymermell-dbt do you have an opinion on this?

johnblust:
Thanks for the code @JKarlavige! After playing around with it a bit, I agree with David that having both glossaries in the same sidebar is a bit clunky (especially when there are so many pages/terms within each category). 

Jason, is there a way to create a navigational index page that then links out to the individual glossaries (like the one you created but without the sidebar)? This would be ideal as I agree with Kira in that not having a centralized place where we can link to all glossaries isn't a scalable solution as we continue to create more

JKarlavige:
> Thanks for the code @JKarlavige! After playing around with it a bit, I agree with David that having both glossaries in the same sidebar is a bit clunky (especially when there are so many pages/terms within each category).
> 
> Jason, is there a way to create a navigational index page that then links out to the individual glossaries (like the one you created but without the sidebar)? This would be ideal as I agree with Kira in that not having a centralized place where we can link to all glossaries isn't a scalable solution as we continue to create more

Without the sidebar, I don't believe it's possible to use the `generated-index` type and build out the pages using the `sidebars.js` file. Instead, creating a docs page similar to the [Join the Community](https://docs.getdbt.com/community/join) page might be what you're looking for, where they used the `Card` component to build the page out: https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/community/join.md. 

And as long as this new docs page isn't added to the `sidebars.js` file it wouldn't have a sidebar.

johnblust:
Okay I think I got it!!! Not a blocker, but there is a slight styling issue @JKarlavige: 

Although there is no sidebar on this page, it looks like the page wants there to be one. All the content is slightly aligned left (as if to hug the sidebar) and there is a darker section of the main nav logo. Once again, not a dealbreaker, but is there a way I can fix this? Maybe using another `div` style instead of `community-home`?

Screen Shot 2022-12-01 at 3 57 17 PM

EDIT: Also, do you know where I can find a list of icons I can use for the `icon:` field?

JKarlavige:
> Okay I think I got it!!! Not a blocker, but there is a slight styling issue @JKarlavige:
> 
> Although there is no sidebar on this page, it looks like the page wants there to be one. All the content is slightly aligned left (as if to hug the sidebar) and there is a darker section of the main nav logo. Once again, not a dealbreaker, but is there a way I can fix this? Maybe using another `div` style instead of `community-home`?
> 
> EDIT: Also, do you know where I can find a list of icons I can use for the `icon:` field?

Woo! I actually have a fix for the navbar in a separate PR. I'll push those same edits to this branch. For the icons, the full list is available in [this directory](https://github.com/dbt-labs/docs.getdbt.com/tree/current/website/static/img/icons)

JKarlavige:
@johnblust just pushed a commit with the navbar fix, so that should be corrected if you pull down the latest code from here. And fyi there's not a preview available for this PR as it's not set to merge into `current`, however once this is merged into #2433, the preview should be available there!

Note - I updated this branch to `current` which includes the Docusaurus 2.2.0 update, so disregard the high number of files changed within this PR, most of those edits are already live.

nghi-ly:
ok, i think i'm following this PR:

- have different glossaries live in the same "space" 
- use the "card landing page" component for each glossary, instead of using the left sidebar 

it's not documented in the docs style guide but we do have a habit of adding a left sidebar entry for every article (docs page). though maybe this is a use case/situation for not doing that. personally, i have no fud with it.

curious, what are your thoughts, @runleonarun ? if you also have no fud, maybe we can document this for future guidance?

nghi-ly (commented):
ok, i think i'm following this PR:

- have different glossaries live in the same "space"
- use the "card landing page" component for each glossary, instead of using the left sidebar

it's not documented in the docs style guide but we do have a habit of adding a left sidebar entry for every article (docs page). though maybe this is a use case/situation for not doing that. personally, i have no fud with it.

curious, what are your thoughts, @runleonarun ? if you also have no fud, maybe we can document this for future guidance?

runleonarun:
Thanks for the ping @nghi-ly!

I think the cards are a nice way to organize content if you're avoiding a sidebar.  👍🏻 

I have a few other thoughts:
* I wonder if people would get confused with a reference section under 
  "Glossaries" when we also have a Reference section in `/references`. We're also repeating "glossary" a lot. Maybe we should simplify the URLs a little: 
  * Glossary > SQL (`docs.getdbt.com/glossary/SQL/term`)
  * Glossary > Analytics engineering (`docs.getdbt.com/glossary/analytics-engineering/term`)

* Should we keep "Glossary" singular? It feels easier to read, and it's still our One Glossary, but now with categories.
* Heads up that we'll eventually need a more general dbt glossary for translations teams to use.

johnblust:
Thanks for taking a look, Leona! Some very good questions here. I'll try to address them.

> * I wonder if people would get confused with a reference section under
>   "Glossaries" when we also have a Reference section in `/references`. We're also repeating "glossary" a lot. Maybe we should simplify the URLs a little:

The concern about the confusion between this and `/reference` is a valid one! I thought about this when making the decision to rename this content to SQL reference. Ultimately, I think creating a clear separation between product content and broader educational topics is the best way to avoid that confusion. The alternative of putting SQL Reference content with the dbt Reference content would prove to be more confusing in my opinion, but let me know if you disagree.

As for the URL question, the two main reasons I didn't want to associate them underneath the same `/glossary` string, is because 1) they are very different in terms of content format and theme, and 2) having `/sql-reference` in the URL will help us rank for the term "SQL reference" in Google. Let me know if that makes sense or if you would still need these to be underneath the same URL pattern.

> Should we keep "Glossary" singular? It feels easier to read, and it's still our One Glossary, but now with categories.

Yeah, that makes sense to me! Good idea. 

> Heads up that we'll eventually need a more general dbt glossary for translations teams to use.

That's good to know, thanks for the info! Do you know when this will be happening and where on the website this will live? As early, I think keeping a clear separation between product content and educational content is the key to avoiding confusion here. Maybe even changing the title of this Hub page to "Educational glossary and references" would be a good decision.

runleonarun:
Closing until we can come up with a cohesive strategy as proposed in [this issue](https://github.com/dbt-labs/docs.getdbt.com/issues/3910).

PR #2489, opened 2022-12-05
create dbt Ecosystem section and move other content

mirnawong1:
https://www.notion.so/dbtlabs/Evaluate-Adapter-development-location-v1-393baab818ff4e038ce510ffbb9879f3
https://www.notion.so/dbtlabs/Re-evaluate-dbt-Ecosystem-IA-v2-2750603111d54b46b5f10b2f2acf2b83

moving adapter to dbt ecosystem per slack thread here: https://dbt-labs.slack.com/archives/C04BN4VPK98/p1670003026686809

and moving some legacy content to best practices + advanced per here: https://dbt-labs.slack.com/archives/C032RA0PA9F/p1669364164603739

## Prerelease docs

## Checklist

- [ ] [Run link testing](https://github.com/dbt-labs/docs.getdbt.com#running-the-cypress-tests-locally) to update the links that point to the deleted page


mirnawong1:
as a note, the links in 'supported data platforms' aren't being redirected as they should: https://deploy-preview-2489--docs-getdbt-com.netlify.app/docs/supported-data-platforms

jasnonaz:
Thank you for putting this together Mirna! Some thoughts that I would love to hear your / Leona's opinion on.

- Do we want to call the ecosystem section dbt ecosystem? That would seem to imply that all of the organizations exist there within our orbit and we may or may not want to imply that vs. a more general "ecosystem".
- There are now three sections with 1 or 2 guides in them (Ecosystem, Advanced and Orchestration). Do we think that this is desirable? Should we try and consolidate?
- I wonder if there is a way in Docusaurus to override the Summary on these cards.  For the testing one in particular it would be nice to have a different intro.
image
- This is not a requirement to get this merged but just since we're looking at the summary pages - wonder if we should swap the summary pages out to be the beautiful new cards we put together for the DevHub launch. 
image

runleonarun:
Hey Jason! This is awesome feedback! 

> There are now three sections with 1 or 2 guides in them (Ecosystem, Advanced and Orchestration). Do we think that this is desirable? Should we try and consolidate?

> This is not a requirement to get this merged but just since we're looking at the summary pages - wonder if we should swap the summary pages out to be the beautiful new cards we put together for the DevHub launch.

I [created an issue](https://github.com/dbt-labs/docs.getdbt.com/issues/2506) for these two bullet points. I think we can address either with Ly's quickstart project OR when with our IA clean up work. 

> Do we want to call the ecosystem section dbt ecosystem? That would seem to imply that all of the organizations exist there within our orbit and we may or may not want to imply that vs. a more general "ecosystem".

@jasnonaz Do you think we lose any clarity by using the more general "Ecosystem?" Regardless, I do appreciate your point. I wonder if there's another adjective with can use? "Developer ecosystem" maybe? I think that would include partners and community developers. @mirnawong1 thoughts? 

> I wonder if there is a way in Docusaurus to override the Summary on these cards. For the testing one in particular it would be nice to have a different intro.

@mirnawong1 both of these seem like reasonable suggestions to me pending my question above! 

Let me know if you need help adding the summaries. I believe we can edit the frontmatter of each page where the `hoverSnippet` populates the card description and the `description` is used for searches. Looks like this in the glossary:
```
---
id: json
title: JSON
description: JSON (JavaScript Object Notation) is a minimal format for semi-structured data used to capture relationships between fields and values.
displayText: JSON
hoverSnippet: JSON (JavaScript Object Notation) is a minimal format for semi-structured data used to capture relationships between fields and values.
---
```

jasnonaz:
Awesome thanks Leona!

You raise a good point here - I don't know that dbt Ecosystem is the best possible description for this but it is clearer than the alternatives I can come up with off the top of my head which would be
- Ecosystem
- Developer ecosystem
- Data tooling ecosystem
- Analytics ecosystem
So I guess we stick with dbt Ecosystem? Apologies on taking us down a rabbit hole then going back to the original wording. Other than that happy to have this go live once we get the descriptions tweaked - thank you Mirna!!

mirnawong1:
thanks Jason and Leona for that invaluable feedback! i’ve made the changes to the description/summary:

deploy link here

> I wonder if there is a way in Docusaurus to override the Summary on these cards. For the testing one in particular it would be nice to have a different intro.

- completely agree with this, fab catch Jason! i’ve gone ahead and updated the summaries and thanks for the guidance, Leona!

> Do we want to call the ecosystem section dbt ecosystem? That would seem to imply that all of the organizations exist there within our orbit and we may or may not want to imply that vs. a more general “ecosystem”.

- i wasn’t involved in the initial discover of ‘dbt ecosystem’ but i think it’s a step in the right direction - how do we feel about `devhub ecosystem`? or the shorten `dev ecosystem`?  i feel like it sits well within the ‘devhub’ brand (e.g devhub blog, dbt developer hub, maybe a future devhub guides?) if non of those feel right, then i think we can stick with dbt ecosystem until we reiterate further - does that make sense?

runleonarun (approved):
This looks great, Mirna!

PR #2490, opened 2022-12-05
Update check_cols.md

ilmari-aalto:
Switch to `check_cols` column name in `check_cols` documentation page

## Description & motivation

Since this page is about `check_cols` and since the option `all` is listed, surely the configuration should be for `check_cols` instead of `unique_key`?


mirnawong1 (approved):

mirnawong1:
hey @ilmari-aalto great catch and thanks for opening up this fix! I've approved and merged this for you ✨

PR #2491, opened 2022-12-05
distinguish URLs for app and user token approaches in Github private …

matt-winkler:
…packages

## Description & motivation

The current documentation is incorrect when using a personal access token to clone private packages. With this approach, *both* the username and token need to be included in the URL. This adds examples for both the app token and personal token methods with comments to clarify.


schottj (approved):
amazing - thanks for correcting!

PR #2507, opened 2022-12-06
Fix return.md example

szn:
## Description & motivation

The example given to `return` defined a `get_data()` macro and then calls it in a model as `getdata()`. This can't work of course.

Also, the macro file name is renamed to `macros/get_data.sql` instead of `macros/example.sql`.
## Prerelease docs
n/a

## Checklist
n/a


mirnawong1 (approved):
hey @szn great contribution, thank you for opening this up. I've approved and merged this for you!

PR #2516, opened 2022-12-07
Update core versions table: v1.0 EOL, into 2023

jtcohen6:
Apparently we're going to keep building `dbt-core`, and releasing some more versions next year!

Also, v1.0 reached End Of Life on Dec 3 (https://github.com/dbt-labs/docs.getdbt.com/pull/2449). It's no longer under "official support," meaning that v1.0.8 was its final patch release; there won't be any others.

**Questions:**
- Should we split up into multiple tables: currently supported versions, future versions, no-longer-supported versions?
- **Support in dbt Cloud:** We're still doing "soft" drawdowns of old versions; we aren't yet in a place of forcefully upgrading accounts, or removing old versions. How should we reflect that on the docs site? [Here's the answer](https://getdbt.slack.com/archives/C37J8BQEL/p1669030531253179?thread_ts=1666014898.024699&cid=C37J8BQEL) I gave a community member:
> After Dec 3, we won't be releasing any new patches of v1.0.x, we won't be providing the same level of ongoing support in Cloud, and we'll strongly encourage you to upgrade to a newer version. I'm hoping it's a relatively painless upgrade — there should be minimal friction, and zero breaking changes.

### DIFF
--- a/website/snippets/core-versions-table.md
+++ b/website/snippets/core-versions-table.md
@@ -1,10 +1,13 @@
 | dbt Core                        | Initial Release | Active Support Until | Critical Support Until  | dbt Cloud Until | Final Patch |
 |---------------------------------|-----------------|----------------------|-------------------------|-----------------|-------------|
 | **v0.X**                        | (various dates) | v1.0.0 release       | Dec 3, 2021             | See below ⚠️     | v0.21.1     |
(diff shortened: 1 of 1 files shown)


mirnawong1:
hey @jtcohen6 was going to have a look at this but may be best for @matthewshaver to review it as we discussed versioning yesterday. since it's related to core and versioning - you may have more ideas, Matt!

jtcohen6:
Closing in favor of #2534, which will do this & a whole lot more!

PR #2545, opened 2022-12-09
fix doris-setup id

johnblust:
## Description & motivation
Fixing `current` which is now broken due to an incorrect syntax

### DIFF
--- a/website/docs/reference/warehouse-setups/doris-setup.md
+++ b/website/docs/reference/warehouse-setups/doris-setup.md
@@ -1,6 +1,6 @@
 ---
 title: "Doris setup"
-id: "Doris-setup"
+id: "doris-setup"
 meta:
   maintained_by: SelectDB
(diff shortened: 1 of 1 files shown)


lfracalossiFA:
Hi @johnblust ,
Is there a particular page I should be looking at, where the error occurred?

johnblust:
> Hi @johnblust , Is there a particular page I should be looking at, where the error occurred?

Hey @lfracalossiFA, `website/docs/reference/warehouse-setups/doris-setup.md` had a capitalized "D" in the in the `id` field so the sidebar.js was breaking due to wrong syntax

lfracalossiFA:
Ah, so as long as the sidebar isn't breaking, we're good. Got it. I'll take a peek.

lfracalossiFA (approved):
lgtm! 🚢

PR #2269, opened 2022-10-17
Pyodide testing

JKarlavige:
Testing Pyodide for dbt in the browser. Do not merge.

Pyodide testing: https://deploy-preview-2269--docs-getdbt-com.netlify.app/test
CLI Prototype page: https://deploy-preview-2269--docs-getdbt-com.netlify.app/docs/dbowser-demo

## 1/3/23 Update
This PR will continue to hold the pyodide/python work, but the rest of the work is now moving forward in PR #2643, which removes the pyodide testing work for the static release

PR #2565, opened 2022-12-12
Private Package Testing

heysweet:
## Description & motivation

image
_Lineage view in the Editor_

image
_What a developer will see without the optional private dependency_

Adding the Lineage View as a private package into the dbtEditor logic. At the time of writing this PR, there's a minor bug in the DAG logic which is causing the change of the primaryNodeId to not re-render the DAG. I'll follow-up with a hotfix for this issue.

This PR will be merged into the `pyodide-testing` branch, and will be eventually merged with [PR #2269
](https://github.com/dbt-labs/docs.getdbt.com/pull/2269)


website/docusaurus.config.js
```diff
@@ -228,6 +228,17 @@ var siteSettings = {
...
+      return {
+        name: "docusaurus-tailwindcss",
+        configurePostCss(postcssOptions) {
+          // Appends TailwindCSS and AutoPrefixer.
+          postcssOptions.plugins.push(require("tailwindcss"));
+          postcssOptions.plugins.push(require("autoprefixer"));
+          return postcssOptions;
+        },
+      };
+    },
```
heysweet: If we can improve how we bundle in styles, we won't need to include tailwind

website/package.json
```diff
@@ -6,6 +6,8 @@
     "test": "npx jest --watch"
   },
   "dependencies": {
+    "@dbt-labs/react-dbt-dag": "0.7.10",
```
heysweet: If we change what we bundle in, this should be the only thing we'd have to include

website/src/components/lineage/utils/transform-lineage-nodes.js
```diff
@@ -0,0 +1,17 @@
...
+ * [another format](https://github.com/dbt-labs/react-dbt-dag/blob/main/src/dag/InternalDagInterfaces.ts#L34-L48).
+ */
+export const transformLineageNodes = (nodes) => {
+    const dagNodes = Object.entries(nodes || {}).map(([nodeId, node]) => {
+        return {
+            id: nodeId,
+            parents: node.depends_on?.nodes || [],
+            label: node.name,
+            resourceType: node.resource_type,
+        }
```
heysweet: This is all the transformation we need for manifest to dag!

website/src/components/lineage/index.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+/** As a private package, not every devleoper will have access to this repo. */
+const DagExports = require('@dbt-labs/react-dbt-dag');
+
+export const Lineage = ({ dagNodes, currentNodeId }) => {
+
+    if (!DagExports) {
+        return <div>Lineage Not Available!</div>
+    }
```
heysweet: I haven't tested this out thoroughly, but @jgiuffrida claimed something like this pattern would allow other developers to still access this in development while not having access to the private repo

website/src/components/dbt-editor/tailwindStyles.css
```diff
@@ -0,0 +1,7 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
heysweet: We shouldn't need to include tailwind long term if we improve how we bundle this repo

website/src/components/dbt-editor/tailwindStyles.css
```diff
@@ -0,0 +1,7 @@
...
+
+.dbt-dag {
+    height: 100%;
+}
```
heysweet: I think this is just a bug with the repo? We really shouldn't need this

heysweet:
image

This is the last real mystery to me @jgiuffrida -- I've marked the modules as optional and switched to dynamic imports, but it's still causing issues in development

heysweet:
> image
> 
> This is the last real mystery to me @jgiuffrida -- I've marked the modules as optional and switched to dynamic imports, but it's still causing issues in development

It 100% works! It just causes this overlay to be displayed on page load, and I'd much prefer to hide this.

website/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = {
+  presets: [dagTheme, sourdoughTheme],
+  darkMode: 'class',
+  content: [
+    './src/**/*.{js,jsx,ts,tsx,mdx}',
+    './node_modules/@dbt-labs/sourdough/dist/sourdough.js',
+    './node_modules/@dbt-labs/react-dbt-dag/dist/react-dbt-dag.es.js',
+  ],
+};
```
heysweet: Delete!

JKarlavige (approved):
- Able to install `react-dbt-dag` package once appropriate token and scopes set.
- Verified DAG running within editor component on demo page. 
- Verified `Lineage Not Available!` message shows if package not installed, and no other errors appear on the site or in the console.

lgtm! :shipit:

PR #2526, opened 2022-12-07
Text clean-up

deanna-minnick:
## Description & motivation
This PR removes whitespace from an inline code snippet and removes an "Options" header that had no contents.

PR #2588, opened 2022-12-15
fix internal link

deanna-minnick:
## Description & motivation
This PR updates a broken link by converting square brackets to parentheses in `dbt-classes.md`. It was previously displaying as "[quoting config][quoting]" on the website.

This PR also adds the full path (in keeping with PR #2526). I tested it by pasting `/reference/project-configs/quoting` after `https://docs.getdbt.com` in my browser, and it took me to the quoting config page as expected. I'd be glad for feedback if there are other ways to test and/or best practices regarding this!


mirnawong1 (approved):
ack, great catch @deanna-minnick, and this is correct, thanks for correcting that for us as always! approved and ready for you to merge!

PR #2599, opened 2022-12-19
combine-state-result-selector-description

jeremyholtzman:
## Description & motivation

@bethanyhipple-dbtlabs and I discovered that the description of the command should be `OR` not `AND` because the set operator used is a ` ` and not a `,`


matthewshaver:
Thank you @jeremyholtzman and @bethanyhipple-dbtlabs for catching and correcting this.

matthewshaver (approved):

PR #2661, opened 2023-01-05
dbt-materialize: update Materialize profile and configs page

morsapaes:
Give some love to the Materialize profile, which had a couple of broken front matter variables.

(CC @dehume for a glimpse into the contribution process to dbt docs!)


dataders (approved):
thanks @morsapaes!

PR #2673, opened 2023-01-06
Prevent Chrome lookalike errors for docs and www

john-rock:
# Code Task Complete ⚡️

✨ View the Asana Task [here](https://app.asana.com/0/1200099998847559/1203325932107038/f) ✨

## Description of Changes
- Adds `assetlinks.json` per [Google's specs](https://chromium.googlesource.com/chromium/src/+/master/docs/security/lookalikes/lookalike-domains.md#automated-warning-removal) to prevent lookalike warnings when viewing deploy previews.

## Additional Info
Companion PR to https://github.com/dbt-labs/www.getdbt.com/pull/1866. Both must be merged at the same time. Once both are merged, we then apply for automatic verification through Google.

I did a bit of research on this and could not find any resources regarding the lookalike warning for deploy previews. We will need to monitor this after submitting approval and may need to adjust the `assetlinks.json` file to include the deploy preview domain if we continue to see these warnings.

### DIFF
--- a/.well-known/assetlinks.json
+++ b/.well-known/assetlinks.json
@@ -1,0 +1,11 @@
+[
+    {
+      "relation": ["lookalikes/allowlist"],
+      "target": { "namespace": "web", "site": "https://getdbt.com" }
+    },
+    {
+      "relation": ["lookalikes/allowlist"],
+      "target": { "namespace": "web", "site": "https://docs.getdbt.com" }
+    }
+  ]
+  


JKarlavige:
Like you mentioned this take some trial and error since it doesn't look like we can test this until the request is submitted. 

On your note about potentially having to add the preview url - there's one note in the Google Specs page that mentions `Do not use subdomains (e.g. use “example.com”, not “www.example.com”).`, which makes me think this won't work as is, and that we might have to put `https://netlify.app` rather than `https://docs.getdbt.com`. 

But i'm not entirely sure here, open to submitting it as-is like you said and making any adjustments from there if needed. Curious to what others think.

lfracalossiFA:
After doing a little bit more digging here, I'm not sure this solution will work, either.

The problem is that Chrome thinks `deploy-preview-XXXX--docs-getdbt-com.netlify.app` is the site we don't own. (I think it knows we own docs.getdbt.com, though, since it is the recommended option). But we also couldn't just put "netlify.app," in assetlinks.json, as we *don't* own that. We also can't put a subdomain (according to the note Jason found), but even if we could, the subdomain changes for each PR, based on the PR number.

Some thoughts:
- It does seem like there's an option for a manual review, so maybe we could do that? It might allow us to explain the situation to a real human being...
- I don't think we get these messages on www, and I've long suspected that it is due to how the deploy preview URL is set up in Netlify. Have you investigated that, to see what the differences there might be?
- We should probably reach out to Netlify, as well, since their deploy preview URLs are causing this issue 😆 ). If you've already done that, please ignore this suggestion!

That said, there's probably no harm in merging this PR... I just don't think it will do much.

runleonarun (approved):
It's worth a whirl to see if it works!

john-rock:
Thanks for the feedback y'all. I'll go ahead a merge this PR and agree 100% with your comment @lfracalossiFA "there's probably no harm in merging this PR... I just don't think it will do much."

I will reach out to Netlify to see if they have any suggestions from their side.

john-rock:
Actually looking at this more the manual verification let's us provide more details and gives us an option if we don't own the site, like in our case when the error is thrown on the Netlify deploy preview link. I'm going to close these PRs as they are not needed for the manual review process and go ahead and submit a manual review.

PR #2699, opened 2023-01-12
In-page anchors not working when loaded from URL

john-rock:
[Asana Task](https://app.asana.com/0/1200099998847559/1202821015046958/f)

- Allows anchor links to work when loaded from a URL

## Preview Links
https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#redshift-unload-one-model-to-s3

https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/sql_header#create-a-bigquery-temporary-udf

https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#execution-ordering

https://deploy-preview-2699--docs-getdbt-com.netlify.app/docs/collaborate/manage-access/audit-log#repository

## Additional Info
This seems to be caused by our custom work on the TOC component. This custom function runs after the TOC is fully loaded and only if there is an anchor link in the URL.


website/src/utils/use-hash-link.js
```diff
@@ -0,0 +1,10 @@
+export default function useHashLink() {
+  if (window.location.hash) {
+    const hashLink = document.getElementById(
+      window.location.hash.replace("#", "")
+    );
+    if (hashLink) {
+      hashLink.scrollIntoView();
```
JKarlavige, line 20: TIL this exists 👀

JKarlavige (dismissed):
lgtm, confirmed the issue occurs on the live site, and tested all preview urls and confirmed they go to the correct heading. :shipit:

website/src/utils/use-hash-link.js
breezyfasano: thought the same thing! like wow this made it such an elegant and native solution. veryyy nice

breezyfasano (dismissed):
Tested out the previews and everythings lookin' great! Simple, clean, and elegant solution!

joellabes (commented):
Yay! I did notice one piece of odd behaviour, but it's inconvenient as opposed to a blocker. 

For [the Redshift link](https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#redshift-unload-one-model-to-s3), the top navbar obscured the h3 element itself, but the [pre hook one](https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#execution-ordering) correctly hides the navbar so that the h3 remains visible. Examples:
image
image

Why is there inconsistent behaviour? Ideally the header wouldn't be obscured by the navbar, either because it scrolls slightly lower or because the navbar dismisses itself. I think I've raised this sort of thing in the past and it's difficult, but I'm mentioning it again since it _does_ work sometimes. 

I tested what happens if you combine a version specifier with a hash link, and [it does work](https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook?version=1.0#grant-privileges-on-a-model) - nice!

john-rock:
> Yay! I did notice one piece of odd behaviour, but it's inconvenient as opposed to a blocker.
> 
> For [the Redshift link](https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#redshift-unload-one-model-to-s3), the top navbar obscured the h3 element itself, but the [pre hook one](https://deploy-preview-2699--docs-getdbt-com.netlify.app/reference/resource-configs/pre-hook-post-hook#execution-ordering) correctly hides the navbar so that the h3 remains visible.

Hey @joellabes that behavior is odd and not only is it inconsistent on links it also seems to be inconsistent across browser. Firefox appears to hide the navbar but chrome only hides it sometimes.

I just added a check in the utility function to force the navbar to be hidden by default if a page is loaded from a hash link. With this I'm now seeing consistent behavior across browsers and pages. I don't particularly love this solution as it's a bit hacky but rolled with it in an effort to prevent having to swizzle the entire navbar component. From what I see it doesn't appear to have negative side effects but let me know what you see on your end!

@JKarlavige @breezyfasano requesting another round of review from y'all if you wouldn't mind 🙂.

JKarlavige (approved):
lgtm! tested with both firefox and chrome and the nav automatically hides when loading from a hash with all preview links.

PR #2681, opened 2023-01-09
Adding 'incremental_predicates1 information

matthewshaver:
## What are you changing in this pull request and why?
Adds new information about `incremental_predicates` feature released with dbt Core v1.4

This was added in https://github.com/dbt-labs/dbt-core/pull/5702

## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) and [About versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) so my content adheres to these guidelines.
- [ ] Add a checklist item for anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

Adding new pages (delete if not applicable):
- [ ] Add page to `website/sidebars.js`
- [ ] Provide a unique filename for the new page

Removing or renaming existing pages (delete if not applicable):
- [ ] Remove page from `website/sidebars.js`
- [ ] Add an entry `_redirects`
- [ ] [Ran link testing](https://github.com/dbt-labs/docs.getdbt.com#running-the-cypress-tests-locally) to update the links that point to the deleted page

Issue #2636, opened 2023-01-02
Document addition of `incremental_predicates`


jtcohen6:
### Contributions

- [X] I have read the contribution docs, and understand what's expected of me.

### Link to the page on docs.getdbt.com requiring updates

https://docs.getdbt.com/docs/build/incremental-models

This might want to go under the "About `incremental_strategy` section," which is really acting as our catch-all for "advanced optimizations."

### What part(s) of the page would you like to see updated?

Support for a new (optional) config: `incremental_predicates`. This config accepts any valid SQL expression. (Note: dbt will not do any validation in advance, so it's up to the user to ensure that their SQL syntax is valid.)

This is an **advanced** use of incremental models, where data volume is large enough to justify additional investments in performance. It's a real case of "trust the user," where some of the abstraction that dbt offers over vendor-specific `merge` statements is stripped away.

For instance, this is a pattern we might expect to see on Snowflake:
```yml
models:
  - name: my_incremental_model
    config:
      materialized: incremental
      unique_key: id
      cluster_by: ['session_start']  # this will affect how the data is stored on disk, and indexed to limit scans
      incremental_strategy: merge
      # limit scan of the existing table to the last 7 days of data
      incremental_predicates: "DBT_INTERNAL_DEST.session_start > datediff(day, -7, current_date)"
```

This will template a `merge` statement like:
```sql
merge into <existing_table> DBT_INTERNAL_DEST
    from <temp_table_with_new_records> DBT_INTERNAL_SOURCE
    on
        -- unique key
        DBT_INTERNAL_DEST.id = DBT_INTERNAL_SOURCE.id
        and
        -- custom predicate: limits data scan in the "old" data / existing table
        DBT_INTERNAL_DEST.session_start > datediff(day, -7, current_date)
    when matched then update ...
    when not matched then insert ...
```

The user will still want to limit data scan of _upstream_ tables, like always, within the body of their incremental model SQL, to limit the amount of "new" data being processed / transformed:
```sql
with large_source_table as (

    select * from {{ ref('large_source_table') }}
    {% if is_incremental() %}
        where session_start > dateadd(day, -3, current_date)
    {% endif %}

),

...
```

Notes:
- Users do need to explicitly alias any columns with either `DBT_INTERNAL_DEST` ("old" data) or `DBT_INTERNAL_SOURCE` ("new" data)
[...]

PR #2707, opened 2023-01-16
Fix v1.4 migration guide link

jtcohen6:
Update link to issue (https://github.com/dbt-labs/docs.getdbt.com/issues/2636) -> link to docs that were added in https://github.com/dbt-labs/docs.getdbt.com/pull/2681

Opening as a PR because I want to verify that the Netlify build works + looks correct. Assuming it does, I'm going to use my Special Permissions to get this live without a review.

### DIFF
--- a/website/docs/guides/migration/versions/04-upgrading-to-v1.4.md
+++ b/website/docs/guides/migration/versions/04-upgrading-to-v1.4.md
@@ -42,7 +42,7 @@ Link to GH discussion forthcoming
 
 ### Advanced configurations for incremental models
 
-- [**`incremental_predicates`** config](incremental-models#about-incremental_predicates) is now supported on the most popular adapters, enabling greater flexibility when tuning performance in `merge` and `delete` statements against large datasets.
+- [**`incremental_predicates`** config](/docs/build/incremental-models#about-incremental_predicates) is now supported on the most popular adapters, enabling greater flexibility when tuning performance in `merge` and `delete` statements against large datasets.
 - **BigQuery:** The `insert_overwrite` incremental strategy supports a new (old) mechanism, [`time_ingestion_partitioning`](bigquery-configs#partitioning-by-an-ingestion-date-or-timestamp) + [`copy_partitions`](#copying-ingestion-time-partitions), that can yield significant savings in cost + time for large datasets.
 
 ### Updates to Python models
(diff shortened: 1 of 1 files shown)


mirnawong1 (dismissed):
hey@jtcohen6 , i think the link isn't working just right in the preview

website/docs/guides/migration/versions/04-upgrading-to-v1.4.md
```diff
@@ -42,7 +42,7 @@ Link to GH discussion forthcoming
 
 ### Advanced configurations for incremental models
 
-- The most popular adapters now support an `incremental_predicates` config, to enable greater flexibility when tuning performance in `merge` and `delete` statements for large datasets. ([dbt-labs/docs.getdbt.com/issues/2636](https://github.com/dbt-labs/docs.getdbt.com/issues/2636))
+- [**`incremental_predicates`** config](incremental-models#about-incremental_predicates) is now supported on the most popular adapters, enabling greater flexibility when tuning performance in `merge` and `delete` statements against large datasets.
```
mirnawong1: trying the new link

```suggestion
- [**`incremental_predicates`** config](/docs/build/incremental-models#about-incremental_predicates) is now supported on the most popular adapters, enabling greater flexibility when tuning performance in `merge` and `delete` statements against large datasets.
```

mirnawong1: @jtcohen6 i've updated the link to the above because the netfliy built URL kept taking me to 'Filtering rows on an incremental run' instead of 'incremental predicates'

jtcohen6: Let's give it a try!

mirnawong1 (approved):
approving it for you @jtcohen6 🙏

website/docs/guides/migration/versions/04-upgrading-to-v1.4.md
jtcohen6: @mirnawong1 I'm seeing the same behavior. Unfortunately, I think this is an underlying "engineering" issue. If I had to guess, it has to do with the sequence in which:
1. the vertical position of each anchor is determined, and the anchored link goes to that spot
2. the page rendering completes, vertical positions change, the page "moves" from its anchored point

Let's open a separate issue for this?

mirnawong1: no worries, its def not a blocker so I will open up an asana task for the eng team on this. thank you!

jtcohen6: thank you!

PR #2715, opened 2023-01-17
Fix `firstVersion` typo

dbeatty10:
Should this line contain `firstVersion` instead of `firstVersiont` (which looks like a typo)?

## Checklist
No checklist items were applicable.


mirnawong1 (approved):
🦅 👀s - thank so much @dbeatty10 !

PR #2716, opened 2023-01-17
Add the `append` incremental strategy for Snowflake

dbeatty10:
It looks like the dbt Snowflake adapter has an `append` strategy which is listed [here](https://github.com/dbt-labs/dbt-snowflake/blob/4b7ee7372d87efc2b9c6929203fcfb41930b0b3e/dbt/adapters/snowflake/impl.py#L236-L237). We should list it!

## Checklist
None of the checklist items are applicable.


mirnawong1 (approved):
thanks again for this and great idea! approved @dbeatty10 !

should we link the user to this github link: https://github.com/dbt-labs/dbt-snowflake/blob/4b7ee7372d87efc2b9c6929203fcfb41930b0b3e/dbt/adapters/snowflake/impl.py#L236-L237

dbeatty10:
> should we link the user to this github link: https://github.com/dbt-labs/dbt-snowflake/blob/4b7ee7372d87efc2b9c6929203fcfb41930b0b3e/dbt/adapters/snowflake/impl.py#L236-L237

This is an interesting idea to include a link to the relevant source code! Up for whatever you think is best. The main thing I'd be worried about the content of the link going stale. If we include a link, I would rather a permalink (like above) so at least it's accurate at a point-in-time. Then a user can compare it to whatever the latest content is to determine if it has changed in the interim.

What do you think?

McKnight-42 (approved):

mirnawong1:
ah gotcha @dbeatty10, i guess nothing else is linked and it may be more effort for the user if we include the link. nonblocking curiosity on my end! thank you for clarifying!

dbeatty10:
@mirnawong1 Sure thing! I'll merge now, and we can definitely circle back at a later date if we decide to include a link to the source code.

PR #2718, opened 2023-01-17
Add SQL Reference content bactch #9

johnblust:
Adds INNER, OUTER, CROSS, LEFT, RIGHT, SELF Join pages.
Adds CASE statement page.
Fix WHERE clause page.

## Checklist

- [x] Add page to `website/sidebars.js`
- [x] Provide a unique filename for the new page


website/docs/sql-reference/clauses/sql-where.md
```diff
@@ -26,7 +26,7 @@ from {{ ref('orders') }}
 where status != 'returned'
 ```
 
-In this query, you’re filtering for any order from the [Jaffle Shop’s](https://github.com/dbt-labs/jaffle_shop) `orders` model whose status is not returned by adding a WHERE clause after the FROM statement. You could additionally filter on string, numeric, date, or other data types to meet your query conditions.
+In this query, you’re filtering for any order from the [Jaffle Shop’s](https://github.com/dbt-labs/jaffle_shop) `orders` model whose status is not `returned` by adding a WHERE clause after the FROM statement. You could additionally filter on string, numeric, date, or other data types to meet your query conditions.
 
 You will likely see WHERE clauses show up 99.99%  of the time in a typical query or dbt model. The other .01% is probably in a <Term id='dml' />  statement, such as DELETE or ALTER, to modify specific rows in tables.
```
KiraFuruichi:
```suggestion
You will likely see WHERE clauses show up 99.99%  of the time in a typical query or dbt model. The other .01% is probably in a <Term id="dml" />  statement, such as DELETE or ALTER, to modify specific rows in tables.
```

website/docs/sql-reference/joins/sql-left-join.md
```diff
@@ -0,0 +1,76 @@
...
+| 3 | green |
+| 4 | yellow |
+
+```sql
+select
+   car_type.user_id as user_id,
+   car_type.car_type as type,
+   car_color.car_color as color
+from {{ ref('car_type') }} as car_type
+left join {{ ref('car_color') } as car_color
```
KiraFuruichi:
```suggestion
left join {{ ref('car_color') }} as car_color
```

website/docs/sql-reference/joins/sql-right-join.md
```diff
@@ -0,0 +1,70 @@
...
+| 3 | green |
+| 4 | yellow |
+
+```sql
+select
+   car_type.user_id as user_id,
+   car_type.car_type as type,
+   car_color.car_color as color
+from {{ ref('car_type') }} as car_type
+right join {{ ref('car_color') } as car_color
```
KiraFuruichi:
```suggestion
right join {{ ref('car_color') }} as car_color
```

website/docs/sql-reference/joins/sql-cross-join.md
```diff
@@ -0,0 +1,74 @@
...
+title: SQL CROSS JOIN
+description: A cross join typically takes two columns between two database objects and creates a table forming a combination of all rows across joined tables, called a cartesian product.
+slug: /sql-reference/cross-join
+---
+
+<head>
+    <title>Working with cross joins in SQL</title>
+</head>
+
+A truly rarely seen, but important join: the cross join. The majority of your analytics engineering work will require you to join tables together to create robust, wide tables that will eventually be exposed to end business users. These models will usually be created using mostly [left](/sql-reference/left) (and some [inner](/sql-reference/inner-join)) joins.
```
KiraFuruichi:
```suggestion
A truly rarely seen, but important join: the cross join. The majority of your analytics engineering work will require you to join tables together to create robust, wide tables that will eventually be exposed to end business users. These models will usually be created using mostly [left](/sql-reference/left-join) (and some [inner](/sql-reference/inner-join)) joins.
```

website/docs/sql-reference/joins/sql-self-join.md
```diff
@@ -0,0 +1,71 @@
...
+
+```
+select
+	<fields>
+from <table_1> as t1
+[<join_type>] join <table_2> as t2
+on t1.id = t2.id
+```
+
+Since you can choose the dialect of join for a self join, you can specify if you want to do a [left](/sql-reference/left), [outer](/sql-reference/outer-join), [inner](/sql-reference/inner-join), [cross](/sql-reference/cross-join), or [right join](/sql-reference/right-join) in the join statement.
```
KiraFuruichi:
```suggestion
Since you can choose the dialect of join for a self join, you can specify if you want to do a [left](/sql-reference/left-join), [outer](/sql-reference/outer-join), [inner](/sql-reference/inner-join), [cross](/sql-reference/cross-join), or [right join](/sql-reference/right-join) in the join statement.
```

website/docs/sql-reference/statements/sql-case-statement.md
```diff
@@ -0,0 +1,75 @@
+---
+id: case
+title: SQL CASE
```
KiraFuruichi:
```suggestion
title: SQL CASE WHEN
```
is it ok to retitle this?

website/docs/sql-reference/statements/sql-case-statement.md
```diff
@@ -0,0 +1,75 @@
...
+slug: /sql-reference/case
+---
+
+<head>
+    <title>Working with the SQL CASE statements</title>
+</head>
+
+SQL case statements are the backbone of analytics engineers and dbt projects. They help add context to data, make fields more readable or usable, and allow you to create specified buckets with your data.
+
+To informally formalize it, case statements are the SQL equivalent of an if statement in other programming languages. They allow you to cascade through multiple scenarios (or cases) in your data, evaluate them if they’re true, and output a corresponding value for each case.
```
KiraFuruichi:
```suggestion
To informally formalize it, case statements are the SQL equivalent of an if-then statement in other programming languages. They allow you to cascade through multiple scenarios (or cases) in your data, evaluate them if they’re true, and output a corresponding value for each case.
```

website/docs/sql-reference/statements/sql-case-statement.md
```diff
@@ -0,0 +1,75 @@
...
+
+In this page, we’ll break down how to use SQL case statements and demonstrate why they’re valuable to modern data teams.
+
+## How to use the SQL case statements
+
+Case when statements are created in SELECT statements along with other fields you choose to select. The general syntax for SQL case when statements is as follows:
```
KiraFuruichi: link to SELECT page

website/docs/sql-reference/joins/sql-full-outer-join.md
```diff
@@ -0,0 +1,72 @@
...
+on t1.id = t2.id 
+```
+
+In this example above, there’s only one field being used to join the table together; if you’re joining between database objects that require multiple fields, you can leverage AND/OR operators, and more preferably, <Term id="surrogate-key">surrogate keys</Term>. You may additionally add [WHERE](/sql-reference/where), [GROUP BY](/sql-reference/group-by), [ORDER BY](/sql-reference/order-by), [HAVING](/sql-reference/having), and other clauses after your joins to create filtering, ordering, and performing aggregations.
+
+A note on full outer joins: it may sound obvious, but because full outer joins can return all rows between two tables, they therefore can return *many* rows, which is not necessarily a recipe for efficiency. When you use full outer joins, you often can find alternatives using different joins or unions to potentially bypass major inefficiencies caused by a full outer join.
+
+### SQL full outer join example
+
+Table A `car_type`
```
KiraFuruichi, line 34: these tables are not being formatted properly 
image

website/docs/sql-reference/joins/sql-cross-join.md
```diff
@@ -0,0 +1,74 @@
...
+    <fields>
+from <table_1> as t1
+cross join <table_2> as t2
+```
+
+Cross joins are one of those SQL concepts that is easier to understand with a tangible example, so let’s jump into it.
+
+### SQL cross join example
+
+Table A `date_spine`
```
KiraFuruichi, line 31: these tables are not being formatted properly

website/docs/sql-reference/joins/sql-left-join.md
```diff
@@ -0,0 +1,76 @@
...
+| 2 | sedan | null |
+| 3 | truck | green |
+
+Because there’s no `user_id` = 2 in Table B, there is no `color` available, thus a null result `color` column for `user_id` 2.
+
+## SQL left join use cases
+
+Left joins are a fundamental in data modeling and analytics engineering work—they allow you to easily join database objects onto each other while  maintaining an original table’s row count (in the from statement). Compared to right joins, that return all rows in a right join database object (and not the from statement), we find left joins a little more intuitive to understand and build off of.
+
+:::tip Ensure your joins are just ~left~ right
```
KiraFuruichi:
```suggestion
:::tip Ensure your joins are just ~right~
```

the strikethrough formatting isn't coming through, so if you can't fix, we can adjust the wording to this :)

website/docs/sql-reference/joins/sql-right-join.md
```diff
@@ -0,0 +1,70 @@
...
+from <table_1> as t1
+right join <table_2> as t2
+on t1.id = t2.id 
+```
+
+In this example above, there’s only one field from each table being used to join the two together together; if you’re joining between two database objects that require multiple fields, you can leverage AND/OR operators, and more preferably, <Term id="surrogate-key">surrogate keys</Term>. You may additionally add [WHERE](/sql-reference/where), [GROUP BY](/sql-reference/group-by), [ORDER BY](/sql-reference/order-by), [HAVING](/sql-reference/having), and other clauses after your joins to create filtering, ordering, and performing aggregations. You may also right (or any join really) as many joins as you’d like in an individual query or <Term id="cte" />.
+
+### SQL right join example
+
+Table A `car_type`
```
KiraFuruichi, line 32: These tables aren't formatting properly

website/docs/sql-reference/joins/sql-full-outer-join.md
```diff
@@ -0,0 +1,72 @@
+---
+id: outer-join
+title: SQL OUTER JOIN
+description: A SQL full outer join is a join between two tables that returns all rows from both tables, regardless of join key match success
```
KiraFuruichi:
```suggestion
description: A SQL full outer join is a join between two tables that returns all rows from both tables, regardless of join key match success.
```

website/docs/sql-reference/statements/sql-case-statement.md
```diff
@@ -0,0 +1,75 @@
...
+     else [fallback result] -- this else is optional
+end as <new_field_name>
+```
+
+Some notes on case statement functionality:
+- Scenarios in case statements are *evaluated in the order they’re listed*. What does this mean? It means that if multiple scenarios evaluate to true, the earliest listed true scenario is the one whose result is returned.
+- The results in each scenario need to be of the same data type; if scenario 1 results in a string, all other scenarios need to be [strings](/sql-reference/strings).
+- Oftentimes data teams will omit a final `else` scenario since the `else [fallback result]`is optional and defaulted to `else null`.
+- In general, case statement performance in select statements is relatively efficient (compared to other SQL functionality like aggregates or clunky joins involving ANDs and ORs); this isn’t to say it’s efficient (or smart) to be comparing a ton of scenarios, but it likely won’t be the bottleneck in your data models.
+- Case when statement results can also be passed into aggregate functions, such as MAX, MIN, and COUNT, or even date functions (ex. `date_trunc('month', <case when statement>`)
```
KiraFuruichi:
```suggestion
- Case when statement results can also be passed into aggregate functions, such as MAX, MIN, and COUNT, or even date functions (ex. `date_trunc('month', <case when statement>`)
```

we should link to the MAX, MIN, and COUNT pages

KiraFuruichi (changes requested):
Fix table formatting on join pages

johnblust:
@KiraFuruichi Thanks for catching those issues! Should be good to go now

KiraFuruichi (approved):

PR #2751, opened 2023-01-25
Run create_next_pr workflow only in the dbt-labs org

flexponsive:
When forking the `dbt-labs/docs.getdbt.com` repo, typically I only want to fork from the main branch `current`. However, with the current configuration, I get a failing workflow from the `create_next_pr` workflow, which tries to create a pull request *in the forked repo* to a branch which does not exist there:

image

This workflow is presumably intended for use in the primary repo but not meaningful in forks. With this PR, I suggest to change the configuration to run this job only in the `dbt-labs/docs.getdbt.com` repo.

## Checklist

N/A

Caveat: I tested that this causes the github action error to no longer arise in the forked repo, but I could not test due to permissions that it continues to run in the dbt-labs org :)


runleonarun:
Hey @flexponsive thanks so much for reporting this! Because we don't usually fork the repos internally, we'd never had known about this failing test! ❤️ 

@JKarlavige I think this looks good. Do you see any issues? The only downside might be someone forking next and not being able to get current merged in, but that's not terrible. I wasn't quite sure how to test it... would we need to create a PR off of next in the PR?

JKarlavige:
Thanks for flagging this @flexponsive! I don't see an issues with merging this in. We currently have this action disabled in the primary repo, but i'm assuming it still runs for forked repos. We can always adjust again if we need to activate this action again in the future

JKarlavige (approved):
lgtm!

flexponsive:
Thanks a lot! After syncing my fork, the workflow is now marked as skipped in the fork as it should be:
image

PR #2670, opened 2023-01-06
remove api v4 docs page since its not really supported

Topherhindman:
## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) and [About versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) so my content adheres to these guidelines.
- [ ] Add a checklist item for anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

Adding new pages (delete if not applicable):
- [ ] Add page to `website/sidebars.js`
- [ ] Provide a unique filename for the new page

Removing or renaming existing pages (delete if not applicable):
- [ ] Remove page from `website/sidebars.js`
- [ ] Add an entry `_redirects`
- [ ] [Ran link testing](https://github.com/dbt-labs/docs.getdbt.com#running-the-cypress-tests-locally) to update the links that point to the deleted page

PR #2776, opened 2023-01-27
Reverting 2670: Re-adding api v4 docs page until deprecation

runleonarun:
Reverts dbt-labs/docs.getdbt.com#2670

We're updating messaging for these docs and their deprecation.


nghi-ly (approved):
🚢

PR #2783, opened 2023-01-30
update package-lock to use latest version of ua-parser-js

JKarlavige:

Dependabot reported a vulnerability for the `ua-parser-js` package for versions less than 0.7.33. This updates our package-lock, which updates `ua-parser-js` to 0.7.33 to clear the vulnerability.

image

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -66,9 +66,9 @@
       }
     },
     "node_modules/@adobe/css-tools": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.0.1.tgz",
-      "integrity": "sha512-+u76oB43nOHrF4DDWRLWDCtci7f3QJoEBigemIdIeTi1ODqjx6Tad9NCVnPRwewWlKkVab5PlK8DCtPTyX7S8g==",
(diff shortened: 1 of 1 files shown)


lfracalossiFA:
@JKarlavige Do we also need to change this in package.json? I believe we can only specify starred version dependencies in package.json, though, right? (eg. 0.7.*) So maybe not...

JKarlavige:
> @JKarlavige Do we also need to change this in package.json? I believe we can only specify starred version dependencies in package.json, though, right? (eg. 0.7.*) So maybe not...

This package is used internally by Docusaurus and not on our end within `package.json`. The `ua-parser-js` package is used by the `fbjs` package, which is internally used by Docusaurus. `fbjs` has `ua-parser-js` set to `^0.7.30`, which will update it to the latest compatible version. So when the package-lock is regenerated, it upgrades `ua-parser-js` to the latest version, which is currently `0.7.33`.

Basically the last time our package-lock was generated, `ua-parser-js` latest version was less than `0.7.33`. They've since made an update, i'm guessing to fix the specific vulnerability that popped up

image

lfracalossiFA:
Ah gotcha. Carry on!

lfracalossiFA (approved):

PR #2803, opened 2023-02-02
Correct row count figure in postgres-configs.md

james-mead:
Should this read `10m` for 10 million?

The docs quote a row count of `10mm` which, I believe, should instead say `10m` as a representation of 10 million.


mirnawong1 (approved):
hey, @james-mead great catch and you're absolutely right! thank you for opening this up, I've merged this for you ✨

PR #2833, opened 2023-02-07
Fix links

zesuma:
## Fix pages that contain broken links

#Asana task
https://app.asana.com/0/1200099998847559/1202377085928734/f

#Description of the Issue:
According to Ahrefs we are linking to broken links on a few of our pages. See the doc here with the page, existing link, and the link it should go to instead: https://docs.google.com/spreadsheets/d/1FSWpwi16wMoD43hBGyx9-oLKoUcsr_lHzNuePWZ32sQ/edit?usp=sharing

Links:
https://docs.google.com/spreadsheets/d/1FSWpwi16wMoD43hBGyx9-oLKoUcsr_lHzNuePWZ32sQ/edit?usp=sharing


JKarlavige (approved):
lgtm! I'm not too concerned overall with these broken links. The docs team recently adjusted Docusaurus to fail the build if broken links are found (Docusaurus checks broken links during the build). These links shouldn't exist on the site anymore, so it makes sense to only add the redirects here.

:shipit:

PR #2838, opened 2023-02-09
Update --output-keys example in list.md

AF414:
Having tried it out initially, then looking through the dbt core code, it appears that `--output-keys` is expecting a list of comma separated strings, no spaces. Input of the form `"name","description"` / `name,"description"` / `"name",description` also work, but I'm not sure including a quoted example is necessary because (as far as I can tell) there are not any keys containing spaces.

I am fixing the documented example of `--output-keys` arguments so others can see what input is expected.
## Checklist


website/docs/reference/commands/list.md
```diff
@@ -72,7 +72,7 @@ $ dbt ls --select snowplow.* --output json
 
 **Listing JSON output with custom keys**
 ```
-$ dbt ls --select snowplow.* --output json --output-keys name description
+$ dbt ls --select snowplow.* --output json --output-keys name,description
```
mirnawong1: correcting this per @jtcohen6 's suggestion, however, it will eventually work in 1.5 _without_ quotes

```suggestion
$ dbt ls --select snowplow.* --output json --output-keys "name resource_type description"

```

mirnawong1 (approved):
hey @AF414 thanks for opening this up and really appreciate your contribution! there's a great discussion going on [here](https://github.com/dbt-labs/dbt-core/issues/6676) on how this should work (hopefully when 1.5 launches). in the meantime, this does work if you quote the whole thing (still including spaces). I've added versioning blocks so you can see the difference when switching versions in the navigation header.

looking forward to your future contributions ✨

PR #2873, opened 2023-02-15
Add a line to fix date type

ernestoongaro:
A user tried this code and was wondering why we converted their dates to unix timestamps. Turns out this is known thing: https://community.snowflake.com/s/article/Snowpark-converts-Pandas-DateTime-type-to-LongType


website/docs/docs/build/python-models.md
```diff
@@ -345,6 +345,7 @@ def model(dbt, session):
     # apply our function
     # (columns need to be in uppercase on Snowpark)
     df["IS_HOLIDAY"] = df["ORDER_DATE"].apply(is_holiday)
+    df["ORDER_DATE"].dt.tz_localize('UTC') #change date type
```
dbeatty10: Wonderful fix @ernestoongaro!

Based on [these](https://pandas.pydata.org/docs/reference/api/pandas.Series.dt.tz_localize.html) docs, what do you think about this comment?
```suggestion
    df["ORDER_DATE"].dt.tz_localize('UTC') # convert from Number/Long to tz-aware Datetime
```

mirnawong1 (approved):
thanks @ernestoongaro and @dbeatty10 - approving this for the merge ✨

PR #2898, opened 2023-02-20
Add database and schema information.

EvidenceN:

I am proposing adding DB and schema information to the sample .yml code above. 

The sample code for .yml file only shows table, but then below that, the documentation says the .yml code above will compile into raw.jaffle_shop.orders, but the database "raw" and schema "jaffle_shop" wasn't specified above in the .yml code. This will create confusion because someone will wonder how DBT will compile code into fully qualified DB.Schema.Table without us providing the DB and Schema in the source file.


website/docs/docs/build/sources.md
```diff
@@ -38,6 +40,8 @@ sources:
 
 </File>
 
+*By default the `schema` will be the same as `name`. Add `schema` only if you want to use a source name that differs to existing schema.
```
mirnawong1:
```suggestion
*By default, `schema` will be the same as `name`. Add `schema` only if you want to use a source name that differs from the existing schema.
```

mirnawong1 (approved):
hey @EvidenceN thanks for opening this up and great contribution! I've made some small tweaks to clarify the requirements and optional fields.  I've gone ahead and merged this for you, looking forward to your future contributions ✨

PR #2915, opened 2023-02-22
Update/beta menu enhancement

breezyfasano:
[Please see the associated Asana ticket here.](https://app.asana.com/0/1200099998847559/1203996144172203/f)

At the time the navbars for mobile and desktop are generated, this code checks if the version linked in the navbar has the `isPrerelease: true` flag. If it is true, it will append `(Beta)` to the end of the menu item indicating a beta release.

[To preview, view this link.](https://deploy-preview-2915--docs-getdbt-com.netlify.app)
1. Check out the desktop navbar versioning dropdown for 1.5, and assure it has "(Beta) attached to it.
2. Follow the same steps but for the mobile version of the site using the web inspector.
3. (Optional) You can pull down the branch and add `isPrerelease: true` to any versions in the `dbt-versions.json` file to see the functionality work for other versions for testing purposes.


john-rock (dismissed):
lgtm :shipit:

JKarlavige (approved):
:shipit: :shipit: :shipit:

matthewshaver:
@breezyfasano This is perfect and exactly what we needed. In the Netlify preview I no longer see the beta banner being presented on those pages. Will that still be present too? We'd like to have that banner and the flag in the dropdown versions menu both be displayed

matthewshaver (dismissed):

JKarlavige (changes requested):
Nice catch on that issue @matthewshaver! @breezyfasano I think I see what's occurring here. The `updateVersion` method is called on click on lines `117` and `200`, and passes the element into that method.

The `updateVersion` method lives in `website/src/stores/VersionContext.js` [here](https://github.com/dbt-labs/docs.getdbt.com/blob/b0bf3fb671ae94d7e59f88f332577983bb087e26/website/src/stores/VersionContext.js#L50). It ultimately gets the text from that element, strips the `v` from `v1.5` for example and setting the `versionValue` to `1.5`.

However, now that we are adding `(Beta)` for prerelease versions, `updateVersion` is still only stripping the `v` from the text, so the `versionValue` becomes `1.5 (Beta)`, which doesn't match any versions in the `dbt-versions.js` file.

So we would have to update the `updateVersion` method to ensure it also strips the `(Beta)` when it's setting the `versionValue` variable.

Let me know if you have any questions and we can sync up!

breezyfasano:
> Nice catch on that issue @matthewshaver! @breezyfasano I think I see what's occurring here. The `updateVersion` method is called on click on lines `117` and `200`, and passes the element into that method.
> 
> The `updateVersion` method lives in `website/src/stores/VersionContext.js` [here](https://github.com/dbt-labs/docs.getdbt.com/blob/b0bf3fb671ae94d7e59f88f332577983bb087e26/website/src/stores/VersionContext.js#L50). It ultimately gets the text from that element, strips the `v` from `v1.5` for example and setting the `versionValue` to `1.5`.
> 
> However, now that we are adding `(Beta)` for prerelease versions, `updateVersion` is still only stripping the `v` from the text, so the `versionValue` becomes `1.5 (Beta)`, which doesn't match any versions in the `dbt-versions.js` file.
> 
> So we would have to update the `updateVersion` method to ensure it also strips the `(Beta)` when it's setting the `versionValue` variable.
> 
> Let me know if you have any questions and we can sync up!

Aaaah thank you SO much Jason, this is exactly what I was looking for! I was figuring something somewhere was reading the menu label to match data, but just couldn't find it. Much appreciated!! I'll let you know if I need a zoom sync. :)

breezyfasano:
@JKarlavige I applied a fix and it looks like that worked beautifully! Thanks again.
Can you re-review when you get a chance?

website/src/stores/VersionContext.js
```diff
@@ -46,8 +46,10 @@ export const VersionContextProvider = ({ value = "", children }) => {
   const updateVersion = (e) => {
     if(!e.target)
       return
-    
-    const versionValue = e.target.text.replace('v', '')
+      
+    const vRegex = /(?:v)?(\d+(\.\d+)*)/ // Regex that will parse out the version number, even if there is/isn't a 'v' in front of version number and a '(Beta)' afterwards.
+    const versionValue = e.target.text.match(vRegex)[1]
```
JKarlavige, line 51: This is muchhhh more efficient than what was originally there, nice work!

JKarlavige (approved):
lgtm! Verified the prerelease banner shows for 1.5, and that the version persists after refreshing the page :shipit:

website/src/stores/VersionContext.js
breezyfasano: thank you so much!

PR #2924, opened 2023-02-24
cypress component test  discourse

wendyvasq:
# Code Task Complete ⚡️

✨ View the Asana Task [here](https://app.asana.com/0/1200099998847559/1203913475407624/f) ✨

## Description of Changes

(A brief idea of what reviewers should be looking for, or the expected behavior of the new update.)

This is the component test for `discourse` .
Note:
1. that this PR includes webpack configuration changes that block `discouseBlogComments` component test.  
2. Block PR https://github.com/dbt-labs/docs.getdbt.com/pull/2925

## Preview Links

https://deploy-preview-2924--getdbt.netlify.app/

## Additional Info

(Any additional comments here, if any)


website/src/components/discourse/discourse.cy.js
```diff
@@ -0,0 +1,76 @@
...
+    })
+
+    it('Should display Discourse data after API fetch', () => {
+        cy.intercept('POST', 'http://localhost:8888/.netlify/functions/get-discourse-topics', (req) => {
+            req.reply(mockResponseData);
+        }).as('getTopics')
+
+        cy.mount(<DiscourseFeed />)
+        cy.wait('@getTopics')
+        cy.get('[data-testid="topics-list"]').as('topicsList')
```
JKarlavige: This is really cool, didn't know about the `.as()` method!

website/src/components/discourse/discourse.cy.js
```diff
@@ -0,0 +1,76 @@
...
+describe('Test Discourse component', () => {
+    it('Should render same text passed into title prop', () => {
+        cy.mount(<DiscourseFeed title='Open topics' />)
+        cy.contains('Open topics').should('be.visible')
+    })
+
+    it('Should display loading icon on inital load', () => {
+        cy.intercept('POST', 'http://localhost:8888/.netlify/functions/get-discourse-topics', {
+            delayMs: 10,
+        }
```
JKarlavige: Question on the url here, does `cy.intercept` actually hit the endpoint here? Or is it just an example request? Wondering if it needs to be adjusted for when live. If it's just an example then i don't see a need to update it!

JKarlavige (commented):
This is AWESOME, nice work @wendyvasq! Definitely the most thorough test we have so far. Just have one question regarding the url in the `.intercept` method. Once that's cleared up i'm ready to call this g2g!

website/src/components/discourse/discourse.cy.js
wendyvasq: I know right I found it reading an article.

website/src/components/discourse/discourse.cy.js
wendyvasq: oh yea, that is a good call out. I don't know if leaving the url as is will be a breaking change when the test are in production. I would need to do some research on that.  However,  I think I read somewhere I can use regex to make that more resilient to different url environments. Cypress intercept does like urlPatternMatch function.

website/src/components/discourse/discourse.cy.js
john-rock: Yeah `.as()` look like it will be very convenient!

john-rock (approved):
Nice work, lgtm!

JKarlavige (approved):
:shipit:

PR #2961, opened 2023-03-08
Fix slugs on materializations guide

gwenwindflower:
The `slugs` YAML frontmatter property was misused and resulting in double-length urls in production on the materializations guide. This will fix those and redirect the landing page for the guide at the double-length URL to the fixed URL.


jasnonaz:
Thanks for fixing this! One non-blocking comment - I believe the docs team has moved from having the top level sidebar items contain content towards having each section explicitly in the child section. Would we consider moving from what we have now:
image

To instead have a line item for "materialization best practices overview" or just "overview" as the first item underneath "Materialization best practices" and above "Available materializations"?

jasnonaz (approved):
😎

gwenwindflower:
we should do that for all the guides in a separate PR 👍🏻 , i thought about that too

PR #2964, opened 2023-03-08
Implement ESLint

JKarlavige:
## What are you changing in this pull request and why?

[Asana task](https://app.asana.com/0/1200099998847559/1203760149887729/f)

This introduces ESLint to automate formatting for markdown and JS files. 

There are two areas where ESLint runs:
- Locally when making a commit, this checks only changed files
- In a Git Action which runs on every push to a PR. This checks all files.

## Testing

To test, open a file and add improperly formatted code. In this example, I added an undefined variable to `website/src/components/author/index.js`.
image

Notice the variable now shows an error within VSCode.

Now i'll try to make a commit. The commit fails due to the `error`: `'breakstuff is not defined`. This has prevented me from making a commit due to the error. The warnings show here due to the error, however ESLint is currently configured to allow `warnings` through in commits.
image

## Notes

There is a separate PR (#2969) where the existing ESLint issues are fixed. I will merge that PR into here when approved.

Here is prior ESLint Git Action run where `errors` were found: https://github.com/dbt-labs/docs.getdbt.com/actions/runs/4376123948/jobs/7657782506

## Checklist
- [x] Setup ESLint with Husky
- [x] Setup an ESLint check within PRs

PR #2969, opened 2023-03-09
Fix eslint issues

JKarlavige:

This fixes the ESLint issues found with the new implementation in #2964.

## Notes

This only fixes the `errors` found with ESLint, and not the `warnings`. We have the ability to limit the amount of warnings we allow if requested, although they are not vital to fix immediately.

Here is the Git Action with the latest ESLint check for this PR: https://github.com/dbt-labs/docs.getdbt.com/actions/runs/4376390981/jobs/7658431509

## Merging

When approved, I have this set to merge into the `add-eslint` branch where the actual eslint implementation is located (#2964). Then once #2964 is approved, I will merge the eslint setup live.


john-rock (approved):
This lgtm. I'll take another look ath the implement eslint PR once this is merged into it.

PR #2975, opened 2023-03-10
Fix dbt_project.yml example on hooks

vitorbaptista:
The `dbt_profiles.yml` example on how to setup hooks incorrectly uses the  `model` key, when it should be `models` (notice the extra `s`).

I also searched the whole docs for `model:` and I could only find the instances in this PR where it should be changed to `models:`


mirnawong1 (approved):
hey @vitorbaptista thanks for opening this up and great catch! I've approved + merged this for you, looking forward to future contributions ✨

PR #3012, opened 2023-03-16
Update/homepage improvements

breezyfasano:
Changes as outlined in the [Asana task](https://app.asana.com/0/1200099998847559/1203986472821200/f)

- [x]  When announcement bar is not present, make sure diamonds in hero do not overlap the header
- [x]  Make the spacing between the rows of cards is an even 2rem across the board  
- [x]  Card fonts - font size is 20, line height is 30
    - [x]  Also for “Latest from Developer Blog”
- [x]  Icon width is 25px
- [x]  50px padding on left and right of entire page.
- [x]  Latest from Dev Blog Section
    - [x]  Reduce space between each card - 30px
    - [x] Transparent gradient on sides of swiper container


website/src/css/custom.css
```diff
@@ -1722,8 +1722,9 @@ section > h2:not(.resource-section) {
   padding-right: 3rem;
   padding-left: 3rem;
   padding-top: 0rem;
-  padding-bottom: 6.25rem;
-  margin-bottom: 0;
+  padding-bottom: 0;
+  margin-bottom: 6.25rem;
+  box-shadow: inset 30px 0 30px -30px rgba(0, 0, 0, 0.3), inset -30px 0 30px -30px rgba(0, 0, 0, 0.3);
```
wendyvasq: I never seen this before so I am not that confident if this is the right  for css.   
The box shadowing works when I tested it. Welp,  I learned something new today.

wendyvasq (dismissed):
lgtm !  I left a comment on a line of css for box shadowing that I am not too sure about.

website/src/css/custom.css
john-rock: On light mode it looks like that gradient is a bit harsh at the edge. Maybe we should smooth that out a bit.

dbt_Developer_Hub

john-rock (changes requested):
All looks good just had one callout on the Swiper gradient.

edit: Just realized Emily had called that out on the task.

lfracalossiFA:
Hey @breezyfasano , question -- I saw this in your notes:

> Card fonts - font size is 20, line height is 30
> Also for “Latest from Developer Blog”

"Latest from Developer Blog" looks to be 25px/41px. Is this correct?

ETA: oh, I'm thinking you mean the h3s inside the card, correct?

website/src/css/custom.css
lfracalossiFA: @wendyvasq For informational purposes... there are some great tools out there that will take inputs from sliders and give you the `box-shadow` values. This is the one I'm familiar with: https://cssgenerator.org/box-shadow-css-generator.html

website/src/components/hero/styles.module.css
```diff
@@ -29,7 +29,7 @@
   .showGraphic{
     position: absolute;
     left: -50px;
-    top:150px;
+    top:125px;
```
lfracalossiFA, line 32: Would it make sense to make this a `vh` value? It might scale a bit better on different screen sizes. I haven't tested this, though, so just a theory. Feel free to ignore me if not relevant :)

website/src/components/postCarousel/index.js
```diff
@@ -1,27 +1,29 @@
 import React from 'react';
 import BlogPostCard from '@site/src/components/blogPostCard';
 import { Swiper, SwiperSlide } from 'swiper/react';
-import { Navigation } from 'swiper';
 import 'swiper/css';
+import { Navigation } from 'swiper';
```
lfracalossiFA, line 5: For my own edification: any reason for moving this import down below the CSS?

website/src/components/postCarousel/index.js
```diff
@@ -1,27 +1,29 @@
...
 import 'swiper/css/navigation';
 
 
 function PostCarousel({ blogPostData }) {
   return (
     <div>
       <Swiper
-        spaceBetween={60}
+        spaceBetween={30}
```
lfracalossiFA, line 13: Are these values pixels?

website/src/css/custom.css
```diff
@@ -1722,8 +1722,9 @@ section > h2:not(.resource-section) {
...
 }
 
 .home .swiper .swiper-button-next, .home .swiper .swiper-button-prev {
```
lfracalossiFA: Curious why GitHub is showing me this line but not the rest of the file, when it looks like there were no more changes after this point... 🤔

lfracalossiFA:
This seems really narrow on mobile (emulating iPhone XR). I wonder if we could reduce the padding/margin at this breakpoint to make it less narrow, or don't show the borders of adjacent boxes. @emilymermell-dbt do you think this is worth fixing?

Screenshot 2023-03-17 at 9 49 51 AM

website/src/css/custom.css
JKarlavige: Yeah that's a good callout @lfracalossiFA, i've always used tools to generate the css for box shadows. Maybe @breezyfasano is a wizard and can whip these together by hand, but i definitely cannot 😅 

Along with using tools, you can also get the box-shadow css from Figma mockups. If you click an element in Figma that has a box-shadow, then click `Inspect` in the upper right, it gives you all the css for that element, including the box-shadow!
image

lfracalossiFA:
And super-duper narrow on Galaxy S8+ (admittedly I'm not an Android user and I have no idea if anyone uses these any more, lol).

Screenshot 2023-03-17 at 9 52 38 AM

website/src/css/custom.css
lfracalossiFA: Also I'm seeing the same think in dark mode:

Screenshot 2023-03-17 at 9 58 34 AM

I noticed you said you added a transparent gradient on the Swiper carousel, so I'm wondering if this is intentional? I'm not sure I like the look of it, though.

lfracalossiFA (changes requested):
A few small things that I think should be changed! Would like Emily's read on the size of the carousel cards on mobile and if it should be fixed.

website/src/css/custom.css
lfracalossiFA: @JKarlavige Oh yeah, I was going to mention getting the style from the Figma comp, but then I realized there wasn't an actual Figma comp for this (I don't think?)

website/src/components/postCarousel/index.js
breezyfasano: There was no particular reason for this that I had in mind, to be honest!

website/src/components/postCarousel/index.js
breezyfasano: Sure are!

breezyfasano:
Team - I went ahead and removed the shadow per Emily's request on the Asana thread!
Still adjusting working on the Swiper card widths on mobile per @lfracalossiFA's feedback.
Thanks for your comments, all!

website/src/components/postCarousel/index.js
breezyfasano: I think I had tried to use a piece of the Swiper package for the shadow, but it didn't work out - so I reverted everything back. But it looked like that shook out to looking like I moved this down one line for no reason 😆

breezyfasano:
@lfracalossiFA and @john-rock - re-requesting a review on this with some changes!
1. I've removed the gradient from the carousel re: Emily's comment in Asana
2. I've also adjusted the carousel on mobile to reflect a less scrunched view!

Let me know what y'all think & if this addressed your changes.
https://deploy-preview-3012--docs-getdbt-com.netlify.app/

lfracalossiFA (approved):
lgtm, but I do see eslint-check throwing up a bunch of warnings within the Files Changed tab. I don't know if those are super important, though.

JKarlavige:
@lfracalossiFA that's okay, the warnings have been in there. We just setup ESLint for PRs, and only sorted out the blocking errors so far, but could always go back and clean up the warnings at some point

PR #3082, opened 2023-03-27
Update cloud-feature-parity.md

will-sargent-dbtlabs:
Updated Cloud Feature Partity Matrix entry "Webhooks" to include the word "(Outbound)" detail and added and entry for "Slim CI".

This comes from customer comments and questions that because their Azure ST didn't have "webhooks" that they couldn't use Slim CI (not true), since technically, it's a type of webhook (their deduction).

Adding that clarification and the second line allows us to be clear about both topics, as this was suddenly a source of major concerns about dbt Cloud adoption for one of our top 5 largest accounts that is just starting to onboard.


mirnawong1 (approved):
hey @will-sargent-dbtlabs thanks for opening this pr up to clarify that -we absolutely want to make sure the features are super clear to users and we appreciate you flagging this! approved this for you to merge!

cc @matthewshaver FYI - I've only glanced at this but do we need to add more features to this matrix?

Issue #3099, opened 2023-03-28
Update multi-project docs for v1.5.0-rc1 (April 13)


jtcohen6:
### Contributions

- [X] I have read the contribution docs, and understand what's expected of me.

### Link to the page on docs.getdbt.com requiring updates

Contracts:
- https://docs.getdbt.com/reference/resource-configs/contract
- https://docs.getdbt.com/reference/resource-properties/constraints
- https://docs.getdbt.com/reference/node-selection/methods#the-state-method

Versions:
- https://docs.getdbt.com/docs/collaborate/publish/model-versions
- https://docs.getdbt.com/reference/resource-configs/alias
- https://docs.getdbt.com/docs/build/custom-aliases#generate_alias_name
- https://docs.getdbt.com/reference/dbt-jinja-functions/ref
- https://docs.getdbt.com/reference/node-selection/methods

### What part(s) of the page would you like to see updated?

Contracts:
- model-level `constraints`
- nicer warning messages for unenforced/unsupported constraints (per data platform)
- nicer error messages for contract mismatch & breaking changes
- ~detect breaking changes to model contracts during `state:modified` for constraints~ (kicked to v1.6)

Versions:
- model `versions` spec
- default alias behavior
- `v`/`version` keyword argument for `ref`
- default `latest_version` behavior (numerically greatest or alphabetically last)
- `version:` selection method (`none`, `latest`, `prerelease`, `old`)

### Additional information

Release:
- https://github.com/dbt-labs/dbt-core/issues/7235

Contracts:
- https://github.com/dbt-labs/dbt-core/issues/6754
- https://github.com/dbt-labs/dbt-core/issues/7067
- https://github.com/dbt-labs/dbt-core/issues/7209
- https://github.com/dbt-labs/dbt-core/issues/7332
- ~https://github.com/dbt-labs/dbt-core/issues/7065~ (kicked to v1.6)

Versions:
- https://github.com/dbt-labs/dbt-core/issues/6866
- https://github.com/dbt-labs/dbt-core/issues/6871
- https://github.com/dbt-labs/dbt-core/issues/6872
- https://github.com/dbt-labs/dbt-core/issues/7263
- https://github.com/dbt-labs/dbt-core/issues/7199

https://www.notion.so/dbtlabs/dbt-Core-v1-5-docs-a2ae11be77bf47629f85faf7a3f27991

PR #3101, opened 2023-03-29
model versions

MichelleArk:
Addresses: https://github.com/dbt-labs/docs.getdbt.com/issues/3099

## Checklist

- [ ] Add an entry `_redirects`


website/docs/docs/collaborate/publish/model-versions.md
```diff
@@ -15,17 +15,71 @@ For more details and to leave your feedback, check out the GitHub discussion:
...
+      - name: customer_id
+        description: This is the primary key
+        data_type: float
+      - name: country_name
+        description: Where this customer lives
+        data_type: string
+    versions:
+      - v: 2
+        columns:
+          - include: *
```
dave-connors-3: @MichelleArk is this intended to be an array? or a dictionary with `include` and `exclude` keys (among possible others)?

MichelleArk: Currently we're thinking it'll be an array element of the `columns` array. So you could do something like: 

```yaml
columns:
  - include: "*"
    exclude: column_name_to_exclude
  - name: customer_id  # included in addition the '*' set. if customer_id were in the '*' set -> override it
    description: This is the primary key
    data_type: float
  - name: country_name # included in addition the '*' set. if country_name were in the '*' set -> override it
    description: Where this customer lives
    data_type: string
```

website/docs/docs/collaborate/publish/model-versions.md
```diff
@@ -15,17 +15,71 @@ For more details and to leave your feedback, check out the GitHub discussion:
...
+        description: This is the primary key
+        data_type: int
+      - name: country_name
+        description: Where this customer lives
+        data_type: string
+```
+
+</File>
+
+If you wanted to make a breaking change to the model - for example, removing a column - you'd create a new model file (e.g. SQL file) encompassing those breaking changes. The default convention is to name the new file would with a `_v<version>` suffix. The new version can then be configured in relation to the original model:
```
matthewshaver:
```suggestion
If you wanted to make a breaking change to the model - for example, removing a column - you'd create a new model file (for example, SQL file) encompassing those breaking changes. The default convention is naming the new file with a `_v<version>` suffix. The new version can then be configured in relation to the original model:
```

website/docs/docs/collaborate/publish/model-versions.md
```diff
@@ -15,17 +15,71 @@ For more details and to leave your feedback, check out the GitHub discussion:
...
+          - include: *
+            exclude: country_name # this is the breaking change!
+      - v: 1
+        config:
+          alias: dim_customers # keep old relation location
+```
+
+</File>
+
+The above configuration will create two models (one for each version), with aliases `dim_customers` (version 1) and `dim_customers_v2`.
```
jtcohen6: We'll definitely want to call out that projects which have historically reimplemented `generate_alias_name` will need to update their custom implementations to account for model versions. Otherwise they'll see something like this as soon as they start using versions:
```
dbt.exceptions.AmbiguousAliasError: Compilation Error
  dbt found two resources with the database representation "database.schema.model_name".
  dbt cannot create two resources with identical database representations. To fix this,
  change the configuration of one of these resources:
  - model.project_name.model_name.v1 (models/.../model_name.sql)
  - model.project_name.model_name.v2 (models/.../model_name_v2.sql)
```
I'm not sure if it makes more sense here or in the migration guide (maybe both?)

website/docs/reference/model-properties.md
```diff
@@ -32,6 +33,25 @@ models:
         [tags](resource-configs/tags): [<string>]
 
       - name: ... # declare properties of additional columns
+    versions:
+      - v: <version_name> # required
```
jtcohen6: - Clarify that this can be string or numeric
- Somewhere we say that we'll do ordering accordingly (but we don't have any features yet that will actually use it - that will come in https://github.com/dbt-labs/dbt-core/issues/7199)

MichelleArk: > but we don't have any features yet that will actually use it 

we do! calling `ref` on a versioned model without the `version` argument relies on the latest model version and ordering implementation (if `latest_version` isn't explicitly configured)

jtcohen6:
@MichelleArk Thank you for the awesome start on this!!

I'm going to merge into #3171 so we can coordinate changes / avoid conflicts, and have a single PR for @matthewshaver to provide editorial review.

PR #3106, opened 2023-03-29
Updates to Vertica content based on core 1.4.0

nrodriguezmicrofocus:
Updates to Vertica content based on core 1.4.0


nrodriguezmicrofocus:
@dataders Can you please help us with these updates to Vertica content? Thank you

website/docs/reference/warehouse-setups/vertica-setup.md
```diff
@@ -2,13 +2,13 @@
...
+  maintained_by: 'Vertica'
+  authors: 'Vertica (Former authors: Matthew Carter, Andy Regan, Andrew Hedengren)'
   github_repo: 'vertica/dbt-vertica'
   pypi_package: 'dbt-vertica'
-  min_core_version: 'v1.3.0'
-  cloud_support: Not Supported
-  min_supported_version: 'Vertica 10.0'
+  min_core_version: 'v1.4.0 and newer'
+  cloud_support: 'Not Supported'
+  min_supported_version: 'Vertica 12.0.0'
```
dbeatty10, line 11: I think the intent here is actually to specify how far back a user could theoretically go.

See [here](https://docs.getdbt.com/reference/warehouse-setups/exasol-setup) for an example.

Looking at the release history [here](https://pypi.org/project/dbt-vertica/#history), it seems like it would be more correct to say **v0.15.0 and newer**

dbeatty10 (approved):
I'm going to merge this as-is.

But I left a comment about the minimum supported versions that could be handled in a follow-up PR.

PR #3114, opened 2023-03-30
Add ability to version control folders

john-rock:

[Asana task](https://app.asana.com/0/1200099998847559/1204169089676905/f)

This adds the ability to version control categories in the sidebar in the same way we currently version control pages.

Category versions are set the same way we currently set page versions inside `dbt-versions.js`. There is now a `versionedCategories` array that will render the categories in the sidebar depending on the `firstVersion` or `lastVersion` value. 

The `category-version-check.js` file was taken directly from the `page-version-check.js` and modified to account for categories instead.

## Example Usage
```
exports.versionedCategories = [
  {
    "category": "Publishing models",
    "firstVersion": "1.5",
  }
]
```

## Testing
In the deploy preview the `Publishing models` category is currently set to show only on version 1.5. Change to a lower version and the `Publishing models` category should no longer be shown.

https://deploy-preview-3114--docs-getdbt-com.netlify.app/docs/introduction

## Checklist


mirnawong1:
insert emoji :meow-wow: !!

website/src/theme/DocSidebarItem/index.js
```diff
@@ -12,18 +12,26 @@ import DocSidebarItemHtml from '@theme/DocSidebarItem/Html';
...
     if(!pageAvailable)
       return null
   }
 
+  // hide versionedCategories if they do not match the current version
+  if(version && versionedCategories) {
+    const { categoryAvailable } = categoryVersionCheck(version, versionedCategories, item.label)
+    if(!categoryAvailable)
+      return null
+  }
```
JKarlavige: Thinking we should add one additional check here to ensure the current `item` is of `type: category`. For example (although unlikely), if any document in the sidebar has the same label as the category, it would be hidden as well. 

I verified this is the case locally by updating a label for a random doc to the same name as the `Publishing models` category. If switching to v1.4 or lower, the doc would be hidden as well by the category version check.

Let me know if you have any concerns with adding the additional check if it could potentially effect anything else here.

image
image

JKarlavige (changes requested):
Nice work with this, all looks to be working correctly. Just left one comment with a small requested change. Let me know if you have any questions with that!

website/src/theme/DocSidebarItem/index.js
john-rock: Good point. I'll get get that check implemented!

john-rock: Added `if(version && versionedCategories && item.type === 'category')`

Let me know if you like that.

I was trying to think of a way to handle it the same as the pages within versionedPages (e.g. `docs/collaborate/publish/model-contracts`) but since Publishing models is labeled only within `sidebars.js` I ended up using the label instead.

@matthewshaver let me know if that works for you.

JKarlavige (approved):
lgtm! :shipit:

website/src/theme/DocSidebarItem/index.js
matthewshaver: This all looks great to me. Thank you @john-rock and @JKarlavige.

matthewshaver (approved):

PR #3119, opened 2023-03-31
Update example in tests.md

remigabillet:

The syntax used in the documentation's example for custom generic tests is incorrect.

Custom generic tests are declared with:

```
{% macro _(model, column_name) %}
   ...
{% endmacro %}
```

instead of:

```
{% test _(model, column_name) %}
   ...
{% endtest %}
```


dbeatty10:
Thanks for opening this PR @remigabillet !

Can you tell me more why you believe the syntax is incorrect?

I tried adding the following generic test definition within [this](https://github.com/dbeatty10/dbt-tests-demo) demo repo, and it worked perfectly for me.

`dbt-tests-demo/tests/generic/my_not_null.sql`
```
{% test not_null(model, column_name) %}
    select *
    from {{ model }}
    where {{ column_name }} is null
{% endtest %}
```

dbeatty10:
@remigabillet I'm going to close this since I believe the example for custom generic tests is correct.

But please let us know if you believe it is incorrect and we can take another look at it.

PR #3121, opened 2023-03-31
Add token field to oauth-secrets example

IvanKaramazov:
I just now beat my head against the wall for half a day trying to get dbt to connect BQ following the example given here for the `oauth-secrets` method. Eventually I looked in the dbt Python code and found that the credentials are being created using the `token` field, which is in the JSON I'm pulling from but not the example in these docs. I didn't trace through the code to figure out where dbt is getting the token value into the profile_credentials dict, but I took a leap of faith and supplied the token value from my JSON in profiles.yml. And it worked immediately! So if a user follows the updated example, it should be problem solved.

## Checklist

### DIFF
--- a/website/docs/reference/warehouse-setups/bigquery-setup.md
+++ b/website/docs/reference/warehouse-setups/bigquery-setup.md
@@ -111,7 +111,8 @@ my-bigquery-db:
       project: [GCP project id]
       dataset: [the name of your dbt dataset] # You can also use "schema" here
       threads: [1 or more]
-      refresh_token: [token]
+      token: [token]
+      refresh_token: [refresh_token]
       client_id: [client id]
       client_secret: [client secret]
       token_uri: [redirect URI]


dbeatty10:
@IvanKaramazov Sorry for your experience trying to get this to work and thank you for taking the time to suggest improvements to the docs to help others in the future!

Looking at the docs, there are two tabs under the ["Oauth Token-Based"](https://docs.getdbt.com/reference/warehouse-setups/bigquery-setup#oauth-token-based) heading:
1. Refresh token
2. Temporary token

Here's how each of them show up in the display which clicking on each tab:

image

image

I suspect you could only see the content of the first tab and it wasn't obvious to click on the second tab to see its content.

Maybe the best path forward would be to remove these two horizontal tabs and split them into two sections that are stacked on top of each other? The advantage is that both would be visible rather than requiring you to click on each tab.

Do you think that would have helped you?

matthewshaver:
Closing as this page has moved. Opened https://github.com/dbt-labs/docs.getdbt.com/pull/4302 to address and split up tabs to make them easier to discover.

PR #3137, opened 2023-04-04
Adding text for version optional

### LINKED ISSUE #3026: update `version` config docs
### Contributions

- [X] I am a PM or subject matter expert at dbt who is responsible for this feature.

### Where does this content belong?

As of 1.5:
- `version: ...` in `dbt_project.yml` is optional
- `config-version: 2` in `dbt_project.yml` is optional
- `version: 2` in dbt resource yml files is optional

We should update:
- https://docs.getdbt.com/reference/project-configs/version
- https://docs.getdbt.com/reference/project-configs/config-version

### Link to source material

- https://github.com/dbt-labs/dbt-core/pull/6718
- https://github.com/dbt-labs/dbt-core/pull/7060

matthewshaver:

Marking config version as optional for 1.5+

## Checklist


website/docs/reference/project-configs/config-version.md
```diff
@@ -2,6 +2,12 @@
 datatype: integer
 ---
 
+<VersionBlock firstVersion="1.5">
+
+Starting in dbt v1.5, the `config-version:` tag is optional.
```
mirnawong1, line 7: is there add'l doc we can link to (the 1.5 migration guide maybe)? or an example that we can display to show users what a successful  project.yml file should look like in v1.5?

also in dbt cloud, does this mean the config-version 2 tag (which automatically appears when you initialize, i think) won't appear or get added?

mirnawong1, line 7:
```suggestion
Starting in dbt v1.5, the `config-version:` tag is now optional and is no longer required. For projects on v1.4 or lower, select the documentation version in the header and refer to the listed specifications.

mirnawong1:
small content suggestion but mostly added a question about whether we should link to a doc for more details on the 1.5 change.

website/docs/reference/project-configs/config-version.md
matthewshaver: I'm not sure what this means for Cloud, we should find the right people to ask, though. This doesn't really change anything other than to make this version feature optional (previously if it wasn't included it would assume syntax was <v1 but those have all been deprecated)

matthewshaver: Is there another way we can word this? Nothing on the page will change if they switch versions other than the statement marking this as optional.

mirnawong1: oh i see, i think this is where i was confused - i assume the page contents changed and they would only see the '1.5 is optional' bit if they were on v1.5. and not see the other definitions, etc. it's a nonblocker for me - the main curiosity was the question about what it means for cloud and whether we should link to something else (migration guide) in the statement.

mirnawong1 (approved):
approving this for you @matthewshaver - just noting it may help users read more details if we link out to something that has more details on the config version change.

PR #3160, opened 2023-04-07
Add new custom DocCarousel React Component

wendyvasq:
# Code Task Complete ⚡️

✨ View the Asana Task [here](https://app.asana.com/0/1200099998847559/1203867647151127/f) ✨

## Description of Changes

(A brief idea of what reviewers should be looking for, or the expected behavior of the new update.)

To enhance the docs site, allow flexibility for writing docs, and improve the internal workflow of the docs team - requesting to add a scrolling carousel in docusaurus feature. similar to the docs homepage. 

The new React Component is called `DocCarousel`

Loom of component walk through 
https://www.loom.com/share/e9fb915ad99b4a6d966e618108cae25f

## Preview Links

https://deploy-preview-3160--docs-getdbt.netlify.app/

## Additional Info

Documentation for this component available [here](https://www.notion.so/dbtlabs/DocCarousel-Component-87776186f83b4a3f9d0933b890f0ed10?pvs=4)


website/src/components/docCarousel/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Swiper, SwiperSlide } from 'swiper/react';
+import 'swiper/css';
+import { Navigation } from 'swiper';
+import 'swiper/css/navigation';
+
+
+function DocCarousel({ children }) {
+    console.log(children)
```
JKarlavige: Small note to remove this console.log before merging.

website/src/components/docCarousel/index.js
```diff
@@ -0,0 +1,33 @@
...
+                    640: {
+                        slidesPerView: 2,
+                        spaceBetween: 30,
+                    },
+                }}
+            >
+                {children.map((item) => <SwiperSlide>{item}</SwiperSlide>)}
+            </Swiper>
+        </div>
+    );
```
JKarlavige: For this `children.map`, one potential scenario that could cause an error is if the DocCarousel component is added to a page but does not wrap any content, which would make `children` undefined.

image

We should add a check here to ensure the carousel has `children` and that the `children` array length is greater than zero before adding any swiper html. There's a couple ways you can do this, but in my opinion, I would do this check before any of the swiper html (starting at `<div className='docswiper'>`) is rendered.

Before the return statement on line 10, you can check if `children` does not exist. If children does not exist return `null`, otherwise, continue on to return the swiper html.

Let me know if you have any questions with this. The [cta component](https://github.com/dbt-labs/docs.getdbt.com/blob/71f9f0efc824152b7d9956c940513b2f3a7a4a07/website/src/components/cta/index.js#L11) in this repo has a good example on doing a check before rendering content.

JKarlavige (changes requested):
Tested the functionality and verified content was transformed into a carousel. Have one requested edit to add a check before attempting to render the swiper.

image

website/src/components/docCarousel/index.js
```diff
@@ -0,0 +1,35 @@
...
+import { Children } from 'react';
+import { Swiper, SwiperSlide } from 'swiper/react';
+import 'swiper/css';
+import { Navigation } from 'swiper';
+import 'swiper/css/navigation';
+
+function DocCarousel({ children }) {
+    if ( !Children.toArray(children).length > 0 ){      
+        return false
+    }
```
JKarlavige: Today i learned a `Children` class exists in React! It looks like it works as intended, although after [looking it up](https://react.dev/reference/react/Children), it looks like React does not recommend using this. It also notes that it's function is to: "manipulate and transform the JSX you received as the [children prop.](https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children)". Since we're not transforming the `children` prop, we won't need to use it here.

Instead, we can interact with the `children` prop directly. This allows you to shorten the check in your if statement. You can wipe out the `Children.toArray` part of the if statement, and simply check the `children` length.

And since there's potential for `children` to be `undefined`, I would also recommend using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) for the `.length` check. For example, if I had a statement: `if(exampleVar.length > 0)`. If `exampleVar` was `undefined`, this code would throw an error as JavaScript is unable to check the length of an undefined variable.

With optional chaining, the code is updated to: `if(exampleVar?.length > 0)`, with the question mark before `.length`. This says: "If `exampleVar` is `undefined`, this is false. Otherwise if `exampleVar` exists, continue to check if its length is > 0.

Before optional chaining, we would have to do two checks with this example to see if `exampleVar` exists, and then if its length was > 0. So it used to look like `if(exampleVar && exampleVar.length > 0)`. Optional chaining just allows us to shorthand that check using the question mark on `.length`.

That was a long-winded way of saying this is super close, I would just remove the `Chilren.toArray` and add an optional chain to the `.length`!

JKarlavige (changes requested):
Just have a small requested change and left a comment going over it!

website/src/components/docCarousel/index.js
wendyvasq: Thank you thats is a good call out ! I appreciate the input.

website/src/components/docCarousel/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Children } from 'react';
```
JKarlavige: @wendyvasq May you remove this `Children` import here to cleanup the eslint warning? Once that's taken out i'm ready to approve!

JKarlavige (approved):
lgtm! :shipit:

JKarlavige:
@mirnawong1, responding to your Asana questions here as it's easier to paste code. Added a `slidesPerView` prop which can be passed into the component to control how many columns the carousel displays on desktop. A max of 4 columns can be set, any number set higher than that will revert to 4 columns. Example below:

```
<DocCarousel slidesPerView={3}>

  <Card />
  <Card />
  <Card />
  <Card />
  <Card />
  <Card />

</DocCarousel>
```

And regarding having multiple rows per carousel slide, that is possible with a little html structuring. Each top-level element is converted into a slide. So if we wanted two cards stacked on top of each other per slide, we can wrap those two cards in a div like the example below:
```
<DocCarousel slidesPerView={2}>
  <div>
    <Card {...props} />
    <br />
    <Card {...props} />
  </div>

  <div>
    <Card {...props} />
    <br />
    <Card {...props} />
  </div>

  <div>
    <Card {...props} />
    <br />
    <Card {...props} />
  </div>
</DocCarousel>
```
Which would ultimately look like:
image

mirnawong1:
omg wow @JKarlavige ! you've just made my week! 🙏 this looks incredibly cool and great to know!

one non-blocking question, when i tried having multiple rows in a slide, i got this weird sizing, which you didn't have - was it something i did or should add to have consistent sizing for all the cards?


Screenshot 2023-05-25 at 17 17 19

JKarlavige:
@mirnawong1 Looks like that's due to the content in the cards being a different size. When stacking two cards on top of each other per column, the only way i can think of keeping them the same height is by using the same amount of content per card. 

But my recommendation would be to use single-row sliders. Each slide is equal height, so when cards are in the slider on a single row, they will always be equal height no matter the content. When stacking 2 on top of each other, those two cards have no way of knowing what the heights are of the next two stacked cards in the next column.

mirnawong1:
ah yea, you're right @JKarlavige ! i was going to for too much here! wow, this looks amazing to me and i think ready to ship??

mirnawong1:
tested. it out and looks awesome! future sugg could be to have the arrow bring you back to the beginning maybe but lets see if its even necessary!

Screenshot 2023-05-26 at 15 49 09

JKarlavige:
Sounds good it is merged, thanks @mirnawong1!!

mirnawong1:
thank you so much, Jason and Wendy (🙏 ) for all your help!

PR #3163, opened 2023-04-10
Update snapshots.md

davidkhala:
The official strategy name is `check`. `check_col` is a config item named which is used in `check` strategy

Document wording fix


dbeatty10 (approved):
Looks good to me -- thanks for submitting this @davidkhala !

PR #3130, opened 2023-04-03
Adding new method

matthewshaver:
## What are you changing in this pull request and why?
Adds the wildcard method to the reference guide for node selection. new in v1.5

## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) and [About versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) so my content adheres to these guidelines.
- [ ] Add a checklist item for anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

Adding new pages (delete if not applicable):
- [ ] Add page to `website/sidebars.js`
- [ ] Provide a unique filename for the new page

Removing or renaming existing pages (delete if not applicable):
- [ ] Remove page from `website/sidebars.js`
- [ ] Add an entry `_redirects`
- [ ] [Ran link testing](https://github.com/dbt-labs/docs.getdbt.com#running-the-cypress-tests-locally) to update the links that point to the deleted page

Issue #2702, opened 2023-01-12
Add wildcard selector method to dbt docs


z3z1ma:
### Contributions

- [X] I have read the contribution docs, and understand what's expected of me.

### Link to the page on docs.getdbt.com requiring updates

https://docs.getdbt.com/reference/node-selection/methods

### What part(s) of the page would you like to see updated?

# The "wildcard" method

The wildcard method can be used to select a model using unix-style wildcard expressions evaluated against the fqn.
See fnmatch [reference](https://docs.python.org/3/library/fnmatch.html#module-fnmatch) for syntax.

```bash
# These are equivalent
dbt run --select 'wildcard:*_model_?'
dbt run --select some_model_a some_model_b some_model_1

# So are these
dbt run --select 'wildcard:*_feature_model_[1-3]'
dbt run --select test_feature_model_1 new_feature_model_2 ml_feature_model_3
```

### Additional information

I have a PR I am opening in dbt-core now. I will post its link for reference. 
This PR is not relevant until the attached PR is merged.

https://www.notion.so/dbtlabs/dbt-Core-v1-5-docs-a2ae11be77bf47629f85faf7a3f27991

z3z1ma:
This becomes relevant after this is merged, tagging it for tracking purposes:

https://github.com/dbt-labs/dbt-core/pull/6599

runleonarun:
Thanks so much for opening this issue, @z3z1ma! I labeled this for 1.5 but let us know if it's available sooner.

z3z1ma:
@runleonarun the PR connected to this documentation change was merged 🎉 

Some small changes. The `wildcard:` is no longer needed as a prefix as the functionality is baked into the default selector. Furthermore all string based (ie. non status based) selectors have the functionality enabled. The list can be seen below. 

    FQN
    Tag
    Source
    Exposure
    Metric
    File
    Package
    TestName

Here is a markdown table to include describing the syntax. My recommendation is a note in all the above selectors that points to a specific section in the documentation containing this table. 

| Wildcard | Description                                               |
| -------- | --------------------------------------------------------- |
| \*       | matches any number of any characters (including none)     |
| ?        | matches any single character                              |
| [abc]    | matches one character given in the bracket                |
| [a-z]    | matches one character from the range given in the bracket |

cc @jtcohen6 @dbeatty10 for visibility/input (if nothing else) since they both grokked the issue through its lifecycle.

runleonarun:
[...]

PR #3189, opened 2023-04-14
Fix wildcard selection

jtcohen6:
Updates wildcard selection documentation (added in https://github.com/dbt-labs/docs.getdbt.com/pull/3130), based on the implementation we actually ended up with:
- https://github.com/dbt-labs/docs.getdbt.com/issues/2702#issuecomment-1455264269
- https://github.com/dbt-labs/dbt-core/pull/6599

Crucially, there is no standalone `wildcard:` method, but rather support for unix-style wildcards in a number of existing methods. See Alex's comment (linked above) above for details.


website/docs/reference/node-selection/methods.md
```diff
@@ -6,6 +6,29 @@ Selector methods return all resources that share a common property, using the
...
+| -------- | --------------------------------------------------------- |
+| \*       | matches any number of any characters (including none)     |
+| ?        | matches any single character                              |
+| [abc]    | matches one character given in the bracket                |
+| [a-z]    | matches one character from the range given in the bracket |
+
+For example:
+```
+dbt list --select **.folder_name.**
+dbt list --select package:*_source
```
dbeatty10: What does the ** mean?

I needed to add single or double quotes to it to work in my environment:
```suggestion
dbt list --select "*.folder_name.*"
dbt list --select "package:*_source"
```

Without adding quotes, I would get output like this (since the globbing was in my zsh shell rather on the graph nodes):
```
zsh: no matches found: package:*_source
```

jtcohen6: Great catch!

matthewshaver (approved):

PR #3215, opened 2023-04-19
slim down slim ci description

leoebfolsom:
I restructured a few lines with the goal of removing a link to an anchor (#deferral-and-state-comparison) which was only a few lines down. (I was actually disoriented when I clicked on the link and found that the page had only scrolled down a few lines. I didn't know where I was until I'd scrolled back up, at which point, I thought, "How silly to have clicked on this link; the content was already visible on my screen.") 

While going about this, I made a few other changes that I think make the description of the (undeniably confusing) concept of Slim CI a bit more readable.

## Checklist


runleonarun (approved):
This looks good, @leoebfolsom !

PR #3224, opened 2023-04-19
Clarify requirements for catalogs and schemas

mosabua:
## Checklist
- [x] verified locally


website/snippets/connect-starburst-trino/schema-db-fields.md
```diff
@@ -1,3 +1,3 @@
 [comment: For context, the section title used for this snippet is "Schemas and databases" ]: # 
 
-When selecting the database (catalog) and the schema, make sure the user has read and write access to both the provided database (catalog) and schema. This selection does not limit your ability to query the catalog. Instead, they serve as the default location for where tables and views are materialized. This _default_ can be changed later from within your dbt project.
\ No newline at end of file
+When selecting the catalog and the schema, make sure the user has read and write access to both. This selection does not limit your ability to query the catalog. Instead, they serve as the default location for where tables and views are materialized. In addition, the Trino connector used in the catalog must support creating tables and views in the catalog. This _default_ can be changed later from within your dbt project.
```
mosabua: Is that correct? Is view creation required? If so I assume we mean Trino views.. and then what connectors are even supported? I think only Hive, Delta Lake and Iceberg .. right?

mosabua: Also ... if this just sets the default for table and view creation.. can your SQL statements in dbt also access other catalogs?

hovaesco: Support for creating tables and views are not a hard requirement here. If a connector doesn't support view creation, it doesn't mean that a user cannot create tables.

nghi-ly (commented):
great callouts, @mosabua ! 

@dataders, curious to know your thoughts about this regarding database vs catalog.

website/snippets/connect-starburst-trino/schema-db-fields.md
dataders: @mosabua can you tell me more about your addition? do some Trino connectors not support creating tables and views?
> Trino connector used in the catalog must support creating tables and views in the catalog

 Or, do you mean that the connector must be correctly configured so that the cluster can make tables and views?

dataders (commented):
one question, and one piece of information.

today the Starburst integration in dbt Cloud prompts for a "database" field, following the convention of more standard database connectors. We have a [ticket internally](https://dbtlabs.atlassian.net/browse/CC-106) to make this change, but I'd prefer to hold off on merging this PR until the change is live. 👁️ @ericbaumann

website/snippets/connect-starburst-trino/schema-db-fields.md
mosabua: There are connectors that only support reading data, others support table creation, but not view creation, and so on. 

I am not sure what the hard requirement for a catalog is and I also cant conclude from the comment from @hovaesco 

At the moment I think table creation is a hard requirement so I will update the docs accordingly.

mosabua:
@nghi-ly and @dataders  .. when it comes to using "database" as a term this is misleading in the Trino context. The problem is that in Trino you configure a catalog to access a data source. Depending on the underlying data source and used connector the database in the data source can show up as the catalog .. or the schema. So using "database" as a term is not clear and should be avoid. 

Also note that in some underlying data sources there is no database at all. E.g. think of the Google Sheets connector where you have different sheets, or the Kafka connector where you deal with streams .. and so on.

And readers of this page will know about catalogs and schemas in Trino and it will be clear.

mosabua:
Rebased now as well

website/snippets/connect-starburst-trino/schema-db-fields.md
hovaesco: > Also ... if this just sets the default for table and view creation.. can your SQL statements in dbt also access other catalogs?

Yes, the connection profile catalog only defines to which catalog data will be loaded. For reading the data the only requirement is just to set up a catalog correctly.

> At the moment I think table creation is a hard requirement so I will update the docs accordingly.

True, that's the minimum, since I believe all connectors which support creating view support creating tables.

mosabua: Great.. I hope that makes sense @dataders

mosabua:
Let me know if I should rebase or you want to do it in the github UI before merge @dataders or @nghi-ly

mosabua:
I rebased now again.. can @dataders and @nghi-ly please have a look and let us know if this can be merged or needs changes?

dataders (approved):
I'm approving the change to only use "catalog" and drop mention to the word database. This change will be done by EOW so I'm fine to merge the docs as is

PR #3232, opened 2023-04-20
Revisions & additions to Model Versions

### LINKED ISSUE #3203: Keeping the original filename and table name when versioning for first time
### Contributions

https://docs.getdbt.com/docs/collaborate/govern/model-versions

[These](https://github.com/dbt-labs/docs.getdbt.com/pull/3195/files#diff-017bcb1f6a6ac50cbc818ffd9b893335887dd422783f1a2bf125e15bee36c895R95-R101) changes need to be reviewed for completeness and correctness.

The instigating issue was this this didn't work for me:
```yaml
  - v: 1
    config:
      alias: dim_customers # keep old relation location
```

This is what worked for me to keep both the filename and the table name the same as when it was un-versioned:
`models/_models.yml`
```yaml
models:
  - name: dim_customers
    versions:
      - v: 1
        defined_in: dim_customers  # keep original file name
```

`models/dim_customers.sql`
```yaml
{{ config(alias="dim_customers") }}

...
```

After my changes in #3195, this is not accurate 😢 :
```yaml
defined_in: dim_customers # keep original relation name
```

It should be something like this instead:
```yaml
defined_in: dim_customers # keep original file name
```

### Additional information

_No response_

### LINKED ISSUE #3226: Referencing specific model versions
### Contributions

Big picture: [this](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref) section has great info, but isn't easy to stumble onto when using the model versioning features of 1.5.

### Page 1
https://docs.getdbt.com/docs/collaborate/govern/model-versions
> In the meantime, anywhere that model is used downstream, it can be [referenced at a specific version](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref).

☝️ this should link to [here](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref)

### Page 2
In https://docs.getdbt.com/docs/collaborate/govern/model-versions, add a link to the ["versioned ref"](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref) in the [related documentation](https://docs.getdbt.com/docs/collaborate/govern/model-versions#related-documentation) section.

### Page 3
This page should have a section called "Referencing versioned models":
https://docs.getdbt.com/docs/collaborate/govern/model-versions#related-documentation

As shown in [versioned ref](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref), we should include examples that show the version parameter of `ref`:
```
select * from {{ ref('dim_customers', v=3) }}

-- or
select * from {{ ref('dim_customers', version=3) }}

-- or
select * from {{ ref('dim_customers', v="blue") }}
```

Alternatively, we can just link to https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref.

### Other pages to review
- https://docs.getdbt.com/reference/resource-properties/latest-version

No matter which of the pages some one lands on in relation to versioning models, it should be clear to them how to [ref](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#versioned-ref) them.

### Additional information

_No response_

jtcohen6:
[Preview: Collaborate with others > Model governance > Model versions](https://deploy-preview-3232--docs-getdbt-com.netlify.app/docs/collaborate/govern/model-versions)

We've written the minimal viable reference docs for this feature. I want to offer some more opinionated guidance & framing, and gesture in the direction of some best practices:
- **Don't** create a new version for every model change
- **Do** actually sunset/deprecate your old model versions

This does require a more personal tone, and a sense of future direction, than a lot of other (more-established) documentation. Very open to feedback.


website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+    default: "{{ target.name == 'dev' }}"
+    definition:
+      method: fqn
+      value: "*"
+      exclude:
+        - method: version
+          value: old
+```
+
+Finally, we intend to add support for **deprecating models** in dbt Core v1.6. When you slate a versioned model for deprecation, dbt will be able to provide more helpful warnings to downstream consumers of that model. Rather than just, "This model is going away," it's - "This older version of the model is going away, and there's a new version coming soon."
```
MichelleArk: Not sure if this is the best spot for it - but it'd be great to raise the tradeoff between cost/clutter (from maintaining multiple versions of a model in a warehouse) and providing consumers with enough time to gracefully migrate off old versions. It could be easy to let old versions of models pile up but deprecation dates create an explicit bound for how costly a migration should to be for a particular model

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+One approach is to force every model consumer to immediately handle the breaking change as soon as it's deployed to production. This is actually the appropriate answer at many smaller organizations, or while rapidly iterating on a not-yet-mature set of data models. But it doesn’t scale well beyond that.
 
-Instead, the model owner can create a **new version**, during which consumers can migrate from the old version to the new.
+Instead, for mature models at larger organizations, the model owner can create a **new version**, during which consumers can migrate from the old version to the new.
 
-In the meantime, anywhere that model is used downstream, it can be referenced at a specific version.
+In the meantime, anywhere that model is used downstream, it can continue to be referenced at a specific version.
+
+In the future, we intend to also add support for **deprecating models**. Taken together, model versions and deprecation offer a pathway for _sunsetting_ and _migrating_. In the short term, avoid breaking everyone's queries. Over the longer term, older & unmaintained versions go away—they do **not** stick around forever.
```
joellabes: This feels a lil fuzzy right now. 

>In the short term, avoid breaking everyone's queries

Who is the subject here? 
- Are _we_ (dbt Labs) avoiding breaking users' queries by shipping versioning first, while in the long term making it possible for old versions to go away? 
- Or is this an imperative to the reader: once deprecation ships, you should use versions to avoid breaking your own queries in the short term, and use the deprecation window to eventually get rid of unmaintained versions?

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+
+## When should you version a model?
+
+Many changes to a model are not breaking, and do not require a new version! Examples include adding a new column, or fixing a bug in modeling logic.
+
+By enforcing a model's contract, dbt can help you catch unintended changes to column names and data types that could cause a big headache for downstream queriers.
+
+It's also possible to change the model in more subtle ways — by recalculating a column in a way that doesn't change its name, data type, or enforceable characteristics—but would substantially change the results seen by downstream queriers.
+
+The process of sunsetting and migrating model versions requires real work, and may require significant coordination across teams. If, instead of using model versions, you opt for non-breaking changes wherever possible—that's a completely legitimate approach. Even so, after a while, you'll find yourself with lots of unused or deprecated columns. Many teams will want to consider a predictable cadence (once or twice a year) for bumping the version of their mature models, and taking the opportunity to remove no-longer-used columns.
```
joellabes: >that's a completely legitimate approach

if anything, I reckon that that's underselling it. You _should_ make non-breaking changes as much as possible, and if you _have_ to make breaking changes to a model, you should try to bunch them all together instead of dribbling out bad news over time. 

An interesting thing I just found while looking at breaking changes best practices: Our friends at HubSpot have a stated policy for deprecating tables, because they offer Snowflake Data Shares: https://developers.hubspot.com/docs/breaking-change-definition#snowflake-data-share

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
-dbt's model `versions` makes it possible to define multiple versions:
-- That share the same "reference" name
-- While reusing the same top-level properties, highlighting just their differences
+## How is this different from just creating a new model?
+
+Honestly, it's only a little bit different! There isn't much magic here, and that's by design.
+
+You've always been able to create a new model, and name it `dim_customers_v2`. Why should you opt for a "real" versioned model instead?
+
+First, the versioned model preserves its _reference name_. Versioned models are `ref`'d by their _model name_, rather than the name of the file that they're defined in. By default, the `ref` resolves to the latest version (as declared by that model's maintainer), but you can also `ref` a specific version of the model, with a `version` keyword.
```
joellabes: "reference name" is a new concept to me here. Is this the same as "model name"? I think it is from context, but if so then I don't think we want to introduce a brand new term as a one-off

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+
+<File name="models/schema.yml">
```
joellabes:
```suggestion
<File name="models/scratchpad.sql">
```

website/docs/docs/collaborate/govern/model-versions.md
joellabes: I think this needs a more worked example with examples and how the different elements combine: 

```yaml
models: 
  - name: dim_customers
    latest_version: 2
    versions:
      - v: 3
        defined_in: dim_customers_NOT_READY_YET.sql
        ...
      - v: 2
        alias: dim_customers
        ...
      - v: 1
        ...      
```

| v | ref syntax                                      | file name                       | table name                 |
|---|-------------------------------------------------|---------------------------------|----------------------------|
| 3 | `ref('dim_customers', v=3)`                       | dim_customers_NOT_READY_YET.sql | analytics.dim_customers_v3 |
| 2 | `ref('dim_customers')` or `ref('dim_customers', v=2)` | dim_customers_v2.sql            | analytics.dim_customers |
| 1 | `ref('dim_customers', v=1)`                       | dim_customers_v1.sql            | analytics.dim_customers_v1 |

website/docs/docs/collaborate/govern/model-versions.md
joellabes: Although if you take the table and sample code above, I think this File block is totally redundant

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+</File>
+
+Second, a versioned model can reuse the majority of its yaml properties and configuration. Each version needs to only say how it _differs_ from the shared set of attributes. This gives you an opportunity to highlight the differences across versions—which is otherwise difficult to detect in models with dozens or hundreds of columns—and to clearly track, in one place, all versions of the model which are currently live.
+
+Third, dbt supports `version`-based selection. For example, you could define a [default yaml selector](node-selection/yaml-selectors#default), to avoid running any old model versions in development—even as you continue to run them in production through a sunset and migration period:
+
+```yml
+selectors:
+  - name: exclude_old_versions
+    default: "{{ target.name == 'dev' }}"
```
joellabes, line 100: Does this need `| as_bool`?

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+    definition:
+      method: fqn
+      value: "*"
+      exclude:
+        - method: version
+          value: old
```
joellabes, line 106: It's not a straightforward jump to https://deploy-preview-3232--docs-getdbt-com.netlify.app/reference/node-selection/methods#the-version-method to find the valid options, should we link to that as well? or list them inline here?

website/docs/docs/collaborate/govern/model-versions.md
joellabes: >"This older version of the model is going away, and there's a new version coming soon."

I read this as "the new version hasn't been developed yet" - do you mean people might deprecate models ahead of when their replacements have been deployed "behind the feature flag" of not being `latest`?

Do we give a warning if `latest` is v3 and I query v2 by name, even if deprecation isn't scheduled? If so then it's more like 
>Rather than just, "This model has been superseded," it's - "This model has been superseded, and you need to act."

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -60,51 +104,140 @@ models:
 
 </File>
 
-If you wanted to make a breaking change to the model - for example, removing a column - you'd create a new model file (SQL or Python) encompassing those breaking changes. The default convention is naming the new file with a `_v<version>` suffix. The new version can then be configured in relation to the original model:
+If you wanted to make a breaking change to the model - for example, removing a column - you'd create a new model file (SQL or Python) encompassing those breaking changes. The default convention is naming the new file with a `_v<version>` suffix. The new version can then be configured in relation to the original model, in a way that highlights the diffs between them. Or, you can choose to define each model version with full specifications, and repeat the values they have in common.
+
+<Tabs>
+<TabItem value="Diffs only">
```
joellabes:
```suggestion
<TabItem value="Diffs only (recommended)">
```

❓

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -60,51 +104,140 @@ models:
...
+
+<File name="models/dim_customers_view.yml">
+
+```sql
+{{ config(alias = 'dim_customers') }}
+
+select * from {{ ref('dim_customers') }}
+```
+
+</File>
```
joellabes: As I was building my table of examples up top, I decided that if we're not going to do this magically behind the scenes, I think we should just encourage people to move the alias definition around in their YAML as they progress their models. Telling people to shepherd an entire extra model around by hand feels gross.

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -114,3 +247,32 @@ dbt.exceptions.AmbiguousAliasError: Compilation Error
...
+{{ dbt_utils.star(from=dim_customers_v1, except=["country_name"]) }}
+from {{ dim_customers_v1 }}
+```
+
+</File>
+
+Of course, if one model version makes meaningful and substantive changes to logic in another, it may not be possible to optimize it in this way. At that point, the cost of human intuition and legibility is more important than the cost of recomputing similar transformations.
+
+We expect to develop more opinionated recommendations as teams start adopting model versions in practice. One recommended pattern we can envision: Prioritize the definition of the `latest_version`, and define other versions (old and prerelease) based on their diffs from the latest. How?
+- Define the properties and configuration for the latest version in the top-level model yaml, and the diffs for other versions below (via `include`/`exclude`)
```
joellabes: 👨‍🍳 💋

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -114,3 +247,32 @@ dbt.exceptions.AmbiguousAliasError: Compilation Error
...
+- Where possible, define other versions as `select` transformations, which take the latest version as their starting point
+- When bumping the `latest_version`, migrate the SQL and yaml accordingly. In this case, we would see if it's possible to redefine `v1` with respect to `v2`.
```
joellabes: >in this case

The sample above where country name is removed? I don't quite follow this bit

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
-In the meantime, anywhere that model is used downstream, it can be referenced at a specific version.
+In the meantime, anywhere that model is used downstream, it can continue to be referenced at a specific version.
+
+In the future, we intend to also add support for **deprecating models**. Taken together, model versions and deprecation offer a pathway for _sunsetting_ and _migrating_. In the short term, avoid breaking everyone's queries. Over the longer term, older & unmaintained versions go away—they do **not** stick around forever.
+
+## When should you version a model?
+
+Many changes to a model are not breaking, and do not require a new version! Examples include adding a new column, or fixing a bug in modeling logic.
+
+By enforcing a model's contract, dbt can help you catch unintended changes to column names and data types that could cause a big headache for downstream queriers.
```
joellabes: True, but nothing to do with versioning on its own. Is the implicit recommendation here "when you break your contract, you should bump versions"?

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -9,32 +9,76 @@ description: "Version models to help with lifecycle management"
...
+
+It's also possible to change the model in more subtle ways — by recalculating a column in a way that doesn't change its name, data type, or enforceable characteristics—but would substantially change the results seen by downstream queriers.
```
joellabes: Likewise here, it feels like this would benefit from driving the point home: if you are going to surprise your querier, probably bump versions

website/docs/docs/collaborate/govern/model-versions.md
joellabes: Does `latest` version of a model get to look for the `_vX`-less sql file? I don't like that the grid here forces new sql files for every version, so [you don't get a nice git diff](https://github.com/dbt-labs/dbt-core/discussions/6736#:~:text=This%20also%20simplifies%20the%20git%20and%20code%20review%20flow%20%2D%20when%20you%20bump%20a%20version%2C%20you%20are%20comparing%20the%20old%20version%20of%20the%20file%20to%20the%20new%20version%20of%20the%20same%20file%20instead%20of%20doing%20a%20diff%20between%20dim_customers_v1.sql%20and%20dim_customers_v2.sql)

joellabes (changes requested):
This is wonderful. Nitpicks, pedantry and clarifying questions from me throughout, but I think that vibe-wise this is great!

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: @joellabes I've had another thought here: It would be possible to implement this as a standard pattern, with a modification to the [`generate_alias_name`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/get_custom_name/get_custom_alias.sql) macro.

```sql
{% macro generate_alias_name(custom_alias_name=none, node=none) -%}
    {%- if custom_alias_name -%}
        {{ return(custom_alias_name | trim) }}
    {%- elif node.version and not node.is_latest_version -%} {# <--- this bit #}
        {{ return(node.name ~ "_v" ~ (node.version | replace(".", "_"))) }}
    {%- else -%} {# latest version has standard behavior #}
        {{ return(node.name) }}
    {%- endif -%}
{%- endmacro %}
```

This way, whichever version is `latest_version`, it always lands in the model's "canonical" location.

**Should we make that the default behavior?** Or should it be something that end users opt into? I'm inclined to be a bit more opinionated, and say this should be the default—**really** emphasize that the latest version is _the thing_, and the old/new versions are mechanisms for managing change—but it does add a bit more inconsistency.

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: Good point - my wording here was not clear.

We'll support `deprecation_date` for both versioned and non-versioned models. I was trying to differentiate here between the experience of downstream consumers, when an upstream producer deprecates a model that's versioned, versus a standalone (non-versioned) model. Again, trying to make the case for - why use versions at all? Why not just create two separate SQL files, `my_model_old` and `my_model_new`, and dbt is none the wiser?

> Do we give a warning if `latest` is v3 and I query v2 by name, even if deprecation isn't scheduled?

As a **consumer**, I don't think we'd give a warning if you query `v2` by name, and it's not scheduled for deprecation. I think we'd strongly encourage **producers** to put a deprecation date on `v2` at the same time that they roll over `latest_version: 3`.

At first, I thought you were asking something slightly different - should we warn the **consumer** about an unpinned reference to a model, as soon as a prerelease version becomes available?
```
12:53:34  4 of 4 START sql view model dbt_jcohen.another_model ........................... [RUN]
12:53:34  FYI!
Found an unpinned reference to versioned model 'my_model' in 'my_dbt_project'.
Resolving to latest version: my_model.v2
A prerelease version 3 is available. It has not yet been marked 'latest' by its maintainer.
When that happens, this reference will resolve to my_model.v3 instead.

  Try out v3: {{ ref(my_dbt_project, my_model, v=3) }}
  Pin to v2:  {{ ref(my_dbt_project, my_model, v=2) }}
```

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: > Does `latest` version of a model get to look for the _vX-less sql file?

That's not the case in the current implementation. **Should it be?** I think we could do this. (Naive attempt: https://github.com/dbt-labs/dbt-core/pull/7435/commits/b13cd2b3587eea8f9ab1d4f78870a8ba00486ab5)

Even if we don't do this, you could do the same thing as with aliases, and keep moving the `defined_in` property around, so that `dim_customers.sql` is always your "latest":

```yml
models: 
  - name: dim_customers
    latest_version: 2
    versions:
      - v: 3
        defined_in: dim_customers_NOT_READY_YET.sql
        ...
      - v: 2
        # because this is the latest, it should have the canonical file name + alias
        alias: dim_customers
        defined_in: dim_customers
        ...
      - v: 1
        ...      
```

But if that's our strong recommendation - let's just make it the default behavior

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: That's a great find! We're taking exactly the same approach, in terms of what we're considering breaking versus non-breaking, and in recommending a clear migration window:

> In this case, HubSpot will add a new table with the same data to the share so that you can begin using the new name. The old table will continue to exist until the end of the 90 day notice period.

website/docs/docs/collaborate/govern/model-versions.md
joellabes: Yes I definitely agree it should be the default. More to come on the Slack thread

website/docs/docs/collaborate/govern/model-versions.md
joellabes: It feels weird/WET to _have_ to move both the `alias` and `defined_in` around over time to just get the same name as is already defined in the `name` key up top. 

Edit: here's my actual objection: `defined_in` feels like it should be a "break glass in case of emergency" property, not something that gets rolled out everywhere. If you're using `defined_in`, you best have a good reason. Encouraging it to be everywhere cheapens that a bit

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: It actually doesn't. I think the reason is because yaml treats the string `"True"` as truthy

jtcohen6:
Thank you for the excellent feedback ❤️

My revision includes two significant assumptions:
1. That we will implement two of the UX changes proposed in https://github.com/dbt-labs/dbt-core/pull/7435 (which I can remove, if implementing proves impossible before Thursday):
    - Latest version can be defined in `<model_name>.sql` (no suffix)
    - Unpinned `ref` will log if a newer prerelease version is detected
2. That, for now, the best / recommended approach to handle aliasing is with a hook that creates a view pointing to the latest version, as in [this gist](https://gist.github.com/jtcohen6/68220cd76b0bde088d3439664ccfb013)

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,65 +6,90 @@ description: "Version models to help with lifecycle management"
...
 
-Model versions are different. Multiple versions of a model will live in the same code repository at the same time and be deployed into the same data environment simultaneously. This is similar to how web APIs are versioned—multiple versions are live simultaneously; older versions are often eventually sunsetted.
+**Versioned models are different.** Defining model `versions` is appropriate when there are people, systems, and processes beyond your team's control, inside or outside of dbt. You can neither simply go migrate them all, nor break their queries on a whim. I need to do my part by offering a migration path, with clear diffs and deprecation dates.
+
+Multiple versions of a model will live in the same code repository at the same time, and be deployed into the same data environment simultaneously. This is similar to how web APIs are versioned—multiple versions are live simultaneously; older versions are often eventually sunsetted.
```
joellabes: (but hopefully not more than 2)

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,65 +6,90 @@ description: "Version models to help with lifecycle management"
...
+
+When you make updates to a model's source code—its logical definition, in SQL or Python, or related configuration—dbt can [compare your project to previous state](project-state), enabling you to rebuild only models that have changed, and models downstream of a change. In this way, it's possible to develop changes to a model, quickly test in CI, and efficiently deploy into production—all coordinated via your version control system.
 
-Model versions are different. Multiple versions of a model will live in the same code repository at the same time and be deployed into the same data environment simultaneously. This is similar to how web APIs are versioned—multiple versions are live simultaneously; older versions are often eventually sunsetted.
+**Versioned models are different.** Defining model `versions` is appropriate when there are people, systems, and processes beyond your team's control, inside or outside of dbt. You can neither simply go migrate them all, nor break their queries on a whim. I need to do my part by offering a migration path, with clear diffs and deprecation dates.
```
joellabes: _You_ can neither simply go migrate them all, nor break their queries on a whim. _I_ need to do my part by offering a migration path, with clear diffs and deprecation dates.

Both of these are the same actor aren't they? Did you just feel weird about telling people they have to do their part?

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -180,11 +217,17 @@ models:
...
 
-The above configuration will create two models (one for each version), and produce database relations with aliases `dim_customers_v1` and `dim_customers_v2`.
 
-By convention, dbt will expect those two models to be defined in files named `dim_customers_v1.sql` and `dim_customers_v2.sql`. (It is possible to override this by setting `defined_in: any_file_name_you_want`, but we strongly encourage you to follow the convention!)
+The above configuration will create two models, `dim_customers.v1` and `dim_customers.v2`.
+
+**Where are they defined?**
+
+
+**Where will they be materialized?** By convention, these will create database relations with aliases `dim_customers_v1` and `dim_customers_v2`. We recommend that you also create a view, named `dim_customers`, pointing to the latest version. Check out guidance on an easy & repeatable way to do that.
```
joellabes: >Check out guidance on an easy & repeatable way to do that.

Where is that guidance? Looks like a link is missing here

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -180,11 +217,17 @@ models:
...
+
+By convention, dbt will expect those two models to be defined in files named `dim_customers_v1.sql` and `dim_customers_v2.sql`. It will also accept `dim_customers.sql` (no suffix) as the definition of the latest version. (It is possible to override this by setting `defined_in: any_file_name_you_want`, but we strongly encourage you to follow the convention!)
```
joellabes: >It is possible to override this by setting `defined_in: any_file_name_you_want`

You have to include the `.sql` suffix right? 

(Related: if it's optional, what happens if you have `any_file_you_want.sql` and `any_file_you_want.py`?)

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -222,18 +265,22 @@ You could use the `alias` configuration:
 
 </File>
 
-Or, you could define a separate view that always points to the latest version of the model. We recommend this pattern because it's the most transparent and easiest to follow.
+Or, you could do one better: In a model or project hook, create a view named `dim_customers` that always points to the latest version of the `dim_customers` model. You can find logic for just such a hook in [this gist](https://gist.github.com/jtcohen6/68220cd76b0bde088d3439664ccfb013/edit). Then, for all the versioned models in your project, it's as simple as:
 
-<File name="models/dim_customers_view.yml">
+<!-- TODO: add the macro from my gist to dbt-core. Better as on-run-end or post-hook? -->
```
joellabes: I'd say post-hook for all the same reasons we used to encourage doing grants in post-hooks - the changes apply immediately instead of having to wait for the entire 3 hour run to complete

website/docs/reference/resource-properties/versions.md
```diff
@@ -35,24 +35,29 @@ The standard convention for naming model versions is `<model_name>_v<v>`. This h
 
 The version identifier for a version of a model. This value can be numeric (integer or float), or any string.
 
-The value of the version identifier is used to order versions of a model relative to one another. If a versioned model does _not_ explicitly configure a [`latest_version`](resource-properties/latest-version), the highest version number is used as the latest version to resolve `ref` calls to the model without a `version` argument.
+The value of the version identifier is used to order versions of a model relative to one another. If a versioned model does _not_ explicitly configure a [`latest_version`](resource-properties/latest_version), the highest version number is used as the latest version to resolve `ref` calls to the model without a `version` argument.
 
 In general, we recommend that you use a simple "major versioning" scheme for your models: `v1`, `v2`, `v3`, etc, where each version represents a breaking change from previous versions. However, you are welcome to use other versioning schemes.
```
joellabes: >However, you are welcome to use other versioning schemes 

as long as they behave correctly when `sort()`ed. (or however we're actually doing it). 

On that note, do we handle people putting `v`s in their yaml? What would happen if I did this? 

```
models:
  - name: dim_customers
    versions: 
      - v: v1
      ...
      - v: 2
```
Both from a sorting perspective and an alias-creation perspective - would I wind up with `dim_customers_vv1` which outranked the `2` in sort order?

joellabes (commented):
🔥

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: :) I suspect I'm switching frequently between first & second person - worth another reread just for this

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: You don't need to include the file extension (and in fact, shouldn't)

> (Related: if it's optional, what happens if you have `any_file_you_want.sql` and `any_file_you_want.py?`)

Not allowed - model file names still need to be globally unique, independent of the file extension. I think I have a note about this in the reference docs for `defined_in` - I think I'll add a link there from here.

website/docs/reference/resource-properties/versions.md
jtcohen6: Yes & yes. I'll add an explicit caution that people should not include `v` in their version identifier.

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: I've actually been reasonably consistent:
- "we" = dbt Labs, developers/maintainers of dbt-core
- "you" = user of dbt, maintainer of a versioned model

jtcohen6:
As with https://github.com/dbt-labs/dbt-core/pull/7435#issuecomment-1520185944, holding off on merging this until we decide whether to vendor [`create_latest_version_view`](https://github.com/dbt-labs/docs.getdbt.com/pull/3232/files#diff-017bcb1f6a6ac50cbc818ffd9b893335887dd422783f1a2bf125e15bee36c895R270-R291) (recommended post-hook) within `dbt-core` directly.

**Update:** Let's opt for, this will come in v1.6; in the meantime, a macro you can copy-paste-edit-post-hook.

In the meantime, comments from other reviewers still welcome!

website/docs/reference/model-properties.md
```diff
@@ -16,7 +16,7 @@ models:
     [description](description): <markdown_string>
     [docs](/reference/resource-configs/docs):
       show: true | false
-    [latest_version](resource-properties/latest-version): <version_identifier>
+    [latest_version](resource-properties/latest_version): <version_identifier>
```
mirnawong1, line 19: i dont understand why an underscore was added here, the actual page is https://docs.getdbt.com/reference/resource-properties/latest-version and a `latest_version` brings the user to a 'page not found'. suggesting it goes back to the `latest-version`

```suggestion
    [latest_version](resource-properties/latest-version): <version_identifier>
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -60,51 +141,178 @@ models:
...
+    {% if model.get('version') and model.get('version') == model.get('latest_version') %}
+
+        {% set new_relation = api.Relation.create(
+            database = this.database,
+            schema = this.schema,
+            identifier = model['name']
+        ) %}
+        
+        {% set existing_relation = load_relation(new_relation) %}
+        {{ drop_relation_if_exists(existing_relation) }}
```
jtcohen6: Since the whole point is that this view should be live-queryable from a BI tool... I don't think we want to be dropping it outside of a transaction / atomic operation.

I don't think we have a handy cross-db way to do this, outside of the actual `view` materialization logic. This might take some fudging.

website/docs/reference/model-properties.md
jtcohen6: @mirnawong1 The name of this resource property is `latest_version` (underscore). I looked at some other similar properties/configs, and they all have underscores in their file name / `id`:
- https://docs.getdbt.com/reference/resource-configs/sql_header
- https://docs.getdbt.com/reference/resource-configs/column_types
- https://docs.getdbt.com/reference/resource-configs/store_failures
- (etc)

So I renamed the page, and added a redirect for it

_redirects
```diff
@@ -278,6 +278,7 @@ docs/dbt-cloud/using-dbt-cloud/cloud-model-timing-tab /docs/deploy/dbt-cloud-job
 /docs/artifacts /docs/dbt-cloud/using-dbt-cloud/artifacts 301
 /docs/bigquery-configs /reference/resource-configs/bigquery-configs 301
 /reference/resource-properties/docs /reference/resource-configs/docs 301
+/reference/resource-properties/latest-version /reference/resource-configs/latest_version 301
```
runleonarun: I think this should be as follows:
```suggestion
/reference/resource-properties/latest-version /reference/resource-properties/latest_version 301
```

website/docs/docs/collaborate/govern/model-contracts.md
```diff
@@ -6,7 +6,7 @@ description: "Model contracts define a set of parameters validated during transf
 ---
 
 :::info New functionality
-This functionality is new in v1.5.
+This functionality is new in v1.5 — if you have thoughts, weigh into the [GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/6726)!
```
runleonarun: "Weigh in" might be vague.
```suggestion
This functionality is new in v1.5 — if you have feedback, then participate in the [GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/6726)!
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,120 @@ description: "Version models to help with lifecycle management"
 ---
 
 :::info New functionality
-This functionality is new in v1.5.
+This functionality is new in v1.5 — if you have thoughts, weigh into the [GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/6736)!
```
runleonarun:
```suggestion
This functionality is new in v1.5 — if you have feedback, then participate in the [GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/6736)!
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,120 @@ description: "Version models to help with lifecycle management"
...
 :::
 
-API versioning is a _complex_ problem in software engineering. It's also essential. Our goal is to _overcome obstacles to transform a complex problem into a reality_.
+Versioning APIs is a hard problem in software engineering. At the root of the challenge is the fact that the producers and consumers of an API have competing incentives:
```
runleonarun:
```suggestion
Versioning APIs is a hard problem in software engineering. The root of the challenge is that the producers and consumers of an API have competing incentives:
```

runleonarun (commented):
Just a few comments now and I will finish reviewing tomorrow!

_redirects
jtcohen6: good catch!

website/docs/docs/collaborate/govern/model-contracts.md
jtcohen6: 👍 replacing with ~"comment on"~ "participate in"

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
 ---
 
 :::info New functionality
-This functionality is new in v1.5.
+This functionality is new in v1.5 — if you have thoughts, participate in [the discussion on GitHub](https://github.com/dbt-labs/dbt-core/discussions/6736)!
 :::
 
-API versioning is a _complex_ problem in software engineering. It's also essential. Our goal is to _overcome obstacles to transform a complex problem into a reality_.
+Versioning APIs is a hard problem in software engineering. The root of the challenge is that the producers and consumers of an API have competing incentives:
+- Producers of an API need the ability to make changes to its logic. There is a real cost associated with maintaining legacy endpoints forever, but losing the trust of downstream users is far costlier.
```
runleonarun: I think we can make this pack more punch by writing out some of the passive language:
```suggestion
- Producers of an API need the ability to modify its logic. Although maintaining legacy endpoints forever incurs a significant expense, it costs more to lose the trust of downstream users.
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+- Consumers of an API need to trust in its stability—their queries will keep working, and won't break without warning. There is a real cost associated with migrating to a newer API version, but unplanned migration is far costlier.
```
runleonarun:
```suggestion
- Consumers of an API need to trust in its stability—their queries will keep working, and won't break without warning. Although migrating to a newer API version incurs an expense, an unplanned migration is far costlier.
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+
+The goal of model versions is not to make the problem go away, nor to pretend it's somehow easier or simpler than it is. Rather, we want dbt to provide tools that make it possible to tackle this problem, thoughtfully and head-on, and to develop standard patterns for solving it.
```
runleonarun: To be clear, the "tools" are model versioning, right? I'd also suggest flipping the two sentences so the what model versioning does do is not buried.

```suggestion
The goal of model versions is not to make the problem go away, nor to pretend it's somehow easier or simpler than it is. Rather, model versioning makes it possible to tackle this problem, thoughtfully and head-on, and to develop standard patterns for solving it.
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+
+There is a real trade-off that exists here—the cost to frequently migrate downstream code, and the cost (and clutter) of materializing multiple versions of a model in the data warehouse. Model versions do not make that problem go away, but by setting a deprecation date, and communicating a clear window for consumers to gracefully migrate off old versions, they put a known boundary on the cost of that migration.
+
+## When should you version a model?
+
+By enforcing a model's contract, dbt can help you catch unintended changes to column names and data types that could cause a big headache for downstream queriers. These changes, when made intentionally, would require a new model version. But many changes are not breaking, and don't require a new version—such as adding a new column, or fixing a bug in an existing column's calculation.
```
runleonarun:
```suggestion
By enforcing a model's contract, dbt can help you catch unintended changes to column names and data types that could cause a big headache for downstream queriers. These changes, when made intentionally, would require a new model version. But when making non-breaking changes, you don't need a new version—such as adding a new column, or fixing a bug in an existing column's calculation.
```

runleonarun: @jtcohen6 for this sentence, do we want to say that they get the option of creating a new version vs fixing the problem? It feels like that the undertone here, but we might want to be explicit. Using "require" kind of sounds like dbt will require it, which doesn't seem the case later.
>These changes, when made intentionally, would require a new model version

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+
+When you make updates to a model's source code—its logical definition, in SQL or Python, or related configuration—dbt can [compare your project to previous state](project-state), enabling you to rebuild only models that have changed, and models downstream of a change. In this way, it's possible to develop changes to a model, quickly test in CI, and efficiently deploy into production—all coordinated via your version control system.
+
+**Versioned models are different.** Defining model `versions` is appropriate when there are people, systems, and processes beyond your team's control, inside or outside of dbt. You can neither simply go migrate them all, nor break their queries on a whim. You need to do my part by offering a migration path, with clear diffs and deprecation dates.
```
runleonarun: Not sure if "do my part" was a typo? 
```suggestion
**Versioned models are different.** Defining model `versions` is appropriate when people, systems, and processes beyond your team's control, inside or outside of dbt, depend on your models. You can neither simply go migrate them all, nor break their queries on a whim. You need to offer a migration path, with clear diffs and deprecation dates.
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+
+Multiple versions of a model will live in the same code repository at the same time, and be deployed into the same data environment simultaneously. This is similar to how web APIs are versioned: Multiple versions are live simultaneously, two or three, and not more). Over time, newer versions come online, and older versions are sunsetted .
+
+## How is this different from just creating a new model?
+
+Honestly, it's only a little bit different! There isn't much magic here, and that's by design.
+
+You've always been able to copy-paste, create a new model file, and name it `dim_customers_v2.sql`. Why should you opt for a "real" versioned model instead?
+
+As the **producer** of a versioned model:
```
runleonarun, line 71: These benefits are super clear! Love how this section reads!

runleonarun: You might consider bullets instead of steps. Then the reader can focus on the content. Steps usually indicate you need to *do* something.

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
 ---
 
 :::info New functionality
-This functionality is new in v1.5.
+This functionality is new in v1.5 — if you have thoughts, participate in [the discussion on GitHub](https://github.com/dbt-labs/dbt-core/discussions/6736)!
 :::
 
-API versioning is a _complex_ problem in software engineering. It's also essential. Our goal is to _overcome obstacles to transform a complex problem into a reality_.
+Versioning APIs is a hard problem in software engineering. The root of the challenge is that the producers and consumers of an API have competing incentives:
```
runleonarun, line 12: We never actually say how model versioning relates to API versioning. I wonder if we could call that out here before we start talking about the problems with versioning APIs?

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+| 3 | "prerelease" | `ref('dim_customers', v=3)`                           | `dim_customers_v3.sql`                          | `analytics.dim_customers_v3`                                             |
+| 2 | "latest"     | `ref('dim_customers', v=2)` **and** `ref('dim_customers')`  | `dim_customers_v2.sql` **or** `dim_customers.sql` | `analytics.dim_customers_v2` **and** `analytics.dim_customers` (recommended) |
+| 1 | "old"        |  `ref('dim_customers', v=1)`                           | `dim_customers_v1.sql`                          | `analytics.dim_customers_v1`                                             |
+
+As you'll see in the implementation section below, a versioned model can reuse the majority of its yaml properties and configuration. Each version needs to only say how it _differs_ from the shared set of attributes. This gives you, as the producer of a versioned model, the opportunity to highlight the differences across versions—which is otherwise difficult to detect in models with dozens or hundreds of columns—and to clearly track, in one place, all versions of the model which are currently live.
```
runleonarun:
```suggestion
As you'll see in the implementation section below, a versioned model can reuse the majority of its YAML properties and configuration. Each version needs to only say how it _differs_ from the shared set of attributes. This gives you, as the producer of a versioned model, the opportunity to highlight the differences across versions—which is otherwise difficult to detect in models with dozens or hundreds of columns—and to clearly track, in one place, all versions of the model which are currently live.
```

website/docs/docs/collaborate/govern/model-versions.md
```diff
@@ -6,39 +6,140 @@ description: "Version models to help with lifecycle management"
...
+Found an unpinned reference to versioned model 'dim_customers'.
+Resolving to latest version: my_model.v2
+A prerelease version 3 is available. It has not yet been marked 'latest' by its maintainer.
+When that happens, this reference will resolve to my_model.v3 instead.
+
+  Try out v3: {{ ref('my_dbt_project', 'my_model', v='3') }}
+  Pin to  v2: {{ ref('my_dbt_project', 'my_model', v='2') }}
+```
 
 ## How to create a new version of a model
```
runleonarun, line 123: Not for this week, but eventually, we might want to pull the procedure section out into its own page so people who don't need the "why" and other context can more easily find the "how."

runleonarun (dismissed):
@jtcohen6 This looks good! I have some wording suggestions to clarify the message here. I also have a few questions that might be worth addressing. 

Approving so you can release during your day.

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: Good clarifying question. If you make a breaking contract change, dbt will raise an error during CI — and you'd need to "merge on red" (you always _can_ do it). https://docs.getdbt.com/reference/resource-configs/contract#detecting-breaking-changes
```
  Consider making an additive (non-breaking) change instead, if possible.
  Otherwise, create a new model version: https://docs.getdbt.com/docs/collaborate/govern/model-versions
```

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: *`do your part`! whoops. I like your suggestion better

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: good call!

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: Added these sentences in between, to try and connect the dots:
```
When sharing a final dbt model with other teams or systems, that model is operating like an API. When the producer of that model needs to make significant changes, how can they avoid breaking the queries of its users downstream?
```

website/docs/docs/collaborate/govern/model-versions.md
jtcohen6: Heard! A lot of the "how" details are also captured in the reference documentation: https://docs.getdbt.com/reference/resource-properties/versions

For now, we should expect that the main audience for these docs is people trying to learn about & understand the new feature. As more the concept becomes more established, and people are already convinced they want to use the feature, we can make it even easier to _just do the thing_

PR #3280, opened 2023-04-26
v1.5 - a few more fixups

jtcohen6:
- Add page missing from `sidebars`
- Fix `versionedPages` (& reorder so alphabetical)
- Fix examples of setting `group` on a model - this is set as a config, not a top-level property


mirnawong1 (approved):
approved this for you @jtcohen6

PR #3292, opened 2023-04-27
Clarify permissions to create a webhook sub

cafzal:
Clarify which user permissions for each plan type can create a webhooks subscription


runleonarun (approved):
LGTM!

PR #3305, opened 2023-04-28
Add Model contracts section

damian3031:
Added Model contracts section in Trino configuration docs.


website/docs/reference/resource-configs/trino-configs.md
```diff
@@ -365,3 +365,7 @@ models:
 ```
 </File>
 
+## Model contracts
+
+The `dbt-trino` adapter supports [Model contracts](/docs/collaborate/govern/model-contracts). However, currently only `not_null` [constraint](/reference/resource-properties/constraints) is supported.
```
nghi-ly: small sugg for a better clarity. 

```suggestion
The `dbt-trino` adapter supports [model contracts](/docs/collaborate/govern/model-contracts). Currently, only [constraints](/reference/resource-properties/constraints) with `type` as `not_null` are supported.
```

website/docs/reference/resource-configs/trino-configs.md
```diff
@@ -365,3 +365,7 @@ models:
...
+Please ensure that underlying connector supports `not null`, before using this constraint in your model.
```
nghi-ly:
```suggestion
Before using a constraint in your model, make sure the underlying connector supports `not_null`.
```

website/docs/reference/resource-configs/trino-configs.md
dataders:
```suggestion
[Model contracts](/docs/collaborate/govern/model-contracts) are supported. However the only supported constraint is `not_null` [contracts: constraints](/reference/resource-properties/constraints)
```

dataders (dismissed):
looks good! @nghi-ly would love your eyes on this if you get a chance

website/docs/reference/resource-configs/trino-configs.md
nghi-ly: curious, what happens if it doesn't? what sort of issues will the user encounter if they do end up using it? it's good practice to inform people _why_ they should do a thing as it helps with learning.

nghi-ly (commented):
thank you so much for adding this section, @damian3031 ! appreciate your contribution! left a couple small suggs

website/docs/reference/resource-configs/trino-configs.md
damian3031: While trying `not_null` constraint on connector which doesn't support it, user will receive Trino error at dbt run stage, e.g.:
`Hive tables do not support NOT NULL columns`
Should I add detailed description in docs what would happen then? I think that it is not necessary in this case, as making sure that you use supported feature is self explanatory IMHO - you can't use feature, which is not supported 🙂 
WDYT?

website/docs/reference/resource-configs/trino-configs.md
```diff
@@ -365,3 +365,7 @@ models:
 ```
 </File>
 
+## Model contracts
+
+The `dbt-trino` adapter supports [model contracts](/docs/collaborate/govern/model-contracts). Currently, only [constraints](/reference/resource-properties/constraints) with `type` as `not_null` are supported.
+Before using `not_null` constraint in your model, make sure the underlying connector supports `not null`.
```
nghi-ly: you're right, @damian3031 , that would be too much. how about this or something similar? 

```suggestion
Before using `not_null` constraints in your model, make sure the underlying connector supports `not null` to avoid running into errors.```

nghi-ly (commented):
updated the sentence in question. lemme know what you think!

website/docs/reference/resource-configs/trino-configs.md
damian3031: adjusted, thanks!

nghi-ly (approved):

PR #3312, opened 2023-05-01
Clarify reasonable use language

cafzal:
Refining language until we define the node size response limit

Making the language clearer until we implement the response size limit, and only for environment-level API


mirnawong1 (approved):
approved @cafzal and ready for you to merge!

PR #3316, opened 2023-05-01
Switch from dots to underscores for versioning

joellabes:
The docs for model versioning use `dim_customers.v1` in a lot of places, instead of `dim_customers_v1`. This was an original proposal for model version naming that got abandoned due to database peculiarities. 

This PR swaps out the `.` for `_` everywhere except for when it's being used in a model's fully qualified name (where the dots are actually how it's represented.)

I'm assuming this was an oversight as the product evolved; maybe it's on purpose across the board? If so then this is more up for debate, but IMO the more time we spend talking about versioned models along our preferred naming scheme (single digit with an underscore) the better. The fact that we use dots behind the scenes isn't relevant to most folks.


nghi-ly (approved):
🚢

joellabes:
Will wait to merge until @jtcohen6 gets a chance to comment if he wants to, in case I am misinterpreting his goals

jtcohen6:
@joellabes I'm open to this change if you think it's clarifying / less confusing for users!

Here's my thought process:
- The model's `name` is always just `dim_customers`. It has a new `version` component, which is `1`, `2`, etc.
- In the model's `unique_id` and `fqn`, these components are dot-separated: `model.<project_name>.dim_customers.v1` and `<project_name>.<subdirectory>.dim_customers.v1`, respectively
- Following the default naming conventions, the model's file name and database alias will both be `dim_customers_v1` (underscore)

So when talking about the model as a dbt node/resource, in abstract of its SQL file definition and its materialized relation identifier, my inclination was to use the dot (`dim_customers.v1`). For example, I can select the model with the dot, but I can't with the underscore:
```
$ dbt --quiet ls -s dim_customers.v1
my_dbt_project.dim_customers.v1
$ dbt --quiet ls -s dim_customers_v1
... nothing ...
```

Should we just fix that, and support selecting versioned models as `<model_name>_v<v>`?

joellabes:
Hmm! 

> Should we just fix that, and support selecting versioned models as `<model_name>_v<v>`?

Yeah I think this is a good idea regardless, to reduce surprises.

Don't know how it makes me feel about the writing in the docs - the dots are _righter_, but unless you're poking around in the artifacts you're never going to interact with them right?

jtcohen6:
>> Should we just fix that, and support selecting versioned models as `<model_name>_v<v>`?

> Yeah I think this is a good idea regardless, to reduce surprises.

We did this for v1.6: https://github.com/dbt-labs/dbt-core/pull/7995

Let's move ahead with this PR!

dbeatty10 (approved):
~`thank.you`~ `thank_you` @joellabes !

PR #3377, opened 2023-05-16
add ability to have four columns

mirnawong1:
adding this css to allow 4 grid columns per @JKarlavige 's amazing guidance. This would most likely work visually for cards that have little text.

[see slack thread](https://dbt-labs.slack.com/archives/C01C2H652BC/p1684250667489089)

## Checklist

- [ ] Add an entry `website/static/_redirects`


website/src/css/custom.css
```diff
@@ -1764,6 +1764,19 @@ section > h2:not(.resource-section) {
...
+  grid-template-columns: repeat(1, 1fr);
+  grid-gap: var(--ifm-spacing-horizontal)
+}
+
+@media(min-width: 996px) {
+  .grid--4-col {
+    grid-template-columns: repeat(4, 1fr);
+    grid-gap: calc(var(--ifm-spacing-horizontal) * 2);
+  }
+}
```
JKarlavige: Testing locally, it looks like the grid of 4 runs out of room around `1500px` wide. So in between ~996px and ~1500px, the cards will start to be pushed out of the container and off the screen.

Adding another media query below the 996px media query could solve this. Rather than having 4 columns down to a min-width of 996px, I recommend creating another media query below this one with a min-width of 1500px, and have the `grid-template-columns: repeat(4, 1fr);` css property set to 4 within this 1500px media query.

Then for this 996px media query, I would change the `grid-template-columns` property from 4 to 2. So on screens 1500px or wider, it will have 4 columns, then from 996px to 1499px it will have 2 columns, then 1 column for anything below 996px.

image

JKarlavige (changes requested):
Left one comment with a requested change to support screens less than 1500px wide! Aside from that the css looks great!

website/src/css/custom.css
mirnawong1: great callout, thank you so much @JKarlavige ! ok I've amended that on my end and i also added a 5-column grid option as well, however i don't think we'll use that one much - is that ok?

also also - i tried synching changes and it brought over a ton of changes so apologies for all the increased file changes it brought over! not sure why it did that!

mirnawong1:
actually @JKarlavige i'm going to close this pr because of those weird changes that came over, and i will create a new one to replace this. i hope that's ok but i just don't want to merge anything weird!

mirnawong1:
will delete branch if jason is ok with me recreating a new pr

JKarlavige:
> will delete branch if jason is ok with me recreating a new pr

Good with me, feel free to delete this branch!

PR #3394, opened 2023-05-18
adding ability to have 4 or 5 columns in cards

mirnawong1:
re-adding this css pull request to allow 4 or 5 grid columns per @JKarlavige ’s amazing guidance. This would most likely work visually for cards that have very little text and no toc sidebar.

[see slack thread](https://dbt-labs.slack.com/archives/C01C2H652BC/p1684250667489089)

closed initial pr https://github.com/dbt-labs/docs.getdbt.com/pull/3377 because of weird changes that came over from synching

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -1772,7 +1772,7 @@ section > h2:not(.resource-section) {
 
 @media(min-width: 996px) {
   .grid--5-col {
-    grid-template-columns: repeat(5, 1fr);
+    grid-template-columns: repeat(2, 1fr);
     grid-gap: calc(var(--ifm-spacing-horizontal) * 2);
   }
   @media(min-width: 1500px) {
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -1764,6 +1764,44 @@ section > h2:not(.resource-section) {
 
 
 /* utilities */
+.grid--5-col {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
+  grid-gap: var(--ifm-spacing-horizontal)
+}
+
+@media(min-width: 996px) {
+  .grid--5-col {
+    grid-template-columns: repeat(5, 1fr);
+    grid-gap: calc(var(--ifm-spacing-horizontal) * 2);
+  }
+  @media(min-width: 1500px) {
+    .grid--5-col {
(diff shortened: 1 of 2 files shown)


mirnawong1:
Hey @JKarlavige , i've opened this new pr which updates the custom CSS file to add 4 and 5 column grids for cards. it looks like these are best used with little text AND no rightside toc contents (they must be hidden). This is because there's card overlap (see screenshot 1) - should i do anything else here? 

I've also tested it with browser or mobile size and it does adjust (screenshot 2 and 3). please let me know what you think!


Screenshot 2023-05-18 at 13 12 10
Screenshot 2023-05-18 at 13 12 54
Screenshot 2023-05-18 at 13 15 46

website/src/css/custom.css
```diff
@@ -1764,6 +1764,44 @@ section > h2:not(.resource-section) {
...
 /* utilities */
+.grid--5-col {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
+  grid-gap: var(--ifm-spacing-horizontal)
+}
+
+@media(min-width: 996px) {
+  .grid--5-col {
+    grid-template-columns: repeat(5, 1fr);
```
mirnawong1:
```suggestion
    grid-template-columns: repeat(2, 1fr);
```

JKarlavige (approved):
lgtm! :shipit: I would definitely recommend using the 5 column grid very sparingly, as it slightly escapes the container on desktop. I don't see that as an issue however as the cards collapse before going out of view, so having them slightly outside the container doesn't create any noticeable issues. 

Tested both locally and verified media queries work correctly and cards collapse to 2 per row then 1 per row for mobile.

image

JKarlavige (approved):
lgtm! :shipit: I would definitely recommend using the 5 column grid very sparingly, as it slightly escapes the container on desktop. I don't see that as an issue however as the cards collapse before going out of view, so having them slightly outside the container doesn't create any noticeable issues. 

Tested both locally and verified media queries work correctly and cards collapse to 2 per row then 1 per row for mobile.

image

mirnawong1:
awesome! thanks sooo much @JKarlavige for your guidance on this 🙏 - totally agree about the 5 columns - probably very rarely needed.

PR #3419, opened 2023-05-24
Bug/versioned toc

breezyfasano:

Re:[this Asana task](https://app.asana.com/0/1200099998847559/1204106336641335/f), this PR resolves a glitch in the way the Table of Contents data was generated, which caused the TOC to be incorrect when a user would refresh on a page with versioned sections.

[Explanation of bug here.](https://github.com/dbt-labs/docs.getdbt.com/pull/3419/files#r1204719433)

## To test
- Open a docs page which includes versioned content, such as [this page](https://deploy-preview-3419--docs-getdbt-com.netlify.app/reference/node-selection/methods), focusing on the The "file" or "fqn" method section.
- Switch the version in the nav to v1.1 or lower.
- Refresh the page
    - [**On production,**](https://docs.getdbt.com/reference/node-selection/methods) The `The "file" or "fqn" method` item remains in the ToC on the right, even when the section is not on the page. This section is only available on v1.2 or greater.
    - [**On the preview of this PR,**](https://deploy-preview-3419--docs-getdbt-com.netlify.app/reference/node-selection/methods) the `The "file" or "fqn" method` section does not appear in the TOC on page refresh.

## Checklist


website/src/theme/DocItem/Layout/index.js
```diff
@@ -42,74 +42,77 @@ function useDocTOC() {
...
+          return {
+            value: heading.innerHTML,
+            id: heading.id,
+            level: level && level
+          }
+        }
 
-        // If updated toc different than current
-        // If so, show loader and update toc 
-        if(currentToc.length !== updated.length) {
```
breezyfasano, line 94: This logic was flawed! 
Here, we were comparing the `length` of the heading data of the page needing to be added to the TOC (`updated`, which equals an array from all the headings on the page) to the `length` of the TOC data currently being used to render the TOC component `currentToc`.

In some cases, even though `currentToc` and `updated` were different content-wise, they could have the same `length`. `updated` has ALL the headings including `h1`s, but Docusaurus only accepts headings with a `level` of 2 or 3 (`h2`s and `h3`s only).

When refreshing a page, I'm taking a guess that Docusaurus's initial value of currentToc (which is pulled from docusaurus's `toc` variable) is the TOC of whatever the newest version of dbt is. 
This might be because we are not using their versioning but our own custom implementation of versioning. So in the case of the `The "file" or "fqn" method` section on `/reference/node-selection/methods`, it was checking the `currentToc` with a length of `15` to `updated`, which was missing the `The "file" or "fqn" method` data BUT included the `h1` data for `Methods` which also made the length `15`.

So, by strict checking the content itself with `if(currentToc !== updated)`, this check will trigger setting the `currentToc` state to whatever `current` is even on page refresh, instead of doing nothing and displaying incorrect TOC data.

lfracalossiFA:
I do notice that on refresh there is a flash of the old `fqn` heading. I'm assuming that's unavoidable, since the content exists in the DOM and JS needs to run in order to remove/hide it?

lfracalossiFA (approved):
This lgtm, but John is more knowledgeable with React than I am, so I'd value his feedback, too!

website/src/theme/DocItem/Layout/index.js
john-rock: Nice discovery and refactor of that logic, makes sense.

john-rock (approved):
lgtm!

john-rock:
> I do notice that on refresh there is a flash of the old `fqn` heading. I'm assuming that's unavoidable, since the content exists in the DOM and JS needs to run in order to remove/hide it?

Yeah unfortunately I don't see an elegant way to prevent that with our current setup.

breezyfasano:
Thanks so much @john-rock and @lfracalossiFA !! Sadly, we don't have much control over the flash of content like John mentioned. I wish we did!

Issue #3330, opened 2023-05-03
1.5 - Missing new config for Snowflake for the `profiles.yml` set up


b-per:
### Contributions

- [X] I have read the contribution docs, and understand what's expected of me.

### Link to the page on docs.getdbt.com requiring updates

https://docs.getdbt.com/reference/warehouse-setups/snowflake-setup

### What part(s) of the page would you like to see updated?

For `Key Pair Authentication`, the doc only mentions `private_key_path` and `private_key_passphrase`. 

[This PR](https://github.com/dbt-labs/dbt-snowflake/pull/292) on the dbt-snowflake adapter side added the ability from 1.5 onwards to pick between `private_key_path` (the path to the private key), and `private_key` (the value of the private key). Only one of the 2 can be defined though.

dbeatty10:
Resolved by https://github.com/dbt-labs/docs.getdbt.com/pull/3432

PR #3432, opened 2023-05-29
Add docs about `private_key`

b-per:
Fix for #3330


website/docs/reference/warehouse-setups/snowflake-setup.md
```diff
@@ -124,6 +124,8 @@ Along with adding the `authenticator` parameter, be sure to run `alter account s
 
 To use key pair authentication, omit a `password` and instead provide a `private_key_path` and, optionally, a `private_key_passphrase` in your target. **Note:** Versions of dbt before 0.16.0 required that private keys were encrypted and a `private_key_passphrase` was provided. This behavior was changed in dbt v0.16.0.
 
+From dbt v1.5.0 onwards, it is possible to provide a `private_key` (Base64-encoded DER-encoded key bytes) string instead of a `private_key_path`.
```
mirnawong1: @b-per @dataders do we want to just version this section so it appears for 1.5 and higher? also do we prefer users to use the `private_key` instead of `private_key_path`?

I'm happy to make that change on my end, but I'm wondering if it's a better method so that the content doesn't go stale in the future

b-per: We can make it dependent on the version, but it will just require a bit more changes (and I added an example in the code block below as well)

There is no preferred way between `private_key` and `private_key_path`, but I have heard some people who wanted to use `private_key` because they were calling Snowflake as part of CI/CD and didn't want to copy the key to a file as it was against their security policies. Using `private_key` with a CI secret was OK with them.

mirnawong1: gotcha, sounds good! happy to make the version changed on my end to get your pr live @b-per  but will also wait for anders to come online and respond.

website/docs/docs/core/connect-data-platform/snowflake-setup.md
```diff
@@ -124,6 +124,8 @@ Along with adding the `authenticator` parameter, be sure to run `alter account s
 
 To use key pair authentication, omit a `password` and instead provide a `private_key_path` and, optionally, a `private_key_passphrase` in your target. **Note:** Versions of dbt before 0.16.0 required that private keys were encrypted and a `private_key_passphrase` was provided. This behavior was changed in dbt v0.16.0.
 
+From dbt v1.5.0 onwards, it is possible to provide a `private_key` (Base64-encoded DER-encoded key bytes) string instead of a `private_key_path`.
```
mirnawong1: hey @b-per , tweaked it slightly and also added a base64 link - does this read correctly, technically? also changed my mind about the versioning

```suggestion
Starting from [dbt v1.5.0](/docs/dbt-versions/core), you have the option to use a `private_key` string instead of a `private_key_path`. The `private_key` string should be in Base64-encoded DER format, representing the key bytes. Refer to [Snowflake documentation](https://docs.snowflake.com/developer-guide/python-connector/python-connector-example#using-key-pair-authentication-key-pair-rotation) for more info on how they generate the key.

```

b-per: Thanks for the update on the version.

I am not too sure about the link to Wikipedia for base64 though. This is a quick explanation of base64 but actually not the full format needed for the key in that case. This might be OK to link but in that case, I would also [link the page of the Snowflake docs](https://docs.snowflake.com/developer-guide/python-connector/python-connector-example#using-key-pair-authentication-key-pair-rotation) that show how this key is generated.

mirnawong1: sounds great! thank you for providing that link - ok made this change now and happy to approve and merge this for you ✨

mirnawong1:
hey @b-per going to merge this today, let me know if you have any objections!

mirnawong1 (approved):

b-per:
All good! Thanks 🙌

PR #3479, opened 2023-06-06
Fix error in service token / API auth docs

cafzal:
Fix error in service token / API auth docs
API is Discovery API, but permission is still "Metadata only"


mirnawong1 (approved):
approving @cafzal !

PR #3480, opened 2023-06-06
Fix error re auth in API docs

cafzal:
Fix error re auth in API docs
API is Discovery API, but permission is still "Metadata only"


mirnawong1 (approved):
looks good to me @cafzal , going to merge this for you!

PR #3523, opened 2023-06-14
Clarify temporary relation config

smitsrr:
I found it very confusing when I read the documentation regarding the temporary relation updates after 1.3. There are still many situations in which a 'table' is still used, even when the default is `view`, whereas this isn't stated explicitly in the documentation. 

I've attempted to clarify the point here. 

## Checklist

- [ ] Add an entry `website/static/_redirects`


website/docs/reference/resource-configs/snowflake-configs.md
```diff
@@ -303,9 +303,9 @@ models:
 
 ## Temporary Tables
 
-Beginning in dbt version 1.3, incremental table merges for Snowflake utilize a `view` rather than a `temporary table`. The reasoning was to avoid the database write step that a temporary table would initiate and save compile time. 
+Beginning in dbt version 1.3, incremental table merges for Snowflake may utilize a `view` rather than a `temporary table`. The reasoning was to avoid the database write step that a temporary table would initiate and save compile time.
```
joellabes:
```suggestion
Beginning in dbt version 1.3, incremental table merges for Snowflake prefer to utilize a `view` rather than a `temporary table`. The reasoning was to avoid the database write step that a temporary table would initiate and save compile time. 
```

website/docs/reference/resource-configs/snowflake-configs.md
```diff
@@ -303,9 +303,9 @@ models:
 
 ## Temporary Tables
 
-Beginning in dbt version 1.3, incremental table merges for Snowflake utilize a `view` rather than a `temporary table`. The reasoning was to avoid the database write step that a temporary table would initiate and save compile time. 
+Beginning in dbt version 1.3, incremental table merges for Snowflake may utilize a `view` rather than a `temporary table`. The reasoning was to avoid the database write step that a temporary table would initiate and save compile time. 
 
-However, many situations remain where a temporary table would achieve results faster. dbt v1.4 adds the `tmp_relation_type` configuration to leverage temporary tables for incremental builds. This is defined as part of the model configuration.
+However, many situations remain where a temporary table would achieve results faster and is used. dbt v1.4 adds the `tmp_relation_type` configuration to leverage temporary tables for incremental builds. This is defined as part of the model configuration. However, you may see an error if an incremental model has `delete+insert` strategy with a unique key declared. In this instance the model must be configured with a `tmp_relation_type: table`.
```
joellabes:
```suggestion
However, some situations remain where a temporary table would achieve results faster or more safely. dbt v1.4 adds the `tmp_relation_type` configuration to allow you to opt in to temporary tables for incremental builds. This is defined as part of the model configuration. 

To guarantee accuracy, an incremental model using the `delete+insert` strategy with a `unique_key` defined requires a temporary table; trying to change this to a view will result in an error.
```

I reworked this a bit, because I don't think it's correct to say that the model _needs to be_ explicitly configured. If I'm [interpreting the linked code](https://github.com/dbt-labs/dbt-snowflake/blob/ed6ed81ec2f8b409b29dafe11ed47a75b477ffec/dbt/include/snowflake/macros/materializations/incremental.sql#L35-L57) from your Slack message correctly, it looks like it can either be left unset _or_ `table` must be set (which is redundant since that will happen anyway).

joellabes (changes requested):
Hey @smitsrr, thanks for opening this! @mirnawong1 asked me to take a peek. I haven't used this setting so I might have misinterpreted it, but from the code you linked I'm surprised to see you were getting an error unless you were specifying `view` as the type. 

Let me know if these suggested changes work for you - if this isn't what you're observing, I'm curious about whether you've actually found a bug in dbt Core!

website/docs/reference/resource-configs/snowflake-configs.md
smitsrr: This is correct, but I also am trying to add the detail that some configuration settings are forbidden and raise the compiler error [here](https://github.com/dbt-labs/dbt-snowflake/blob/ed6ed81ec2f8b409b29dafe11ed47a75b477ffec/dbt/include/snowflake/macros/materializations/incremental.sql#L36). 

I was able to generate the compilation error in my environment with the following configurations: 
```
{{
    config(
        materialized='incremental',
        unique_key ='zendesk_user_id',
        incremental_strategy='delete+insert',
        tmp_relation_type='view'
    )
}}
```
image

smitsrr: My main goal of the PR is to clarify the wording a bit since when I first read it I was surprised my environment was still using tmp tables, but in fact, tables are still used quite often based on the configurations. 

Thanks for your feedback!

joellabes (approved):
Thank you for calling this out @smitsrr!

PR #3531, opened 2023-06-14
Add slash to fix broken link

runleonarun:
without the slash, the link assumes the same directory, but this file is not under `/build` it's under `/reference`.


mirnawong1 (approved):
lgtm!

PR #3543, opened 2023-06-15
RSS Feed Publish Date Adjustment

JKarlavige:
This adjusts the release notes RSS feed plugin to get the correct date when date is added as a frontmatter `tag` with the format: `<month>-<year>`. 

Currently, the feed is not correctly picking up the date from the `tags` field, and defaults to today's date.

## Preview
In the feed preview below, verify the `pubDate` field for each item in the feed matches the correct directory for each release note.

For example, the items in the RSS feed under the `May-2023` [release notes](https://github.com/dbt-labs/docs.getdbt.com/tree/current/website/docs/docs/dbt-versions/release-notes/08-May-2023) now correctly show a `pubDate` of May (defaults to May 1st).

image

https://deploy-preview-3543--docs-getdbt-com.netlify.app/feeds/rss.xml

The current RSS feed which incorrectly shows today's date for each item in the feed:
https://docs.getdbt.com/feeds/rss.xml


runleonarun (approved):
Hey @JKarlavige I'm so glad you're fixing this! I was going to give feedback about it being the last day of the month, but you've fixed that already!

We will always add descriptions to improve the output as well! This will replace the "No comment" messages under some release notes in the feed.

john-rock (approved):
lgtm :shipit:

PR #3594, opened 2023-06-21
Correct example query in modelbyenvironment discovery API schema

cafzal:
Correcting example queries so they work as copy and paste for users


runleonarun (approved):
I couldn't get this to work, but approving in case it's just me!

PR #3595, opened 2023-06-21
Correct example queries in discovery api use case guide

cafzal:
Correcting example queries so they work as copy and paste for users


runleonarun (approved):
Looks good!

PR #3615, opened 2023-06-26
Display job title and org in more places in the devblog

joellabes:
After the [discussion here](https://dbt-labs.slack.com/archives/C0101LM7Z89/p1687328973338709), we realised that authors' job titles and companies don't display on blog posts, only on their author bio pages. 

I poked around a bit and it looks like the blog views were relying on a property called `title`, which as far as I can see doesn't exist. It's possible that it's meant to be magically concatenated behind the scenes in a way that I don't understand, but I suspect it was actually just something that got renamed partway through development and wasn't noticed. 

To resolve this, I copied the logic from [website/src/components/author/index.js](https://github.com/dbt-labs/docs.getdbt.com/blob/7ec373e32ef3f1023a88f194d0a519a07748b50d/website/src/components/author/index.js#L46) and removed the unhelpful `title`. 

Before:
image

After: 
image

### DIFF
--- a/website/src/theme/BlogPostItem/Header/Author/index.js
+++ b/website/src/theme/BlogPostItem/Header/Author/index.js
@@ -15,7 +15,7 @@ function MaybeLink(props) {
 */
 
 export default function BlogPostItemHeaderAuthor({author, className}) {
-  const {name, title, url, imageURL, email, key} = author;
+  const {name, url, imageURL, email, key, job_title, organization} = author;
   const link = url || (email && `mailto:${email}`) || undefined;
   return (
     <div className={clsx('avatar margin-bottom--sm', className)}>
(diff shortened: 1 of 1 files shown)


mirnawong1:
lgtm but following to learn and see what frontend say 🙏

john-rock (approved):
lgtm! I'm not sure when that was changed but confirming from the authors.yml file the property is in fact` job_title`.

Edit: Actually it looks like the `title` property was updated to `job_title` on [this commit](https://github.com/dbt-labs/docs.getdbt.com/commit/bace0dde87a78e0d9b4c90f9916c2b9588768628) from Oct 19, 2021. 

```
andrew_escay:
  image_url: /img/blog/authors/aescay.jpeg
  job_title: Analytics Engineer
  links:
    - icon: fa-twitter
      url: https://twitter.com/_aescay
  name: Andrew Escay
  organization: dbt Labs
```

PR #3628, opened 2023-06-27
rename file and alter header

joellabes:
- Forgot to change the file name when i renamed from slack to community rules
- doesn't make sense to say "let the community speak first" when vendors are also community members


runleonarun (approved):
👍🏻

PR #3652, opened 2023-06-30
Update post-hook for "latest" model version

jtcohen6:
This is still just a placeholder "YMMV" solution until we decide on a real implementation:
- https://github.com/dbt-labs/dbt-core/issues/7442

---

If you're versioning `some_model` for the first time:
- that model is materialized as a `table`
- `my_db.my_schema.some_model` will exist as a table in the database
- needs to be dropped before running `create or replace view`

(NOTE: I need to actually test this code)


joellabes (approved):
This just worked on my branch and looks good, so let's do it :shipit:

PR #3658, opened 2023-06-30
Update meta.md

leoebfolsom:

I could be completely wrong, but I didn't think `config:` was needed in the yaml.


mirnawong1:
hey @leoebfolsom , thanks so much for opening this up! I'll take a look at this for you ✨

nghi-ly:
hi @leoebfolsom , thanks for flagging! looks like `config` is needed but, you're right, there is some inconsistency here; this is captured in Issue #3710 so the docs can be updated for a better explanation.

PR #3679, opened 2023-07-05
Update SQL example SL getting started

Jstein77:
Update SQL example in SL getting started guide.


mirnawong1 (approved):
lgtm !

PR #1721, opened 2022-07-15
Update cents_to_dollars macro

alexphilion:

PR #3708, opened 2023-07-11
Fix cents_to_dollars macro example

nghi-ly:

Reviewing PR #1721. The `alexphilion` repo has been deleted so opening this PR to incorporate their changes. 

Looking at the docs on https://docs.getdbt.com/reference/resource-properties/argument-type , it should be `scale` instead of `precision`. 

tysm @alexphilion for your contibution to the dbt docs! 🙏


runleonarun (approved):
Looks good in the output!

PR #3715, opened 2023-07-12
Semantic Model attribute is `defaults` not `default`

jtcohen6:
https://github.com/dbt-labs/dbt-semantic-interfaces/blob/229a417070b22326a843fe7a2f1fcf2fed195dcf/dbt_semantic_interfaces/validations/semantic_models.py#L200C27-L201

Following the docs in a toy project & setting this to `default` raises this error:

```
Parsing Error
  Invalid semantic_models config given in FilePath(searched_path='models', relative_path='something.yml', modification_time=1689170172.2842774, project_root='/Users/jerco/dev/scratch/testy') @ semantic_models: {'name': 'transaction', 'model': "ref('fact_transactions')", 'description': 'Transaction fact table at the transaction level. This table contains one row per transaction and includes the transaction timestamp.', 'default': {'agg_time_dimension': 'transaction_date'}, 'entities': [{'name': 'transaction', 'type': 'primary', 'expr': 'transaction_id'}, {'name': 'customer', 'type': 'foreign', 'expr': 'customer_id'}], 'dimensions': [{'name': 'transaction_date', 'type': 'time', 'type_params': {'time_granularity': 'day'}}, {'name': 'transaction_location', 'type': 'categorical', 'expr': 'order_country'}], 'measures': [{'name': 'transaction_total', 'description': 'The total value of the transaction.', 'agg': 'sum'}, {'name': 'sales', 'description': 'The total sale of the transaction.', 'agg': 'sum', 'expr': 'transaction_total'}, {'name': 'median_sales', 'description': 'The median sale of the transaction.', 'agg': 'median', 'expr': 'transaction_total'}]} - at path []: Additional properties are not allowed ('default' was unexpected)
```

## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) and [About versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) so my content adheres to these guidelines.
- [ ] Add a checklist item for anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

Adding new pages (delete if not applicable):
- [ ] Add page to `website/sidebars.js`
- [ ] Provide a unique filename for the new page

Removing or renaming existing pages (delete if not applicable):
- [ ] Remove page from `website/sidebars.js`
- [ ] Add an entry `website/static/_redirects`
[...]

PR #3717, opened 2023-07-12
change semantic model attribute to defaults

mirnawong1:
following up from pr #3715, changing the rest of the docs from `default` to `defaults` cc @jcohen6


runleonarun (approved):
Great catches!

PR #3743, opened 2023-07-14
replace screenshot with code block

mirnawong1:
replacing screenshot with code block to improve accessibility and ux.

also removed image file


website/docs/reference/commands/compile.md
```diff
@@ -33,13 +33,42 @@ dbt compile --select stg_payments
 dbt compile --inline "select * from {{ ref('raw_orders') }}"
 ```
 
-<Lightbox src="/img/docs/reference/dbt-compile.png" title="dbt compile --select stg_payments"/>
+returns the following:
+
+
+```bash
+dbt compile --select stg_orders
```
mirnawong1, line 40: should this be ` --select stg_payments`? instead per example above this section? 

```
dbt compile --select stg_payments
dbt compile --inline "select * from {{ ref('raw_orders') }}"

```

mirnawong1: so should the example change to `stg_orders`?

joellabes: Having the basic demo and the worked example reference the same things isn't the worst idea! Feel free to change it 👍

joellabes (approved):

mirnawong1:
thanks for the approval @joellabes ! i think i understand it a bit more 🙏

PR #3776, opened 2023-07-19
Update auto-update config

nghi-ly:

Currently, auto update is a bit noisy with all its messages in PRs. Experimenting with our GH auto-update config so hopefully its less noisy, particularly for long-running PRs

## Checklist

- [x] Needs review from frontend eng team

### DIFF
--- a/.github/workflows/autoupdate.yml
+++ b/.github/workflows/autoupdate.yml
@@ -4,9 +4,9 @@ on:
   # This will trigger on all pushes to all branches.
   push: {}
   # Alternatively, you can only trigger if commits are pushed to certain branches, e.g.:
-  # push:
-  #   branches:
-  #     - current
+  push:
+    branches:
+      - current
   #     - unstable
 jobs:
   autoupdate:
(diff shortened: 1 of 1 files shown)


.github/workflows/autoupdate.yml
```diff
@@ -4,9 +4,9 @@ on:
   # This will trigger on all pushes to all branches.
   push: {}
```
john-rock: We can't have two `push` commands let's comment or remove this one so it doesn't conflict with the line 7 `push`.

john-rock (changes requested):
Line 5 just needs to be removed or commented and then I can approve!

.github/workflows/autoupdate.yml
nghi-ly: oops, i missed that! great catch!

nghi-ly:
```suggestion
#  push: {}
```

nghi-ly:
made the change, @john-rock . can i get a re-review, please? 🙏

john-rock (approved):
lgtm! Let's 🚢 and monitor the results to make sure we are seeing expected behavior and decreased noise.

PR #3795, opened 2023-07-21
add specs + timespine

Jstein77:
Add BQ timespine example and add specs for entities, dimensions and semantic models.


mirnawong1 (approved):
thanks so much @Jstein77 , committed a few things so the content also had a table of the required/optional parameters, and also other indention fixes. but merging for you!✨🚢

PR #3814, opened 2023-07-25
Update grants.md with grant syntax note

alison985:
Adds a note to avoid some frustration when using grants in a dbt_project.yml file.

## Checklist


mirnawong1:
hey @alison985 , thanks so much for opening this pr up! I'm curious why the `+` is required though. The `+` in +grants should be the same `+` as any config set in dbt_project.yml. It should work with and without.

alison985:
@mirnawong1 I _think_ it has to do with mistaking the word grants for a folder and the parser not being happy when it doesn't find that folder. Or put another way, in theory, you could have a /models/grants/ folder so without the `+` it doesn't know what to do when it doesn't find that folder.

Without `grants` and its config lines the `dbt parse` command passes. Add `grants` information and it fails. The error is:

```bash
Compilation Error
  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
  There are 1 unused configuration paths:
  - models.PROJECT_NAME.FOLDER_A.SUB_FOLDER_B.grants
```

Add the `+` to make it `+grants` and it passes. 

Since you control so many things by folder, you have to be able to do grants to specific folders too.

mirnawong1:
hey @alison985 ah ok, thanks so much for that insight! I've checked in with the team internally and we'd love to look at this further since it's not something we expect to happen. 

can you let us know what dbt version you're on? cc @dbeatty10, not sure if this is a dbt core bug but I've initially checked with @b-per on this and it may need more looking into - is there another place we should raise this?

alison985:
This was dbt v1.4.6 in Docker to SQL Server.

dbeatty10:
This doesn't look like a bug to me -- this looks like [expected behavior](https://docs.getdbt.com/reference/resource-configs/grants#grant-config-inheritance) (for the reasons @alison985 described [here](https://github.com/dbt-labs/docs.getdbt.com/pull/3814#issuecomment-1651957312)).

i.e., the `+` is required within `dbt_project.yml` (or else it will be interpreted as a folder name):
image

Whereas within a model `config`, there is no `+`:

image

b-per:
My expectation is that `+` is optional in `dbt_project.yml` and only required if there is a folder called `<config_object>`. 

The following error message mentions that dbt is looking for `grants` as a folder but this folder doesn't exist. If the `grants` folder doesn't exist I would expect dbt to consider `grants` as the config point even though there is no `+`.
```
Compilation Error
  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
  There are 1 unused configuration paths:
  - models.PROJECT_NAME.FOLDER_A.SUB_FOLDER_B.grants
```

dbeatty10:
Thanks for that insight @b-per 🧠 -- opened https://github.com/dbt-labs/dbt-core/issues/8315 as a result 👍 

In the meantime, I'd feel comfortable with the docs update proposed in this PR which adds [this comment](https://deploy-preview-3814--docs-getdbt-com.netlify.app/reference/resource-configs/grants#grant-config-inheritance):

image

website/docs/reference/resource-configs/grants.md
```diff
@@ -121,7 +121,7 @@ For example:
 
 ```yml
 models:
-  +grants:
+  +grants: #In this case the + is not optional, you must include it for your project to parse.
```
dbeatty10: Nit: adding more spaces in the comment

```suggestion
  +grants:  # In this case the + is not optional, you must include it for your project to parse.
```

dbeatty10 (approved):

PR #3851, opened 2023-08-02
Update 'How we structure' for SL

gwenwindflower:
Adds a page to the 'How we structure' guide covering our recommendations re the Semantic Layer.


website/docs/guides/best-practices/how-we-structure/5-semantic-layer-marts.md
```diff
@@ -0,0 +1,48 @@
...
+    └── stg_supplies.yml
+```
+
+## When to make a mart
+
+- ❓ If we can go directly to staging models and it's better to serve normalized models to the Semantic Layer, then when, where, and why would we make a mart?
+  - 🕰️ We have models that have measures but no time dimension to aggregate against. The details of this are laid out in the [Semantic Layer guide](https://docs.getdbt.com/guides/best-practices/how-we-build-our-metrics/semantic-layer-1-intro) but in short, we need a time dimension to aggregate against in MetricFlow. Dimensional tables that
+  - 🧱 We want to **materialize** our model in various ways.
+  - 👯 We want to **version** our model.
+  - 🛒 We have various related models that make more sense as **one wider component**.
```
Jstein77, line 45: What's an example here?

website/docs/guides/best-practices/how-we-structure/5-semantic-layer-marts.md
```diff
@@ -0,0 +1,48 @@
...
+  - 1️⃣ We have similar models across multiple data sources that make more sense **unioned together**.
+  - ⌚ We have models in our project we **need to time to refactor** but want to serve up to the Semantic Layer quickly.
```
Jstein77, line 47: This is going to be very common

Jstein77 (approved):
LGTM!

website/docs/guides/best-practices/how-we-structure/5-semantic-layer-marts.md
gwenwindflower: honestly this is more of a catchall for other situations like the order_items/products/supplies through join thing where maybe you don't have the specific constraints around measures and time dimensions but it still just makes sense to join them up into one concept because, say, `order_item` is the most granular the business ever gets or something like that

PR #3861, opened 2023-08-04
Update Cloud support for retry with additional nuance

jtcohen6:

- `dbt retry` is supported in the dbt Cloud IDE
- Clarify that scheduled runs is what's in progress / coming soon

### DIFF
--- a/website/docs/reference/dbt-commands.md
+++ b/website/docs/reference/dbt-commands.md
@@ -53,7 +53,7 @@ Use the following dbt commands in the [dbt Cloud IDE](/docs/cloud/dbt-cloud-ide/
 - [compile](/reference/commands/compile): compiles (but does not run) the models in a project
 - [deps](/reference/commands/deps): downloads dependencies for a project
 - [docs](/reference/commands/cmd-docs) : generates documentation for a project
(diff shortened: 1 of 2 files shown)


website/docs/reference/dbt-commands.md
```diff
@@ -53,6 +53,7 @@ Use the following dbt commands in the [dbt Cloud IDE](/docs/cloud/dbt-cloud-ide/
 - [compile](/reference/commands/compile): compiles (but does not run) the models in a project
 - [deps](/reference/commands/deps): downloads dependencies for a project
 - [docs](/reference/commands/cmd-docs) : generates documentation for a project
+- [retry](/reference/commands/retry): retry the last run `dbt` command from the point of failure (requires dbt 1.6 or higher)
```
nghi-ly: i was trained to use later/older (vs higher/lower) when documenting version requirements. i think newer/older works too if you prefer 

```suggestion
- [retry](/reference/commands/retry): retry the last run `dbt` command from the point of failure (requires dbt 1.6 or later)
```

nghi-ly (commented):
thanks for updating this! left one small sugg

nghi-ly (approved):

PR #3884, opened 2023-08-08
Remove dbt updates tag from devblog

joellabes:
We experimented with product updates in the dev blog for a lil while, but it's now better implemented elsewhere. So that it doesn't look like we just stopped improving the product 18 months ago, we're going to remove this tag but keep the articles around so they don't 404. 

NB that a handful of new product feature articles were incorrectly tagged (using the label not its ID) so I've just removed those tags altogether as well.

Before and after: 

Screenshot 2023-08-08 at 4 12 03 PM
Screenshot 2023-08-08 at 4 12 43 PM

### DIFF
--- a/website/blog/2022-10-19-polyglot-dbt-python-dataframes-and-sql.md
+++ b/website/blog/2022-10-19-polyglot-dbt-python-dataframes-and-sql.md
@@ -4,7 +4,7 @@ description: "Going polyglot is a major next step in the journey of dbt Core. It
 slug: polyglot-dbt-python-dataframes-sql
 
 authors: [doug_beatty]
-
+tags: [dbt tutorials]
 hide_table_of_contents: false
 
 date: 2022-10-18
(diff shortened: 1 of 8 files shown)


website/blog/2022-10-19-polyglot-dbt-python-dataframes-and-sql.md
```diff
@@ -5,7 +5,6 @@ slug: polyglot-dbt-python-dataframes-sql
 
 authors: [doug_beatty]
 
-tags: [dbt product updates]
```
runleonarun, line 8: @joellabes should this be tagged `dbt tutorials` maybe?

runleonarun (approved):
One minor/non-blocking suggestion then 🚢 it!

website/blog/2022-10-19-polyglot-dbt-python-dataframes-and-sql.md
joellabes: works for me!

PR #3896, opened 2023-08-09
SL fast follows, query link, and clarify primary entity

mirnawong1:
this pr:
- adds the sql query syntax link in key places like the migration guide, get started guide as well as the avail integrations page
- clarifies language in get started pages
- updates typos
- adds contextual info wrt primary entity
other nits as well.


Jstein77 (approved):
LGTM!

PR #3935, opened 2023-08-16
ADAP-768: Update docs for materialized views/dynamic tables in v1.6

mikealfare:

- Pull common materialized view configuration into a central area to avoid repeating those configurations in each adapter
- Add materialized views as a fifth built-in materialization since they exist in `dbt-core`
- Add a Monitored Configuration Changes section to relevant adapters that corresponds to the new configuration item `on_configuration_change`; these items are almost always adapter-specific

## Checklist

- [x] Needs review by DX
- [x] Needs review by Partnerships


website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+### Materialized View
+
+The `materialized view` materialization allows the creation and maintenance of materialized views
+in the target database. This materialization makes use of the `on_configuration_change` config, which
+aligns with the incremental nature of the namesake database object. This setting tells dbt to attempt to
+make configuration changes directly to the object when possible, as opposed to completely recreating
+the object to implement the updated configuration. Using `dbt-postgres` as an example, indexes can
+be dropped and created on the materialized view without the need to recreate the materialized view itself.
+
+The `on_configuration_change` config has three settings:
```
mikealfare, line 116: This section currently only pertains to materialized views (and dynamic tables for Snowflake), but could presumably apply to other materializations in the future. Should this be its own section? If so, where is the right spot?

website/docs/reference/resource-configs/postgres-configs.md
```diff
@@ -113,66 +113,41 @@ models:
...
-{{
-  config(
-    materialized = 'materialized_view',
-    on_configuration_change = 'apply',
-  )
-}}
-```
+The Postgres adapter supports [materialized views](https://www.postgresql.org/docs/current/rules-materializedviews.html).
+Indexes are the only configuration that is specific to `dbt-postgres`.
+The remaining configuration follows the general [materialized view](/docs/build/materializations#Materialized-View) configuration.
```
mikealfare: Did I link to the `Materialized View` section of `materializations` correctly here?

website/docs/docs/build/materializations.md
Fleid: I'm good with burning that bridge when we get there ;)

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+- `continue` &mdash; allow runs to continue while also providing a warning that the object was left untouched
+  - *Note:* this could result in downstream failures as those models may depend on these unimplemented changes
+- `fail` &mdash; force the entire run to fail if a change is detected
+
+Materialized views are implemented following this "drop through" life cycle:
+1. If an object does not exist, create a materialized view
+2. If an object exists, other than a materialized view, that object is dropped and replaced with a materialized view
+3. If `--full-refresh` is supplied, replace the materialized view regardless of changes and the `on_configuration_change` setting
+4. If there are no configuration changes, refresh the materialized view
+5. At this point there are configuration changes, proceed according to the `on_configuration_change` setting
```
Fleid, line 128:
```suggestion
5. At this point if there are configuration changes, proceed according to the `on_configuration_change` setting
```

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+  - *Note:* if any individual configuration change requires a full refresh, a full refresh be performed in lieu of individual alter statements
+- `continue` &mdash; allow runs to continue while also providing a warning that the object was left untouched
+  - *Note:* this could result in downstream failures as those models may depend on these unimplemented changes
+- `fail` &mdash; force the entire run to fail if a change is detected
+
+Materialized views are implemented following this "drop through" life cycle:
+1. If an object does not exist, create a materialized view
+2. If an object exists, other than a materialized view, that object is dropped and replaced with a materialized view
+3. If `--full-refresh` is supplied, replace the materialized view regardless of changes and the `on_configuration_change` setting
+4. If there are no configuration changes, refresh the materialized view
```
Fleid, line 127: Aren't we bypassing refresh in certain cases?

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -104,54 +104,57 @@ models:
...
+
+Changes to `backup` will result in a full refresh. Redshift requires a materialized
+view to be dropped and recreated to apply a change to the `backup` setting.
+
+#### Auto Refresh
+
+The `auto_refresh` setting can be updated via an `ALTER` statement. This setting effectively toggles
+automatic refreshes on or off. The default setting for this config is off (`False`). If this
+is the only configuration change for the materialized view, dbt will choose to apply
+an `ALTER` statement instead of issuing a full refresh,
```
Fleid, line 139: Here we will refresh even when Auto Refresh is on, when doing a normal run with no configuration changes (continue), as advised by AWS.

website/docs/docs/build/materializations.md
mikealfare: I'm explicitly calling out that we have identified configuration changes, hence we proceed according to the `on_configuration_change` setting. I can reword that, but "if" would be misleading.

website/docs/docs/build/materializations.md
mikealfare: Potentially. We attempt to run refresh as the default (against original spec, but missed in the 1.6 release). However, the refresh macro may do nothing, for example if the platform doesn't support it. I'll look through each adapter to make sure we're actually refreshing the materialized view, and update this line if there are cases where we aren't.

website/docs/reference/resource-configs/redshift-configs.md
mikealfare: If `on_configuration_change = "continue"` and there no changes, yes, this should issue a manual refresh. This happens because we only get to the `on_configuration_change` setting if there are changes (which there are not in this scenario).

With that in mind, if there ***are*** configuration changes, and `on_configuration_change = "continue"`, should we also refresh there? We are not currently doing that; we currently do nothing and move on.

website/docs/docs/build/materializations.md
dataders: yeah. i think for 1.7 we should aim to have a standalone page for materialized views. but for now it can stay as a subsection of this page

website/docs/reference/resource-configs/postgres-configs.md
```diff
@@ -113,66 +113,41 @@ models:
 
 <VersionBlock firstVersion="1.6">
 
-## Materialized view
+## Materialized Views
```
dataders: i'm pretty sure the docs team's style guide only initialized the first word of a heading

```suggestion
## Materialized views
```

website/docs/reference/resource-configs/postgres-configs.md
dataders, line 1: any idea why the diff is offset here? hard for me to compare the revision

website/docs/reference/resource-configs/postgres-configs.md
dataders: i think heading anchors are lower case. the upper-case version didn't bring me to the heading we want
```suggestion
The remaining configuration follows the general [materialized view](/docs/build/materializations#materialized-view) configuration.
```

dataders (commented):
left some comments. while i can't directly diff the before and after, I know that any improvement here will help our users, so i'm happy ship this and refine in a future PR.

some questions:
1. what's the value of introducing the `Changelog` blocks in addition to the `VersionBlock`s?

website/docs/reference/resource-configs/postgres-configs.md
mikealfare: Ah ok. I was trying to figure that out by looking at the headings and it seemed mixed.

website/docs/reference/resource-configs/postgres-configs.md
mikealfare: Not sure. I added the changelog block below the header, but it doesn't seem to be able to figure that out for some reason.

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+5. At this point there are configuration changes, proceed according to the `on_configuration_change` setting
+
+Materialized views are a combination of a view and a table, and serve use cases similar to incremental models.
+
+* **Pros:**
+  * Materialized views combine the query performance of a table with the data freshness of a view
+  * Materialized views operate much like incremental materializations, however they are usually
+able to be refreshed on a regular cadence (depending on the database), forgoing the regular refresh
+required with incremental materializations
```
amychen1776:
```suggestion
able to be refreshed on a regular cadence (depending on the database), forgoing the regular dbt batch refresh
required with incremental materializations
```

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+  * `dbt run` on materialized views correspond to a code deployment, just like views
+* **Cons:**
+  * Due to the fact that materialized views are more complex database objects, they tend to have
```
amychen1776: What does this mean? What are the missing configuration options?  Can we name any specifics?

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+less configuration options
+  * Materialized views may not be supported by every database platform
+* **Advice:**
+    * Consider materialized views for use cases where incremental models are sufficient, 
+but you want to set the refresh to a schedule instead of running dbt
```
amychen1776:
```suggestion
but you would like the data platform to manage the incremental logic and refresh.
```

website/docs/reference/resource-configs/postgres-configs.md
```diff
@@ -113,66 +113,41 @@ models:
...
-<File name='my_model.sql'>
-
-```yaml
-
-{{ config(
-    materialized="table" # or any model type eg view, incremental
-) }}
+Swapping an already materialized model to a materialized view, and vice versa, is not supported.
+The workaround is to manually drop the existing materialization in the data warehouse prior to calling `dbt run`.
+Normally, re-running with the `--full-refresh` flag would resolve this, but not in this case.
```
amychen1776:
```suggestion
Running with `--full-refresh` flag will not work to drop the existing table or view and create the materialized view (and vice versa).
```

website/docs/reference/resource-configs/postgres-configs.md
```diff
@@ -113,66 +113,41 @@ models:
...
+This would only need to be done once as the existing object would then be a materialized view.
 
-```
-
-</File>
+For example, assume a table, `my_model`, has already been materialized to the underlying data platform via `dbt run`.
```
amychen1776:
```suggestion
For example,`my_model`, has already been materialized as a table in the underlying data platform via `dbt run`.
```

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -104,54 +104,57 @@ models:
...
-    materialized = 'materialized_view',
-    auto_refresh = False,
-    on_configuration_change = 'apply',
-  )
-}}
-```
+#### Dist
 
-</File>
+Changes to `dist` will result in a full refresh. Redshift requires a materialized view to be
```
amychen1776: Is this at next `dbt run`?

website/docs/reference/resource-configs/snowflake-configs.md
```diff
@@ -392,13 +404,31 @@ models:
 
 </File>
 
+### Monitored Configuration Changes
+
+The settings below are monitored for changes applicable to `on_configuration_change`.
+
+#### Target Lag
+
+Changes to `target_lag` can be applied via an `ALTER` statement. Refreshing is essentially
```
amychen1776:
```suggestion
Changes to `target_lag` can be applied by running an `ALTER` statement. Refreshing is essentially
```

mikealfare:
> left some comments. while i can't directly diff the before and after, I know that any improvement here will help our users, so i'm happy ship this and refine in a future PR.
> 
> some questions:
> 
>     1. what's the value of introducing the `Changelog` blocks in addition to the `VersionBlock`s?

Honestly, I saw that other places were doing that, so I added it. I think it's just an indicator to the reader that this feature started at a certain point in time. I have no problem removing them if they cause confusion.

website/docs/docs/build/materializations.md
mikealfare: The configuration options vary by platform. For instance, there's a "Limitations" section [here](https://docs.aws.amazon.com/redshift/latest/dg/materialized-view-create-sql-command.html) that talks about Redshift materialized views; of note, temporary tables and traditional views may not be used in the query. For BigQuery there's [this](https://cloud.google.com/bigquery/docs/materialized-views-create); of note, any outer joins are not allowed, nor are window functions. We link to these pages in the adapter specific docs, but I don't think we call out that there are limitations. I wouldn't want to duplicate what's on the platform's docs as that information might change, but we could at least call users' attention to the fact that these exist.

website/docs/reference/resource-configs/redshift-configs.md
mikealfare: Most likely. Though if the materialized view was manually created and then dbt run was run for the first time with that materialized view as a dbt model, it would initiate a full refresh there too. It's depends on whether the materialized view already exists and if the dist config in the dbt model is different from that of the materialized view in the database, not if this is the first or second time dbt is run. That sounds like semantics, but that's my answer. Regardless, if you think this is less confusing to our users by inserting that, I'll definitely make the change.

amychen1776: Ah okay - that's fine. The issue is `refresh` takes place in multiple flows.  How does this sound?

```suggestion
Changes to `dist` will result in a full refresh of the existing materialized view (applied at the time of the next `dbt run` of the model). Redshift requires a materialized view to be
```

mikealfare: > The issue is refresh takes place in multiple flows. How does this sound?

That's correct. It will happen in three scenarios:

- The user has requested it via `--full-refresh`
- There is an existing object that is not a materialized view
- There is an existing materialized view with different configuration, such that some piece of that different configuration requires a full refresh, and the user has supplied `on_configuration_change='apply'` (this is also the default, so omitting this setting behaves the same for this scenario)

website/docs/reference/resource-configs/redshift-configs.md
Fleid: I think we should in the future yes

website/docs/docs/build/materializations.md
amychen1776: Ah I see. I read this as dbt limitations so I think we should just call out that there might be data platform limitations but go to your own docs to show. Calling this out here associated limitations to dbt

website/docs/reference/resource-configs/redshift-configs.md
mikealfare: This is captured in a separate story and the related docs updates are part of that story. This won't be resolved here.

website/docs/reference/resource-configs/postgres-configs.md
mikealfare: I removed the changelog block and this is much clearer now. Since that's causing issues for git diff, let's add them in as a separate PR if we decide to do so.

website/docs/docs/build/materializations.md
mikealfare: @amychen1776 I updated the verbiage. Does this address your concerns? I still want to keep it in because it's one of the major factors to consider (e.g. you can't always swap incremental table code to materialized views); however, I tried to make it clear that it's a data platform limitation and not a dbt limitation.

Fleid (approved):

website/docs/docs/build/materializations.md
```diff
@@ -103,6 +104,45 @@ When using the `table` materialization, your model is rebuilt as a <Term id="tab
...
+3. If `--full-refresh` is supplied, replace the materialized view regardless of changes and the `on_configuration_change` setting
+4. If there are no configuration changes, refresh the materialized view
+5. At this point there are configuration changes, proceed according to the `on_configuration_change` setting
+
+Materialized views are a combination of a view and a table, and serve use cases similar to incremental models.
+
+* **Pros:**
+  * Materialized views combine the query performance of a table with the data freshness of a view
+  * Materialized views operate much like incremental materializations, however they are usually
+able to be refreshed on a regular cadence (depending on the database), forgoing the regular dbt batch refresh
```
amychen1776:
```suggestion
able to be refreshed without manual interference on a regular cadence (depending on the database), forgoing the regular dbt batch refresh
```

amychen1776 (approved):

mikealfare:
@dataders, it looks like I'm not able to merge this despite it being all green. Is that something you can do as a code owner?

amychen1776:
> @dataders, it looks like I'm not able to merge this despite it being all green. Is that something you can do as a code owner?

@mikealfare I think I might still be one so I merged it in for you.

mikealfare:
Thanks @amychen1776!

PR #3972, opened 2023-08-23
Fix rendering of `config-version: 2` (and its link)

dbeatty10:
[Preview](https://deploy-preview-3972--docs-getdbt-com.netlify.app/reference/resource-configs/plus-prefix)

The end of the second sentence on the ["Using the + prefix"](https://docs.getdbt.com/reference/resource-configs/plus-prefix) page is not rendering correctly, and this PR fixes it.

### Before

image

### After (option A)

I prefer this option since [config-version](https://docs.getdbt.com/reference/project-configs/config-version) is optional beginning in dbt v1.5:

image

### After (option B)

image

## Checklist


mirnawong1 (approved):
lgtm, thanks for fixing that up @dbeatty10 ! i don't feel strongly about the change in 'only' vs 'not' since both ways are clear to the user, i believe.

PR #4032, opened 2023-09-06
drop cloud-specifc requirements from Trusted adapter requirements

dataders:

prompted by the result of [a good convo in #db-duckdb](https://getdbt.slack.com/archives/C039D1J1LA2/p1693555610047379), we're relaxing these requirements, which stemmed from planning for adapters to be included in dbt Cloud. while they're in Core land? This is less important.


jasnonaz (commented):
Approved via conversation [here](https://dbt-labs.slack.com/archives/C05GJJMC56F/p1693947378081399?thread_ts=1693931594.129239&cid=C05GJJMC56F)

jasnonaz (approved):
🌞

PR #4038, opened 2023-09-07
add metric offset

mirnawong1:
updating derived metric offset info [per metricflow](https://github.com/dbt-labs/metricflow/pull/744) change. 

pr adds examples on offset_window and offset_to_grain

### DIFF
--- a/website/docs/docs/build/derived-metrics.md
+++ b/website/docs/docs/build/derived-metrics.md
@@ -22,7 +22,6 @@ In MetricFlow, derived metrics are metrics created by defining an expression usi
 | `alias` | Optional alias for the metric that you can use in the expr. | Optional |
 | `filter` | Optional filter to apply to the metric. | Optional |
 | `offset_window` | Set the period for the offset window, such as 1 month. This will return the value of the metric one month from the metric time. This can't be used with `offset_to_grain`. | Required |
-| `offset_to_grain` | Specifies the granularity or level of detail for the offset, such as a day. This means if you set your `offset_to_grain: day`, the offset is applied at the daily level. If you set it to "hour," it means the offset is at the hourly level. This can't be used with `offset_window`. | Required |
 
 The following displays the complete specification for derived metrics, along with an example.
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/derived-metrics.md
```diff
@@ -85,17 +86,97 @@ metrics:
...
+|   | 7030 | 2017-07-30 to 2017-07-02 |
+|   | 78 | 2017-07-01 |
+| Total  | 7438 | 2017-07-01 |
+
+3. Next, we calculate July's bookings with a 7-day offset. The following table displays the range of days that make up this monthly aggregation. Note that the month begins 7 days later (offset by 7 days) on 2017-07-24.
+
+|   | Orders | Metric_time |
+| - | ---- | -------- |
+|   | 329 | 2017-07-24 |
+|   | 6840 | 2017-07-23  to 2017-06-35 |
```
mirnawong1: hey @Jstein77  - this should be 2017-06-30 right? not 35?

website/docs/docs/build/derived-metrics.md
```diff
@@ -85,17 +86,97 @@ metrics:
...
+      metrics:
+        - name: bookings
+          alias: current_bookings
+        - name: bookings
+          offset_window: 7 days
+          alias: bookings_7_days_ago
+
+```
+
+**Using `offset_to_grain`**
```
mirnawong1: added offset_to_grain based on the [slack convo](https://dbt-labs.slack.com/archives/C050Z6UP4RK/p1694036670014879?thread_ts=1694035028.060429&cid=C050Z6UP4RK) but would like technical verification please? i think it would def be beneficial to add an example, even if it's not as common.

website/docs/docs/build/derived-metrics.md
```diff
@@ -21,7 +21,8 @@ In MetricFlow, derived metrics are metrics created by defining an expression usi
 | `metrics` |  The list of metrics used in the derived metrics. | Required  |
 | `alias` | Optional alias for the metric that you can use in the expr. | Optional |
 | `filter` | Optional filter to apply to the metric. | Optional |
-| `offset_window` | Set the period for the offset window, such as 1 month. This will return the value of the metric one month from the metric time. | Required |
+| `offset_window` | Set the period for the offset window, such as 1 month. This will return the value of the metric one month from the metric time. This can't be used with `offset_to_grain`. | Required |
+| `offset_to_grain` | Specifies the granularity or level of detail for the offset, such as a day. This means if you set your `offset_to_grain: day`, the offset is applied at the daily level. If you set it to "hour," it means the offset is at the hourly level. This can't be used with `offset_window`. | Required |
```
mirnawong1: add to table and also clarified they can't be used together [per pr here](https://github.com/dbt-labs/metricflow/pull/744/files#diff-ef677cc7ad90e1061403e18f214a05fcffd11419f99d094c6bce47be9934e31eL737) which states that. can you verify @courtneyholcomb ?

Jstein77: Can we take out `offset_to_grain` from the docs for now? I'll write up some example use cases when I'm back from OOO and we can add it back. 

I think this option was added to satisfy a really specific use case from a customer at Transform. We need to do a bit more testing internally to figure out the right situations to use this option before we document it.

mirnawong1: fair enough, i'll get rid 👌

website/docs/docs/build/derived-metrics.md
Jstein77: Per my earlier comment lets leave this out for now until we can add some examples.

website/docs/docs/build/derived-metrics.md
Jstein77: Yes! Fat fingered it. Thanks for catching.

Jstein77 (approved):
Added a few comments! Thanks for getting this up.

courtneyholcomb (approved):
Looks great!! Thank you!

PR #4050, opened 2023-09-08
Update builtins.md

### LINKED ISSUE #3802: Updating the builtin `ref`/`source` functions is most likely different in dbt 1.5+
### Contributions

https://docs.getdbt.com/reference/dbt-jinja-functions/builtins

The page provide examples on how to update the `ref` command but it doesn't support:
- cross project ref (coming in 1.6)
- model versions (from 1.5)

I have not tried it but I think that those 2 features would not work if someone copy pasted the code example in their project.

We should have separate code snippets for `<=1.4`, `1.5` and `>=1.6`

### Additional information

_No response_

mirnawong1:
adding 1.5+ and <=v1.5 versioning and explains macro 

Resolves #3802


website/docs/reference/dbt-jinja-functions/builtins.md
```diff
@@ -15,9 +15,43 @@ The `builtins` variable is a dictionary containing the following keys:
...
-The following macro overrides the `ref` method available in the model compilation context to return a [Relation](/reference/dbt-classes#relation) with the database name overriden to `dev`.
+<VersionBlock firstVersion="1.6">
+  
+From dbt v1.6 and higher, use the following macro to extract user-provided arguments, including <code>version</code>, and call the <code>builtins.ref()</code> function with either a single <code>modelname</code> argument or both <code>packagename</code> and <code>modelname</code> arguments, based on the number of positional arguments in <code>varargs</code>:
 
+<br /><br />
+
+  
+```jinja2
+jinja2.ext.do
```
jtcohen6: I don't think we should include this `jinja2.ext.do`, because a dbt user wouldn't need/want to include it when copy-pasting this code. Did you find that it was necessary to support the syntax highlighting? If so, I would just remove `jinja2` above as well.

website/docs/reference/dbt-jinja-functions/builtins.md
```diff
@@ -15,9 +15,43 @@ The `builtins` variable is a dictionary containing the following keys:
 
 ## Usage
 
-The following macro overrides the `ref` method available in the model compilation context to return a [Relation](/reference/dbt-classes#relation) with the database name overriden to `dev`.
+<VersionBlock firstVersion="1.6">
```
jtcohen6: This should actually be for versions v1.5+, and the older examples for <=1.4

website/docs/reference/dbt-jinja-functions/builtins.md
```diff
@@ -15,9 +15,43 @@ The `builtins` variable is a dictionary containing the following keys:
...
+    {% do return(builtins.ref(packagename, modelname, version=version)) %}
+{% else %}
+    {% do return(builtins.ref(modelname, version=version)) %}
+{% endif %}
+
+{% endmacro %}
 ```
+</VersionBlock>
+
+<VersionBlock lastVersion="1.5">
```
jtcohen6: The previous examples actually did something _different_ — either updating or excluding the `database` component of the returned `Relation` — whereas the example above is just the same as default built-in `ref`.

> return a `Relation` with the database name overridden to `dev`

I think we should update the example above to do the same:

```
-- call builtins.ref based on provided positional arguments
{% set rel = None %}
{% if packagename is not none %}
    {% set rel = return(builtins.ref(packagename, modelname, version=version)) %}
{% else %}
    {% set rel = return(builtins.ref(modelname, version=version)) %}
{% endif %}

-- finally, override the database name with "dev"
{% set newrel = rel.replace_path(database="dev") %}
{% do return(newrel) %}
```

website/docs/reference/dbt-jinja-functions/builtins.md
```diff
@@ -15,9 +15,43 @@ The `builtins` variable is a dictionary containing the following keys:
...
+
+From dbt v1.5 and lower, use the following macro to override the `ref` method available in the model compilation context to return a [Relation](/reference/dbt-classes#relation) with the database name overriden to `dev`:
```
jtcohen6: *dbt v1.4 and lower

I think this example macro just _never_ supported two-argument `ref`. I guess it was a lot less common in earlier versions, but still! Some of the complexity above arrived with the addition of `version`, but some of it really should have been there all along.

jtcohen6 (changes requested):
@mirnawong1 Thanks so much for spearheading this!

What do you think of some sort of callout that this whole `builtins` business is a pretty "advanced" use case? Users should be aware that they will need to maintain/update these overrides as they upgrade into the future.

website/docs/reference/dbt-jinja-functions/builtins.md
mirnawong1: ok gotcha, thanks for clarifying that @jtcohen6. i was getting an error when  it managed to help me pass the checks. now that I've removed it, it's passing 🤦‍♀️ so idk what happened on fri

mirnawong1:
hey @jtcohen6 , thank you for this awesome feedback! I've made changes based on your suggestions - [how does it look now?](https://deploy-preview-4050--docs-getdbt-com.netlify.app/reference/dbt-jinja-functions/builtins)

PR #4054, opened 2023-09-09
fix cte query

stavhaco:
does not referencing the right name


dataders (approved):
thanks @stavhaco

PR #4083, opened 2023-09-16
Split out the `file` and `fqn` selection methods into their own sections

dbeatty10:
### 🎩 Previews
- [The "file" method](https://docs-getdbt-com-git-dbeatty-split-file-fqn-dbt-labs.vercel.app/reference/node-selection/methods#the-file-method)
- [The "fqn" method](https://docs-getdbt-com-git-dbeatty-split-file-fqn-dbt-labs.vercel.app/reference/node-selection/methods#the-fqn-method)

The `file` and `fqn` selections methods are distinct from each other, but they've been documented within the same section. They deserve their own sections, just like the rest of the methods.

### History

The `fqn` method was undocumented for some amount of time, and #1851 added the initial documentation.

### On con
- If anyone has a hyperlink that is using the previous anchor, it will no longer go to that same section.

i.e., this will no longer exist:
- https://docs.getdbt.com/reference/node-selection/methods#the-file-or-fqn-method

Instead, it will be split into these two:
- https://docs.getdbt.com/reference/node-selection/methods#the-file-method
- https://docs.getdbt.com/reference/node-selection/methods#the-fqn-method

But #1851 would have had the same issue (but in reverse), so in a sense, we'd be just returning to the anchor that was in use previously.

Side note: there might be some way to override the value of the anchor if we wanted it to be different than the default.

My recommendation is to just leave this PR as-is (without messing with the auto-generated anchors).

## Checklist


mirnawong1 (approved):
thanks @dbeatty10 and def agree that both fqn and file should have their own sections so it's more explicit for users. 

in terms of redirects, unfortunately, we can't redirect from links that contain anchor/hash links.

PR #4095, opened 2023-09-19
Add .vercel to gitignore

JKarlavige:

Adds `.vercel` to both `.gitignore` files to prevent Vercel build files from being included in repo.

_Opened this PR to test if the `deploy/netlify` check was disabled, but it looks like it still ran. But might as well git this edit in place while we're at it._


john-rock (approved):
:shipit:

PR #4169, opened 2023-10-02
Update sl-graphql.md

mirnawong1:
fixing link


mirnawong1:
merging this now as its a small link fix

PR #4181, opened 2023-10-04
Update about-cloud-ide.md

dave-connors-3:

The IDE only uses dbt-rpc for dbt-core <1.5, and will not use it in the future, as the `dbt-rpc` package has been deprecated


mirnawong1 (approved):
great catch, thank you @dave-connors-3 ✨ will merge for you!

PR #4189, opened 2023-10-05
add coalesce banner to docs site

mirnawong1:
Updates banner to point to coalesce.

background color can be changed to 047377 if purple isn't preferred


JKarlavige (approved):
lgtm! :shipit:

PR #4319, opened 2023-10-24
adding klipfolio docs to SL integrations

mirnawong1:
adding klipfolio's docs and cards to SL available integrations page per Bobby's feedbac.


website/docs/docs/use-dbt-semantic-layer/avail-sl-integrations.md
```diff
@@ -25,7 +25,6 @@ import AvailIntegrations from '/snippets/_sl-partner-links.md';
 You can also integrate the following tools with the dbt Semantic Layer:
 - [Push.ai](https://docs.push.ai/semantic-layer-integrations/dbt-semantic-layer)
 - [Delphi](delphihq.com)
-- KlipFolio Power Metrics - Documentation coming soon
```
matthewshaver, line 28: Is there a reason this gets promoted to the cards but push.ai and Delphi remain footnotes? Should we move them up to cards as well?

mirnawong1:
yea @matthewshaver and i believe it's a product discussion with partnerships.

website/snippets/_sl-partner-links.md
```diff
@@ -28,6 +28,20 @@ The following tools integrate with the dbt Semantic Layer:
     </a>
   </div>
 
+<div className="card-container">
+  <Card
+    title="KlipFolio Power Metrics"
+    body="Refer to KlipFolio's documentation to learn how to connect, and deliver a streamlined metrics catalog to your business users."
+    link="https://support.klipfolio.com/hc/en-us/articles/18164546900759-PowerMetrics-Adding-dbt-Semantic-Layer-metrics"
+    icon="klipfolio"/>
```
matthewshaver, line 36: The dark mode icon is hard to see (black icon on dark background). Not a showstopper but might not be ideal from a branding perspective.

matthewshaver (approved):

website/snippets/_sl-partner-links.md
mirnawong1: totally agree and i played around it already but i couldn't do much with that unfortunately. this is because they only have [black or white](https://www.klipfolio.com/branding-guidelines/klipfolio-badge) so it would either be difficult in light mode or dark mode. the other option is just black which also isn't great

mirnawong1: the other icons from their branding page were too small so went with badges

PR #4378, opened 2023-10-31
Update materializations-guide-4-incremental-models.md

bcrant:
Match description with the IDs used in the table.


matthewshaver:
Great catch! Thank you @bcrant

matthewshaver (approved):

PR #4392, opened 2023-11-03
Remove incorrect dash under owner dict

joellabes:
Fixing an errant hyphen reported by @adinsmoor


adinsmoor (approved):
lgtm

PR #4402, opened 2023-11-06
remove callout and add tabs as examples

mirnawong1:
addressing @joellabes' request to remove old callouts for alias, database, and schema pages. also added example tabs at the beginning to help provide users with more example at the beginning of the doc


website/docs/reference/resource-configs/alias.md
```diff
@@ -4,10 +4,47 @@ description: "Read this guide to understand the alias configuration in dbt."
...
-:::caution Heads up!
-This is a work in progress document. While this configuration applies to multiple resource types, the documentation has only been written for seeds.
 
-:::
+<Tabs>
+<TabItem value="model" label="Model">
+
+Specify a custom alias for a model in your `dbt_project.yml` file. 
+
+For example, if you have a model that calculates `sales_metric`s and want to give it a more user-friendly alias, you can alias it like this:
```
joellabes: Metrics might be a confusing term to give here given the other sort of metrics dbt supports

website/docs/reference/resource-configs/alias.md
```diff
@@ -4,10 +4,47 @@ description: "Read this guide to understand the alias configuration in dbt."
 datatype: string
```
joellabes, line 4: These are too far away from edited content to make a suggestion directly, but: 
- what does that `resource_types` array do? Does it configure how the docs render somehow? 
- Let's update the `description` to something like "Aliasing a resource allows it to be created in the database with a custom name, instead of relying on the filename"

website/docs/reference/resource-configs/alias.md
```diff
@@ -4,10 +4,47 @@ description: "Read this guide to understand the alias configuration in dbt."
...
+  your_project:
+    product_categories:
+      +alias: categories_data
+```
+
+This would return the name `analytics.finance.categories_data` in the database.
+
+</File>
+</TabItem>
+</Tabs>
```
joellabes: It looks like snapshots and tests can also be aliased: 
- [Snapshots](https://github.com/dbt-labs/dbt-core/issues/4469) example
- Tests example (the longer-term options in [this comment](https://github.com/dbt-labs/dbt-core/issues/3626#issuecomment-886694083))

website/docs/reference/resource-configs/alias.md
```diff
@@ -4,10 +4,47 @@ description: "Read this guide to understand the alias configuration in dbt."
...
 
 ## Definition
```
joellabes, line 65: or test or snapshot

website/docs/reference/resource-configs/alias.md
```diff
@@ -25,9 +62,8 @@ To learn more about changing the way that dbt generates a relation's `identifier
 ## Usage
```
joellabes: Do you think we still need this section based on the examples you've provided above? I think we could remove it cos it's largely repeating itself?

website/docs/reference/resource-configs/alias.md
```diff
@@ -25,9 +62,8 @@ To learn more about changing the way that dbt generates a relation's `identifier
...
 
 ### Seeds
-Configure a seed's alias in your `dbt_project.yml` file.
 
-The seed at `seeds/country_codes.csv` will be built as a <Term id="table" /> named `country_mappings`.
+In this example, the seed at `seeds/country_codes.csv` will be built as a <Term id="table" /> named `country_mappings`.
```
joellabes: Although it is useful to make clear that the seed lives at `seeds/country_codes.csv` so maybe that should migrate up to the example file before deletion

website/docs/reference/resource-configs/database.md
```diff
@@ -5,10 +5,47 @@ datatype: string
 description: "Read this guide to understand the database configuration in dbt."
```
joellabes:
```suggestion
description: "Override the default database when dbt creates resources in your data platform."
```

website/docs/reference/resource-configs/database.md
```diff
@@ -5,10 +5,47 @@ datatype: string
...
 ---
 
-:::caution Heads up!
-This is a work in progress document. While this configuration applies to multiple resource types, the documentation has only been written for seeds.
+<Tabs>
+<TabItem value="model" label="Model">
```
joellabes, line 9: Is it possible to put tests into a custom database? I assume that's alongside store-failures behaviour? Inferring this from the resource_types array above.

website/docs/reference/resource-configs/schema.md
```diff
@@ -5,10 +5,66 @@ description: "Schema - Read this in-depth guide to learn about configurations in
 datatype: string
```
joellabes, line 5: description: "Override the default schema when dbt creates resources in your data platform."

website/docs/reference/resource-configs/schema.md
```diff
@@ -5,10 +5,66 @@ description: "Schema - Read this in-depth guide to learn about configurations in
...
 ---
 
-:::caution Heads up!
-This is a work in progress document. While this configuration applies to multiple resource types, the documentation has only been written for seeds.
+<Tabs>
```
joellabes, line 8: This one's great 🎉

joellabes (changes requested):
ty @mirnawong1! Added a few comments and tweaks but this is going to be so much better 😍

website/docs/reference/resource-configs/alias.md
mirnawong1: changed it to `sales_total`

```
models:
  your_project:
    sales_total:
      +alias: sales_dashboard
```

website/docs/reference/resource-configs/alias.md
mirnawong1: asked in the slack channel and updated desc

website/docs/reference/resource-configs/alias.md
```diff
@@ -4,10 +4,47 @@ description: "Read this guide to understand the alias configuration in dbt."
...
+```yml
+models:
+  your_project:
+    sales_metrics:
+      +alias: sales_dashboard
+```
+</File>
+
+This would return `analytics.finance.sales_dashboard` in the database, instead of the default `analytics.finance.sales_metrics`.
+</TabItem>
```
mirnawong1, line 27: or is it `analytics.finance.unique_order_id_test`? [this](https://docs.getdbt.com/reference/resource-properties/tests#custom-test-name) makes me think it's the _ one?

mirnawong1, line 27: do we add config or is that only for `properties.yml`?

website/docs/reference/resource-configs/alias.md
mirnawong1: i was going to do exactly that but then i wondered how many ppl might have the #usage anchor link? and since we can't redirect anchor links to another link, i stayed conservative 😭

website/docs/reference/resource-configs/database.md
mirnawong1: hmm good question! have askd in dbt help to confirm but i also did add an example just incase

mirnawong1:
hey @joellabes thank you so much for this! I've made changes to it and added items you suggested. i did ask about tests in custom database to confirm, but i also added an example just in case. also i have a question below for you if you can clarify because i see some do and some don't based on the issues you linked above. thank you!

- can you also use config block for all 4 resource types?

website/docs/reference/resource-configs/alias.md
```diff
@@ -40,3 +56,68 @@ seeds:
...
+```
+
+This would return the name `analytics.finance.updated_at_id` in the database.
+
+</File>
+</TabItem>
+
+<TabItem value="test" label="Tests">
+
+Configure a test's alias in your `dbt_project.yml` file.
```
joellabes: You can't create tests in a dbt_project.yml, this would be in a schema.yml file

website/docs/reference/resource-configs/alias.md
```diff
@@ -40,3 +56,68 @@ seeds:
...
+    config:
+      target_database: analytics
+      target_schema: finance
+      unique_key: id
+      strategy: timestamp
+      updated_at: updated_at
+      +alias: updated_at_id
+```
+
+This would return the name `analytics.finance.updated_at_id` in the database.
```
joellabes: It would, but that's a confusing example given that it's not related to updated_at. Maybe set the alias to `your_snapshot` or something that is related to the original name?

joellabes: Also the + isn't necessary here

website/docs/reference/resource-configs/alias.md
```diff
@@ -40,3 +56,68 @@ seeds:
...
+<File name='dbt_project.yml'>
+
+```yml
+models:
+  - name: orders
+    columns:
+      - name: order_id
+        tests:
+          - unique
+            +alias: unique_order_id_test
```
joellabes: The + isn't necessary here

website/docs/reference/resource-configs/alias.md
```diff
@@ -40,3 +56,68 @@ seeds:
...
+```
+
+This would return the name `analytics.finance.orders_order_id_unique_order_id_test` in the database.
```
joellabes: Pretty sure this is only when using `--store-failures`

website/docs/reference/resource-configs/database.md
```diff
@@ -2,38 +2,85 @@
...
 
 ```yml
 seeds:
-  +database: RAW
+  your_project:
+    product_categories:
+      +database: staging
+```
+
+This would result in the generated relation being located in the `staging` database, so the full relation name would be `staging.finance.product_categories_data`.
```
joellabes: Where did the `_data` suffix come from here?

joellabes (changes requested):
Looking better! A couple of tweaks on the net-new stuff again

joellabes:
> * can you also use config block for all 4 resource types?

Yes: 

```
models:
  - name: first_model
    config:
      alias: second_model
    columns:
      - name: id
        tests:
          - unique:
              config:
                alias: custom_test_name_when_storing_failures
```
```
snapshots:
  - name: first_snapshot
    config:
      alias: second_snapshot
```

website/docs/reference/resource-configs/alias.md
mirnawong1: thank you! wait is schema.yml the same as properties.yml? https://docs.getdbt.com/reference/configs-and-properties#schemayml-files

website/docs/reference/resource-configs/database.md
mirnawong1: oh fudge! i think i copied the <tabs> over from alias and didnt remove the _data bit

mirnawong1 (commented):
thank you @joellabes ! ok added your feedback now - can you re-review when you have a sec?

mirnawong1:
going to enable automerge in case it's all good to go

website/docs/reference/resource-configs/alias.md
joellabes: Yep!

joellabes (approved):

PR #4411, opened 2023-11-07
auto copy header link

mirnawong1:
This PR adds a headerLinkCopy function/script that will auto copy H2 and H3 header links when header or 🔗 icon is clicked.  The script also has a pop up confirming the link was auto copied and times out after 3 seconds.


website/src/css/custom.css
```diff
@@ -2034,6 +2034,68 @@ html[data-theme="dark"] .theme-doc-sidebar-container>div>button.button:hover {
...
+  padding: 10px;
+  border-radius: 5px;
+  z-index: 9999;
+}
+
+.close-button {
+  cursor: pointer;
+  margin-right: 10px;
+  font-size: 20px;
+}
```
JKarlavige: The `.close-button` class has `margin-right: 10px;`. In the popup, the `x` appears pretty close to the text to the left.

Rather than setting only right margin, I'd recommend setting both left and right margin using the css: `margin: 0 10px`. This is shorthand saying "set top/bottom margin to 0, and left/right margin to 10px".

website/docs/docs/collaborate/documentation.md
```diff
@@ -6,7 +6,7 @@ pagination_next: "docs/collaborate/build-and-view-your-docs"
 pagination_prev: null
 ---
 
-## Related documentation
+## Related documentation <Lifecycle status="Beta" />
```
mirnawong1:
```suggestion
## Related documentation
```

website/static/js/headerLinkCopy.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable */
+
+  // Get all the headers with anchor links. 
+  // The 'click' event worked over 'popstate' because click captures page triggers, as well as back/forward button triggers
+  window.addEventListener("click", copyHeader);
+  
+  // separating function from eventlistener to understand they are two separate things
+  function copyHeader () {
+    const headers = document.querySelectorAll("h2.anchor, h3.anchor");
+    console.log("click");
```
JKarlavige: Feel free to remove this console.log before merging!

website/static/js/headerLinkCopy.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable */
+
+  // Get all the headers with anchor links. 
+  // The 'click' event worked over 'popstate' because click captures page triggers, as well as back/forward button triggers
+  window.addEventListener("click", copyHeader);
```
JKarlavige, line 6: Seeing one scenario where this doesn't fire which may cause the copy function to not work on initial load. Here's a Loom video showing what I'm seeing: https://www.loom.com/share/8ca6256d333d4e669b10b8f500bbaa63?sid=3a49f567-8b8c-48f0-a5bf-4ec16a2086f2

Rather than only listening on `click`, we can update this part to listen for both `load` and `click`. That way if a url is entered in the browser and the page is loaded, a user can click a header immediately, without having to make a click first.

[This post](https://stackoverflow.com/questions/8796988/binding-multiple-events-to-a-listener-without-jquery) on stackoverflow has some hints on how to do this!

JKarlavige (changes requested):
Awesome work with this @mirnawong1! Just have one requested change. I shared a loom along with a hint on what to adjust. Let me know if you have any questions and we can walk through it!

website/static/js/headerLinkCopy.js
mirnawong1: ok cool! thank you @JKarlavige ! ok i added another evenlistener line focusing on 'load'. i also think i could've done created a function for this but i couldn't get it working.

mirnawong1:
hey @JKarlavige thank you so much! ok i've tested this out and added another eventlistener - it looks like this world in a page refresh too. can you give it another review whenever you have a chance. thank you!

website/static/js/headerLinkCopy.js
JKarlavige: This works great!! No need to update anything here or do things differently!

JKarlavige (approved):
all lgtm! great idea and awesome work again on this @mirnawong1! it even works in the blog which is pretty cool.

mirnawong1:
thank you so much for your guidance and help @JKarlavige ! 🚀

PR #4423, opened 2023-11-09
Update regions-ip-addresses.md

joellabes:
A typo in the regions table - `use-east-1` instead of `us-east-1`. 

`use-east-1` does not exist, [as shown here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html)


mirnawong1 (approved):
hot damn great catch @joellabes and thank you for opening this up to fix it up!

PR #4429, opened 2023-11-09
adds dbt metrics and legacy SL notices

mirnawong1:
this pr adds a deprecation notice for users that use dbt metrics or legacy SL (v1.5 orlower)


website/snippets/_sl-deprecation-notice.md
```diff
@@ -0,0 +1,6 @@
+:::info Deprecation of dbt Metrics and the legacy dbt Semantic Layer
+For users on dbt version 1.5 or earlier &mdash; starting on December 15th, 2023, support for dbt Metrics and the old dbt Semantic Layer (v1.5 and below) will end. To benefit from the latest features, upgrade to the most [recent version](/guides/migration/sl-migration) of dbt and embrace the enhanced dbt Semantic Layer, powered by MetricFlow.
```
matthewshaver: The link text is "upgrade to the most [recent version] of dbt" but the linked doc doesn't appear to contain any instructions for upgrading their dbt version, nor does it link to those instructions at the bottom of the page.

Suggest removing the "Starting" language with "end" in the same sentence so that we're as clear as possible
```suggestion
For users on dbt version 1.5 or earlier &mdash; On December 15th, 2023, support for dbt Metrics and the old dbt Semantic Layer (v1.5 and below) will end. To benefit from the latest features, upgrade to the most [recent version](/guides/migration/sl-migration) of dbt and embrace the enhanced dbt Semantic Layer, powered by MetricFlow.
```

website/snippets/_sl-deprecation-notice.md
```diff
@@ -0,0 +1,6 @@
...
+
+After this date, dbt Labs won't offer support, bug fixes, or security updates for these deprecated features. Documentation for them will be discontinued, and they will be removed from the dbt Cloud user interface.
```
matthewshaver: The way it's written it could be construed as the user still having access to the features in some manner, it just won't be updated. Suggest clear, concise language that this will be gone.

```suggestion
After December 15th, dbt Labs will no longer support these deprecated features, they will be removed from the dbt Cloud user interface, and their documentation removed from the docs site.
```

website/docs/docs/use-dbt-semantic-layer/tableau.md
```diff
@@ -16,11 +16,12 @@ This integration provides a live connection to the dbt Semantic Layer through Ta
 
 ## Prerequisites
 
-1. You must have [Tableau Desktop](https://www.tableau.com/en-gb/products/desktop) installed with version 2021.1 or greater
+- You have [configured the dbt Semantic Layer](/docs/use-dbt-semantic-layer/setup-sl) and are using dbt v1.6 or higher.
```
matthewshaver, line 19: We should come to a consensus as a team on how to handle versioning. Here we say v1.6 or higher but below we say 1.6 or earlier. We should choose either `earlier and later` or `higher and lower` I don't have a preference either way but we should stay consistent. 
```suggestion
- You have [configured the dbt Semantic Layer](/docs/use-dbt-semantic-layer/setup-sl) and are using dbt v1.6 or higher.
```

website/snippets/_legacy-sl-callout.md
```diff
@@ -1,11 +1,10 @@
-:::important Upgrade to access the new dbt Semantic Layer
+:::important Upgrade to access the latest dbt Semantic Layer
```
matthewshaver, line 1: `Latest` indicates that we will be upgrading/versioning the semantic layer regularly. Is that what we'll be doing moving forward?

mirnawong1: good question and i think we'd probably be doing that for metricflow (which powers the SL). i was told previously to use 'latest' or 'more recent' instead of 'new' when talking about a newly launched features because the word 'new' will get outdated soon.

website/snippets/_sl-deprecation-notice.md
mirnawong1: i'll clarify this - you need to change you sl configs first before changing your dbt version.

website/docs/docs/use-dbt-semantic-layer/tableau.md
mirnawong1: i tend to use higher and lower usually

matthewshaver (approved):

PR #4434, opened 2023-11-10
Link to reference/resource-configs/schema#tests

gatewaycat:

Current text reads "you can configure the schema to a different value," alluding to https://docs.getdbt.com/reference/resource-configs/schema#tests, but does not link to it. Adding such a link.

I believe this link will help dbt practitioners find the --store-failures schema configuration, in particular because when you google "--store-failures schema," this page https://docs.getdbt.com/reference/resource-configs/store_failures is the 1st result, but the page you actually want https://docs.getdbt.com/reference/resource-configs/schema#tests does not appear at all, as google hides it "In order to show you the most relevant results, we have omitted some entries very similar to the 100 already displayed." This has been the case since at least 2022.

In the existing text "you can configure the schema to a different value," I added the link to the word "configure." This seemed consistent with other links on this page. However, if you believe a different word should get the link, or a wordier addition should be made such as "For more information, refer to the schema configurations" (with the link going on "schema configurations"), or something similar, please do go ahead and make the adjustment as you see fit, or ask me to do it in a comment.

## Checklist

### DIFF
--- a/website/docs/reference/resource-configs/store_failures.md
+++ b/website/docs/reference/resource-configs/store_failures.md
@@ -10,7 +10,7 @@ Optionally set a test to always or never store its failures in the database.
 - If specified as `true` or `false`, the
 `store_failures` config will take precedence over the presence or absence of the `--store-failures` flag.
 - If the `store_failures` config is `none` or omitted, the resource will use the value of the `--store-failures` flag.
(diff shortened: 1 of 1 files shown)


matthewshaver (approved):
Great suggestion. Thanks @gatewaycat

PR #4444, opened 2023-11-12
add timescaledb as community adapter

sdebruyn:
I am adding TimescaleDB as a community-maintained adapter.


matthewshaver:
Hi @sdebruyn - Looks good and happy to add this but have a question about the link - the page it leads to says that the adapter is supported on dbt Core v1.17. Should that say 1.7?

matthewshaver (approved):

sdebruyn:
> Should that say 1.7?

My bad, thanks for spotting! Fixed 😀

PR #4446, opened 2023-11-13
Update meta.md

mirnawong1:
header isn't rendering correctly, this pr fixes it by removing a space before the h3


dbeatty10 (approved):
Good catch!

Can confirm that [this](https://docs-getdbt-com-git-mirnawong1-patch-12-dbt-labs.vercel.app/reference/resource-configs/meta#override-one-meta-attribute-for-a-single-model) preview fixes the problem.

### Before

image

### After

image

Issue #3964, opened 2023-08-21
Missing docs for resource types + general configs


dbeatty10:
### Contributions

- [X] I have read the contribution docs, and understand what's expected of me.

### Link to the page on docs.getdbt.com requiring updates

See below.

### What part(s) of the page would you like to see updated?

There is a many-to-many relationship between resource types and general configs. Each resource type has its own docs page as does each general config. These pages should refer to each other both(!) directions when a general config is applicable to a particular resource type.

The ✅ below is when a link exists in at least one direction, and the ☑️ are when something is linked one direction, but not the other.

⚠️ The tables below **only** point out where there is internal inconsistency; it does **not** highlight any potential instances where existing links are incorrect or utterly incomplete.

### Look at it one direction

The ☑️ below indicate that a general config page (like [`enabled`](https://docs.getdbt.com/reference/resource-configs/enabled) contains an example for a resource type (like `metrics`), but the config page for the resource type is missing a similar example.

|              | [models](https://docs.getdbt.com/reference/model-configs#general-configurations) | [seeds](https://docs.getdbt.com/reference/seed-configs#general-configurations) | [snapshots](https://docs.getdbt.com/reference/snapshot-configs#general-configurations) | [tests](https://docs.getdbt.com/reference/test-configs#general-configurations) | [sources](https://docs.getdbt.com/reference/source-configs#general-configurations) | [analyses](https://docs.getdbt.com/reference/analysis-properties) | [exposures](https://docs.getdbt.com/reference/exposure-properties) | [macros](https://docs.getdbt.com/reference/macro-properties) | metrics |
|--------------|--------|-------|-----------|-------|---------|----------|-----------|--------|---------|
| [enabled](https://docs.getdbt.com/reference/resource-configs/enabled)      |    ✅   |   ✅   |     ✅     |   ✅   |    ✅    |          |     ✅     |        |    ☑️    |
| [tags](https://docs.getdbt.com/reference/resource-configs/tags)         |    ✅   |   ✅   |     ✅     |   ✅   |         |     ✅    |           |        |         |
| [pre-hook](https://docs.getdbt.com/reference/resource-configs/pre-hook-post-hook)     |    ✅   |   ✅   |     ✅     |       |         |          |           |        |         |
| [pre-hook](https://docs.getdbt.com/reference/resource-configs/pre-hook-post-hook)    |    ✅   |   ✅   |     ✅     |       |         |          |           |        |         |
[...]

PR #162, opened 2020-05-20
Restructure model configs

clrcrl:
## Description & motivation
Restructure the "models" docs so that they follow the new structure

## Not done in this PR:
- Some of the reference docs need some more attention — I've left some comments there
- I **did not** implement any 0.17.0 changes, and left the `using-variables` article untouched

PR #4448, opened 2023-11-13
add config descriptions per user feedback

mirnawong1:
per user feedback during coalesce, the config sections don't explain the differences between resource-type configs vs general configs. 

this pr adds a config description to the resource config section and general config section.


website/snippets/_config-description-general.md
```diff
@@ -0,0 +1 @@
+General configurations provide broader operational settings applicable across multiple resources. These include enabling or disabling resources, tagging, database and schema specifications, and operational settings like pre-hooks and post-hooks. These can also be set in the project file, property file, or within the resource file.
```
dbeatty10: Since not all resource types support database, schema, pre-hooks, etc, it might be easier to read without the exact examples, especially since the code example underneath the snippet shows the actual general configs supported by that particular resource type.

So maybe something shorter like this?
```suggestion
General configurations provide broader operational settings applicable across multiple resource types. Like resource-specific configurations, these can also be set in the project file, property files, or within resource-specific files.
```

website/snippets/_config-description-resource.md
```diff
@@ -0,0 +1 @@
+Resource-specific configurations are applied to individual dbt resources. They influence specific attributes of the resource, such as [`sql-header`](/reference/resource-configs/sql_header), [`column_types`](/reference/resource-configs/column_types), [`strategy`](/reference/resource-configs/strategy), and so on. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro.
```
dbeatty10: Since the code example underneath the snippet shows the _actual_ resource-specific configs (along with hyperlinks) supported by that particular resource type, what would you think about simplifying it to this?

```suggestion
Resource-specific configurations are applicable to only one dbt resource type rather than multiple resource types. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro.
```

dbeatty10 (commented):
Awesome use of snippets 💪

### Practical distinction

At a high level, I'm not aware of a meaningful distinction between resource-specific and general configs that affects end users and how they use dbt.

### Conceptual distinction

The only difference I'm aware of:
- "Resource-specific configurations" are applicable to only one dbt resource type and "general configurations" are applicable to multiple resource types.

But I don't know a practical difference that affect how users think and act.

### Radical alternative

So one radical alternative would be to combine the resource-specific and general configs into a single section and then no additional explanations are even needed.

For seeds, the combined listing for `dbt_project.yml` might look like this (either give or take the inline comments):
```yaml
seeds:
  <resource-path>:

    # resource-specific configs only available to seeds
    +quote_columns: true | false
    +column_types: {column_name: datatype}
    +delimiter: <string>

    # general configs also available in other resource types
    +enabled: true | false
    +tags: <string> | [<string>]
    +pre-hook: <sql-statement> | [<sql-statement>]
    +post-hook: <sql-statement> | [<sql-statement>]
    +database: <string>
    +schema: <string>
    +alias: <string>
    +persist_docs: <dict>
    +full_refresh: <boolean>
    +meta: {<dictionary>}
    +grants: {<dictionary>}
```

### History

Here are the pull requests where these separate headings originally appeared for each different resource types:
- [#111](https://github.com/dbt-labs/docs.getdbt.com/pull/111)
- [#158](https://github.com/dbt-labs/docs.getdbt.com/pull/158)
- [#162](https://github.com/dbt-labs/docs.getdbt.com/pull/162)
- [#705](https://github.com/dbt-labs/docs.getdbt.com/pull/705)

### See also

#3964 has some tables showing which general configs apply to which resources. It does not contain any resource-specific configs.

website/snippets/_config-description-resource.md
```diff
@@ -0,0 +1 @@
+Resource-specific configurations are applicable to only one dbt resource type rather than multiple resource types. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro.
```
mirnawong1:
```suggestion
Resource-specific configurations apply to only one dbt resource type rather than multiple resource types. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro. 
```

mirnawong1:
thank you for your detailed technical comments, i appreciate it. 🙏  

I think combing them is a great idea, however i do think doing that is still not explicit enough for users. i like the idea of being more explicit so users have more insight into how they can apply each type of config for their own use case, as oppose to inferring it somehow in the comments of header (currently it's inferred via the header, i think). 

I've tried to also make it clearer that the resource-specific config is only available that source using props but it's not working for me so I'm going to investigate more.

website/snippets/_config-description-resource.md
```diff
@@ -0,0 +1 @@
+Resource-specific configurations are applicable to only one dbt resource type rather than multiple resource types. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro.<br /><br /><span>The following resource-specific config code block is only available to {props.meta.resource_type}</span>.
```
dbeatty10: The `{props.meta.resource_type}` part isn't rendering correctly for some reason -- do you know why?

Here are the previews that I'm looking at:

- [models](https://docs-getdbt-com-git-mwong-config-descriptions-dbt-labs.vercel.app/reference/model-configs#available-configurations)
- [seeds](https://docs-getdbt-com-git-mwong-config-descriptions-dbt-labs.vercel.app/reference/seed-configs#available-configurations)
- [snapshots](https://docs-getdbt-com-git-mwong-config-descriptions-dbt-labs.vercel.app/reference/snapshot-configs#available-configurations)
- [sources](https://docs-getdbt-com-git-mwong-config-descriptions-dbt-labs.vercel.app/reference/source-configs#available-configurations)
- [tests](https://docs-getdbt-com-git-mwong-config-descriptions-dbt-labs.vercel.app/reference/test-configs#available-configurations)

image

dbeatty10:
> I've tried to also make it clearer that the resource-specific config is only available that source using props but it's not working for me so I'm going to investigate more.

The distinctions aren't working for me either. I suspect it's because there _isn't_ really any meaningful distinction between resource-specific configs and general configs.

I'm guessing the reason the two different headers exist is just to make it easier to separate the configs that are unique to that particular resource type. e.g., show which configuration options for the snapshot resource type are different than all the rest of the resource types.

> I think combing them is a great idea, however i do think doing that is still not explicit enough for users. i like the idea of being more explicit so users have more insight into how they can apply each type of config for their own use case, as oppose to inferring it somehow in the comments of header (currently it's inferred via the header, i think).

What would be some ways to give insight on how to apply each type of config? Do you think the [examples](https://docs.getdbt.com/reference/snapshot-configs#examples) section helps with that?

mirnawong1:
will create a separate issue to restructure the config pages

website/snippets/_config-description-resource.md
```diff
@@ -0,0 +1,3 @@
+Resource-specific configurations are applicable to only one dbt resource type rather than multiple resource types. You can define these settings in the project file (`dbt_project.yml`), a property file (`models/properties.yml` for models, similarly for other resources), or within the resource’s file using the `{{ config() }}` macro.<br />
+
+<span>The following resource-specific config code block is only available to {props.meta.resource_type}</span>:
```
dbeatty10: I'm going approve in a moment, which will trigger the auto-merge.

But adding this to a follow-up PR might read a little easier:

```suggestion
<span>The following resource-specific configurations are only available to {props.meta.resource_type}</span>:
```

Here's how it looks like right now:

image

dbeatty10 (approved):

PR #4472, opened 2023-11-15
add clone step in cloud cli doc

mirnawong1:
adding clonining step in cloud cli install docs to clarify user confusion. also adding add'l step on 'configure' page.


dave-connors-3 (approved):
🦞

PR #4473, opened 2023-11-15
Update `saved_queries` and `group_by`

Jstein77:
Update the saved query spec to conform with the user input core expects.


mirnawong1 (approved):
lgtm!

PR #4479, opened 2023-11-15
Removed repeated line of text.

richardgourley:

I have removed the repetition of the line _'Do the column names and values look correct for your use case?'_ from the section **'Structuring'**.

It doesn't look to me like the above line was repeated for emphasis.  I think it looks like a possible mistake?


matthewshaver (approved):

matthewshaver:
Great catch, thank you @richardgourley

PR #4487, opened 2023-11-16
Conclusion - changed 'and' to 'or'

richardgourley:

In the 'Conclusion' section I have changed '....Weigh your options in terms of whether to go with **open source and enterprise.**.....' to '.....Weigh your options in terms of whether to go with **open source or enterprise**.....'

If I have understood the article correctly, I think it should be 'or'?


mirnawong1 (approved):
hey @richardgourley , indeed you are right! this is a great catch and thank you so much for catching this. appreciate it and looking forward to future contributions!

PR #4488, opened 2023-11-16
Update redshift-qs.md

mirnawong1:
image not rendering due to case sensitivity. updated image URL to S3 instead s3. also adding more clarity that `IAM` doesn't support dbt cloud per cs feedback.


mirnawong1:
staging links shows all fixed: https://docs-getdbt-com-git-mirnawong1-patch-12-dbt-labs.vercel.app/guides/redshift?step=3

jennlw (approved):
Change looks good to be more prominent so customer's see it is not supported in dbt Cloud. Thanks!

PR #4540, opened 2023-11-28
Update explanation of contract enforcement

joellabes:
Rephrasing the explanation of when contracts might be supported but not enforced. It's a _bold_ editorial choice to describe Postgres as a legacy database; it's still current, but has different priorities. 

Open to alternative phrasings but directionally this is where we should be headed.


website/docs/docs/collaborate/govern/model-contracts.md
```diff
@@ -91,7 +91,7 @@ When building a model with a defined contract, dbt will do two things differentl
 Select the adapter-specific tab for more information on [constraint](/reference/resource-properties/constraints) support across platforms. Constraints fall into three categories based on support and platform enforcement:
 
 - **Supported and enforced** &mdash; The model won't build if it violates the constraint.
-- **Supported and not enforced** &mdash; The platform supports specifying the type of constraint, but a model can still build even if building the model violates the constraint. This constraint exists for metadata purposes only. This is common for modern cloud data warehouses and less common for legacy databases.
+- **Supported and not enforced** &mdash; The platform supports specifying the type of constraint, but a model can still build even if building the model violates the constraint. This constraint exists for metadata purposes only. Enforcement is more common for transactional databases, unlike cloud data warehouses.
```
mirnawong1:
```suggestion
- **Supported and not enforced** &mdash; The platform supports specifying the type of constraint, but a model can still build even if building the model violates the constraint. This constraint exists for metadata purposes only. This approach is more typical in cloud data warehouses than in transactional databases, where strict rule enforcement is more common.
```

mirnawong1 (approved):
thanks @joellabes, small tweak and approving for you! let me know if the phrasing is not what you're expecting though!

PR #4549, opened 2023-11-29
removing limitation bullet

mirnawong1:
removing `metric_time` limitation as this was resolved in https://github.com/dbt-labs/metricflow/issues/825

### DIFF
--- a/website/docs/docs/build/cumulative-metrics.md
+++ b/website/docs/docs/build/cumulative-metrics.md
@@ -38,7 +38,7 @@ metrics:
 
 ## Limitations
 Cumulative metrics are currently under active development and have the following limitations:
-- You can only use the [`metric_time` dimension](/docs/build/dimensions#time) to check cumulative metrics. If you don't use `metric_time` in the query, the cumulative metric will return incorrect results because it won't perform the time spine join. This means you cannot reference time dimensions other than the `metric_time` in the query.
+- You are required to use [`metric_time` dimension](/docs/build/dimensions#time) when querying cumulative metrics. If you don't use `metric_time` in the query, the cumulative metric will return incorrect results because it won't perform the time spine join. This means you cannot reference time dimensions other than the `metric_time` in the query.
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/cumulative-metrics.md
```diff
@@ -38,10 +38,7 @@ metrics:
...
-2. If you use `metric_time` in your query filter but don't include "start_time" and "end_time," cumulative metrics will left-censor the input data. For example, if you query a cumulative metric with a 7-day window with the filter `{{ TimeDimension('metric_time') }} BETWEEN '2023-08-15' AND '2023-08-30' `, the values for `2023-08-15` to `2023-08-20` return missing or incomplete data. This is because we apply the `metric_time` filter to the aggregation input. To avoid this, you must use `start_time` and `end_time` in the query filter.
-
+- You can only use the [`metric_time` dimension](/docs/build/dimensions#time) to check cumulative metrics. If you don't use `metric_time` in the query, the cumulative metric will return incorrect results because it won't perform the time spine join. This means you cannot reference time dimensions other than the `metric_time` in the query.
```
Jstein77: > You can only use the [`metric_time` dimension](/docs/build/dimensions#time) to check

I would re-phrase this too "You are required to use `metric_time` when querying cumulative metrics"

Jstein77 (approved):
Left one comment, otherwise looks good!

website/docs/docs/build/cumulative-metrics.md
mirnawong1:
```suggestion
- You are required to use [`metric_time` dimension](/docs/build/dimensions#time) when querying cumulative metrics. If you don't use `metric_time` in the query, the cumulative metric will return incorrect results because it won't perform the time spine join. This means you cannot reference time dimensions other than the `metric_time` in the query.
```

mirnawong1:
thanks @Jstein77 ! folded i your feedback and will merge this now!

PR #4557, opened 2023-11-30
adding custom materialization link

mirnawong1:
[per slack thread](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1701341336153579), adding a linnk to the materializataion guide and also removing the 'core' label to that guide since it's applicable to core and cloud users.


dbeatty10 (approved):
🚀

PR #4561, opened 2023-11-30
removing hex video link

mirnawong1:
this pr removes the old hex video loom and enables user to go directly to hex docs


rpourzand (approved):
lgtm!

PR #4585, opened 2023-12-05
Updated as per dbt-teradata 1.7.0

tallamohan:
* Updated compatibility matrix with dbt-core 1.7.0 support
* Added newly migrated cross-database macros


matthewshaver (approved):
Thank you @tallamohan

PR #4627, opened 2023-12-12
Remove unnecessary step from dbx orch guide

joellabes:
There was a step in this guide which was just links to the next steps of the guide (which didn't work anyway due to anchor oddities). This removes that. 

image


matthewshaver (approved):
Thanks @joellabes - makes sense to remove this now that it's in the new guide format.

PR #4647, opened 2023-12-14
remove callout

mirnawong1:
multi hop joins are supported now for the SL and MF so this pr removes callout that states multi hop joins aren't supported


Jstein77 (approved):
LGTM!

PR #4655, opened 2023-12-14
Explain how transactions (and rollbacks) are handled for incremental strategies

### LINKED ISSUE #4635: Explain how transactions (and rollbacks) are handled for incremental strategies
### Contributions

https://docs.getdbt.com/docs/build/incremental-models

### Question

The goal of this issue is to document the answer to this question from https://github.com/dbt-labs/dbt-core/issues/7011#issuecomment-1443698015:

> For the DML, we know that only the delete+insert strategy will work, and not merge. With delete+insert incremental strategy, does Dbt wrap both in a single transaction? If not, and they are treated as 2 separate transactions that would be bad. For example, the delete could succeed, and the subsequent insert could fail. The table would then be left with deleted records = Not Good. Can you elaborate on how Dbt already handles that. I cannot find it anywhere in the doc.

### Answer

On data platforms that use transactions, dbt will roll back open transactions if any statement in the materialization fails. This is useful to restore the database to a good state even if the materialization breaks in the middle (due to bad data or a logic bug like a divide by zero issue).

### Additional information

This was originally included as part of https://github.com/dbt-labs/docs.getdbt.com/issues/2895 on 2023-02-20.

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty-explain-incremental-957f61-dbt-labs.vercel.app/docs/build/incremental-models#how-do-incremental-models-work-behind-the-scenes)

resolves #4635

The goal of this issue is to document the answer to this question from https://github.com/dbt-labs/dbt-core/issues/7011#issuecomment-1443698015:

> For the DML, we know that only the delete+insert strategy will work, and not merge. With delete+insert incremental strategy, does Dbt wrap both in a single transaction? If not, and they are treated as 2 separate transactions that would be bad. For example, the delete could succeed, and the subsequent insert could fail. The table would then be left with deleted records = Not Good. Can you elaborate on how Dbt already handles that. I cannot find it anywhere in the doc.

### 🎩

image

## Checklist


website/docs/docs/build/incremental-models.md
```diff
@@ -174,7 +174,7 @@ dbt's incremental materialization works differently on different databases. Wher
 
 On warehouses that do not support `merge` statements, a merge is implemented by first using a `delete` statement to delete records in the target table that are to be updated, and then an `insert` statement.
 
-Transaction management is used to ensure this is executed as a single unit of work.
+On data platforms that support it, transaction management is used to ensure this is executed as a single unit of work. dbt will roll back open transactions and restore the database to a good state if any statement within the materialization fails.
```
mirnawong1:
```suggestion
Transaction management, a process used in certain data platforms, ensures that a set of actions is treated as a single unit of work (or task). If any part of the unit of work fails, dbt will roll back open transactions and restore the database to a good state.
```

mirnawong1 (approved):
hey @dbeatty10 , lgmt and a small sugg on my end!

PR #4687, opened 2023-12-21
add hover function + disclaimer css

mirnawong1:
this pr adds the hover functionality and disclaimer css to the existing faq component. this is because our detailsToggle component has a hover functionality and i think we should have consistent behavior for users.

this PR also removes the underline under the 'hover to view' string.


nghi-ly:
hey @mirnawong1 ! i really like this and appreciate the consistency across the faqs. 

for a little more polish 💅 , is it possible to have the text "Hover to view" _not_ be underlined since that's not actually part of the faq content (it's just a helper note) ?

mirnawong1:
thanks @nghi-ly and that's a good suggestion! I'll look into to removing the underline so the distinction is clearer 🙌

mirnawong1:
update: i've removed the underline from the 'hover to view' for both the faq and detailsToggle components.

website/src/components/detailsToggle/index.js
```diff
@@ -40,7 +40,7 @@ useEffect(() => {
         onMouseLeave={handleMouseLeave}
       >
         <span className={`${styles.toggle} ${isOn ? null : styles.toggleUpsideDown}`}></span>&nbsp;
-        <span>{alt_header}</span>
+        <span className={styles.headerText}>{alt_header}</span>
```
mirnawong1, line 43: adding span to separate alt_header and .disclaimer CSS and assign underline class to alt_header only.

JKarlavige (approved):
all lgtm! verified faqs opening on hover, while still being able to click to open the faqs. Great work @mirnawong1! :shipit:

mirnawong1:
thanks so much for the review @JKarlavige!

PR #4713, opened 2024-01-04
`DO_NOT_TRACK=0` is not the same as `DBT_SEND_ANONYMOUS_USAGE_STATS=True`

dbeatty10:

While responding to https://github.com/dbt-labs/dbt-core/issues/9336, I set `DO_NOT_TRACK=0` and examined the result. It does not behave the same way as `DBT_SEND_ANONYMOUS_USAGE_STATS=True`.

Looking at the source code [here](https://github.com/dbt-labs/dbt-core/blob/11cc71b75fa64b09888461339eb1eb3b394f9528/core/dbt/cli/flags.py#L252-L254) explains why.

So we can safely just remove this line from the docs for [`send_anonymous_usage_stats`](https://docs.getdbt.com/reference/global-configs/usage-stats) to avoid confusion.

## Additional info

We first added support for the [Console Do Not Track](https://consoledonottrack.com/) initiative within https://github.com/dbt-labs/dbt-core/pull/5000 as described in https://github.com/dbt-labs/dbt-core/issues/3540.

Any of the following are equivalent to `DBT_SEND_ANONYMOUS_USAGE_STATS=False` (whether they are uppercase, lowercase, or mixed case):
```
export DO_NOT_TRACK=1
export DO_NOT_TRACK=t
export DO_NOT_TRACK=true
export DO_NOT_TRACK=y
export DO_NOT_TRACK=yes
```

Any other values of `DO_NOT_TRACK` are ignored altogether and not have any effect.

## Checklist


mirnawong1 (approved):
lgtm and thank you for opening this up @dbeatty10 !

PR #4658, opened 2023-12-15
Blog post addition: Serverless dlt + dbt project

euanjohnston-dev:
## What are you changing in this pull request and why?

PR related to blog post ticket raised here: (https://github.com/dbt-labs/docs.getdbt.com/issues/4657).

PR #4749, opened 2024-01-15
Fix date in article because it affects sort order

adrianbr:


mirnawong1:
hey @adrianbr thanks for opening up your first PR! can you say little more about what issue you're experiencing with the blog? the page doesn't really have a sort so would love to understand what you're experiencing! thnk you!

adrianbr:
@mirnawong1 the page sorts by time of article. We created an article in Dec 2023 and I guess they wanted to update the date to  Jan 2024 but instead they did jan 2023, which causes the article to be burried  and not visible among the latest articles

mirnawong1 (approved):
thanks for clarifying @adrianbr and i understand what you're saying! approving and merging for you so the order is updated and correct - thank you again ✨!

original here; https://github.com/dbt-labs/docs.getdbt.com/pull/4658

PR #4754, opened 2024-01-16
add plan/accounts guidance to style guide

mirnawong1:
add this clarification per docs discussion a few weeks back. [source](https://www.notion.so/dbtlabs/Docs-team-meeting-8852fdf3f9354945aba185625e8eaec4?pvs=4)

### DIFF
--- a/contributing/content-style-guide.md
+++ b/contributing/content-style-guide.md
@@ -519,6 +519,7 @@ enter (in the command line) | type (in the command line)
 email | e-mail
 on dbt | on a remote server
 person, human | client, customer
+plan(s), account | organization, customer
 press (a key) | hit, tap
 recommended limit | soft limit
 sign in | log in, login
@@ -528,6 +529,8 @@ username | login
 dbt Cloud CLI | CLI, dbt CLI
 dbt Core | CLI, dbt CLI
 </div></b>
+
+Make sure you keep the reader close to the content and documentation (second person). For example, to explain that a feature is available to a particular dbt Cloud plan &mdash; use “XYZ is available on Enterprise plans” or “People on Enterprise plans can XYZ..” to keep users closer to the documentation. This will signal to users to check their plan or account status on their own. 
 
 ## Links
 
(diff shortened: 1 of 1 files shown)


contributing/content-style-guide.md
```diff
@@ -529,6 +530,8 @@ dbt Cloud CLI | CLI, dbt CLI
 dbt Core | CLI, dbt CLI
 </div></b>
 
+Make sure you keep the reader close to the content and documentation (second person). For example, to explain that a feature is available to a particular dbt Cloud plan &mdash; use “XYZ is available on Enterprise plans” or “People on Enterprise plans can XYZ..” to keep users closer to the documentation. This will signal to users to check their plan or account status on their own.
```
matthewshaver:
```suggestion
Make sure you keep the reader close to the content and documentation (second person). For example, to explain that a feature is available to a particular dbt Cloud plan &mdash; use “XYZ is available on Enterprise plans” or “Customers on Enterprise plans can XYZ...” to keep users closer to the documentation. This will signal to users to check their plan or account status on their own. 
```

mirnawong1: i believe the above rec came from Leona and everyone agreed. also our style guide says to avoid the use of 'customers' so I'm going to keep people to be consistent with the avoidance of using 'customer' and keeping the user closer to the docs. 'customer' already has a layer of separation i believe.

matthewshaver (approved):

contributing/content-style-guide.md
mirnawong1: can change to this from thread and bypass docs sync sugg

```suggestion
Note, let's make sure we're talking to our readers and keep them close to the content and documentation (second person). 

For example, to explain that a feature is available to a particular dbt Cloud plan, you can use:
- “XYZ is available on Enterprise plans”,
- “If you're on an Enterprise plan, you can access XYZ..” or 
- "Enterprise plans can access XYZ..." to keep users closer to the documentation. 

This will signal users to check their plan or account status independently.
```

PR #4779, opened 2024-01-22
update handlescroll

mirnawong1:
this pr updates the handlescroll function and removes setOn(false so that the toggle will no longer close automatically when the user starts scrolling. this is to allow the user to continue reading the content inside the toggle without interruption

per slack feedback here: https://dbt-labs.slack.com/archives/C069P1RQ8LT/p1705925559782539?thread_ts=1705925278.391799&cid=C069P1RQ8LT


JKarlavige (approved):
lgtm! verified toggles stay open on scroll

PR #4790, opened 2024-01-24
update to add links and add clarity on model contracts support

mirnawong1:
adding links to clarify or add context on operations and the run operation command. also adding info on model contracts and its relation to ephemeral materialization [per feedback](https://getdbt.slack.com/archives/C0441GSRU04/p1705684634213419)


dbeatty10 (approved):

mirnawong1:
thank you @dbeatty10 !!

PR #4804, opened 2024-01-25
adding more usecases to lifecycle component

mirnawong1:
this pr enhances the existing lifecycle component to add more badge support:
- support for plans
- support for lifecycle (same)
- support for mulitiple badges
- set custom colors
- use fetElements to ignore lifecycle css in the right toc

see [asana task for more info](https://app.asana.com/0/1200099998847559/1206377128451074/f)

*note, still waiting for color and add'l use case before final review. 

### way to use it

```
<Lifecycle status="enterprise" /> // single status, predefined color
<Lifecycle status={['team', 'enterprise']} /> // multiple statuses, predefined colors
<Lifecycle status="custom" backgroundColor="#123456" /> // Custom color
```

### colors
predefined colors are:
- prev used #f3f4f6 for the plan use case (enterprise, team), however now (as of 2/9, using #EBEDF0)
- #047377 for the lifecycle use case (new, beta, ga, public preview)


mirnawong1:
hey @JKarlavige whenever you have a chance, can you review this PR please? i believe @nghi-ly and i were asking Jey to clarify the 'gray' color as it's looking washed out when you view the docs in light mode. the hex i used it #f3f4f6

mirnawong1:
adjusted font per feedback
Screenshot 2024-02-07 at 13 18 04
Screenshot 2024-02-07 at 13 18 08

website/src/theme/DocItem/Layout/index.js
```diff
@@ -46,6 +46,9 @@ function useDocTOC() {
   async function fetchElements() {
     // get html elements
     const headings = await getElements(".markdown h1, .markdown h2, .markdown h3, .markdown h4, .markdown h5, .markdown h6")
+
+    // filter out the Lifecycle badge from the headings
+    const filteredHeadings = headings.filter(heading => !heading.classList.contains('lifecycle'));
```
JKarlavige: For this - it looks like it is errors out with the error: `Uncaught (in promise) TypeError: headings.filter is not a function`
image

Headings is a `NodeList`, which is not a standard array. This means `.filter` is not available on the `headings` variable. Since it's erroring out, the javascript is stopping and the TOC is not refreshing (needed for versioning).

It looks like by default, Docusaurus excludes components in TOCs. However, since we have customization built in, the Lifecycle component html is added in during the TOC refresh _(lets just ignore this for now as it gets pretty confusing)_.

We'll have to remove this `filteredHeadings` line to get the TOC customization back up and running. And then to hide the Lifecycle badge from the TOC, the simplest way to do this is with CSS. The TOC is wrapped in a `.table-of-contents` class which we can use. I'd also recommend adding a string class to the lifecycle span tag, for example:
```
className={`${styles.lifecycle} lifecycle-badge`}
```

Then in website/src/css/custom.css, we can add a line to target `.table-of-contents .lifecycle-badge` and `display: none` those elements.

There's a lot going on here so i'm game to jump in a zoom to dive into this further if you had any questions!

website/src/components/lifeCycle/index.js
```diff
@@ -1,11 +1,41 @@
...
+  new: '#047377',
+  beta: '#047377',
+  ga: '#047377',
+  'public preview': '#047377',
+};
+
+const fontColors = {
+    enterprise: '#000000',
+    team: '#000000', 
+    developer: '#000000',
```
JKarlavige: The dark text throughout the site is a dark navy blue (`#262A38`) rather than black, can we update these to use the `#262A38` value

website/src/components/lifeCycle/index.js
JKarlavige, line 1: Similar validation note as your expandable component. Since the `status` prop looks to be required, we should check if the `status` prop is passed in before applying any html. Otherwise it will leave a small element with no text if the status prop is not passed in.
image

website/src/components/lifeCycle/index.js
```diff
@@ -1,11 +1,41 @@
...
+
 export default function Lifecycle(props) {
-    if (!props.status) {
-        return null;
-    }
-    return (
-        <span className={styles.lifecycle}>{props.status}</span>
-    );
+  // Check if props.status is an array or a single value
+  const statuses = Array.isArray(props.status) ? props.status : [props.status];
```
JKarlavige, line 26: This is a great way to ensure `statuses` is always an array! 

Also wanted to share a shorthand way of achieving the same result using the [spread operator](https://www.samanthaming.com/tidbits/92-6-use-cases-of-spread-with-array/): 
`const statuses = [...props.status];` 

If `props.status` is a string, it will spread it into an array. If `props.status` is an array, it will create a clone of it, essentially being the same as the initial array.

website/src/components/lifeCycle/index.js
```diff
@@ -1,11 +1,41 @@
...
+
+  return (
+    <>
+      {statuses.map((status, index) => {
+        const style = {
+          backgroundColor: props.backgroundColor || statusColors[status] || '#047377', // default to teal if no match
+          color: fontColors[status] || '#fff' // default font color if no matc
```
JKarlavige, line 33: This looks like it works great, nice way to solve this!

JKarlavige (changes requested):
Great work @mirnawong1! Left some feedback, let me know if you have any questions! Especially around the TOC side of it as it might get confusing.

Also, in the task there's a note: `Support adding badge to page level (h1)`. Should this be included within this PR, or are we holding off on that for now?

mirnawong1:
thanks so much @JKarlavige ! will review this 🙏

mirnawong1:
amazing! thank you so much -- @JKarlavige and @nghi-ly this is ready for re-review please

loom here: https://www.loom.com/share/b7a0482788674381a70574c9fd0bd503?sid=13b7add0-583b-4c98-a696-23b2d8768373

mirnawong1:
i've also updated the asana task and asked for color verification: https://app.asana.com/0/1200099998847559/1206377128451074/f

website/src/components/lifeCycle/index.js
```diff
@@ -1,11 +1,44 @@
...
 export default function Lifecycle(props) {
-    if (!props.status) {
-        return null;
-    }
-    return (
-        <span className={styles.lifecycle}>{props.status}</span>
-    );
+  if (!props.status || (Array.isArray(props.status) && props.status.length === 0)) {
+    return null;
+  }
```
JKarlavige, line 24: 🔥

JKarlavige (approved):
lgtm! :shipit: all edits look to work great, nice work getting those adjustments in!

Re: H1s - good call on adding the `# H1 header` to pages where this will be needed and adding the lifecycle component to that line. I think that's much cleaner and less lift than trying to support adding the lifecycle to the title from the frontmatter.

Awesome work! 😁

mirnawong1:
yayyyy! thank you soo much @JKarlavige !! am going to wait for @nghi-ly to come online for next steps wrt to merging 🙏

nghi-ly (approved):
lgtm! 

re: H1s, just to double-check, will using `# h1 title` to override in the MD affect anything adversely like search results, SEO, google search results text, etc?

mirnawong1:
i'll defer to @JKarlavige on this but after researching, it looks like search engines might give more weight to the original frontmatter title than the overridden `<H1>` content. so would it make sense to use the `#h1` tag (which will have the same content as the frontmatter title) for visual rendering and styling purposes? also generally the lifecycle status will get removed after a particular time so i'm not sure if we want that or not in the SEO?

JKarlavige:
> i'll defer to @JKarlavige on this but after researching, it looks like search engines might give more weight to the original frontmatter title than the overridden `<H1>` content. so would it make sense to use the `#h1` tag (which will have the same content as the frontmatter title) for visual rendering and styling purposes? also generally the lifecycle status will get removed after a particular time so i'm not sure if we want that or not in the SEO?

The `title` frontmatter sets the `<title>` meta tag. This holds more importance over the h1 tag, as it's what shows in search results and in social sharing, and helps search engines know what the page is about.

The h1 tag provides additional context for search engines as to what the content on the page is related to. 

Even if the frontmatter `title` property is different than the h1, as long as they accurately represent what the page is about there shouldn't be an issue.

PR #4811, opened 2024-01-26
Mark v1.4 as EOL

jtcohen6:

dbt Core v1.4 was released on 25 January 2023. As of today, it's considered "End of Life"


mirnawong1 (approved):
lgtm thank you @jtcohen6 ! i think we should prob add some of these dates to our docs release calendar @matthewshaver @runleonarun @nghi-ly so we protectively update the other versions . I'm sure there's a more automatic way to do it, but suggesting the calendar as the simplest way now.

mirnawong1:
Preview looks good https://docs-getdbt-aq9mlvqgy-dbt-labs.vercel.app/docs/dbt-versions/core

PR #4824, opened 2024-01-29
fixed references in group private access example.

MarkJacksonSeccl:
the ref was pointing to `finance_model` when it is defined as `finance_private_model`

the ref was pointing to `finance_model` when it is defined as `finance_private_model`


mirnawong1 (approved):
hey @MarkJacksonSeccl great catch! thank you for opening this pr up to fix this up -- will get this merged and looking forward to future contributions ✨

PR #4844, opened 2024-02-01
clarify cloud cli config file confusion

mirnawong1:
this pr adds clarification that the user should store the dbt_cloud.yml file in a safe place and that it should be saved in their local computer storage per slack thread: https://dbt-labs.slack.com/archives/C06FWKP5V28/p1706649874140829 from @TimKlense


TimKlense (approved):
Thanks for clarifying this part of installing cloud cli, it looks good to me

PR #4847, opened 2024-02-01
new expandable component

mirnawong1:
this pr creates completes this asana task: https://app.asana.com/0/1200099998847559/1206422502609164/f and creates an expandable component similar to notion toggles. it should:
1. display the expanded content just looks like the other content on the page (ie no box border and no color)   
2. make the alt_header inkable like a section title

we can use the following compoent to use it:

```
<expandable alt_header="add your info here">
add your content here
</expandable>
```


mirnawong1:
adding the add'l orange arrow color requested in the asana ticket: Recording Feb


and loom explaining here: https://www.loom.com/share/6ce68b13f2eb49b4b899dfd38ce6c3b3?sid=ccfb843d-0020-424c-9807-a38d6f734fa6

website/package.json
```diff
@@ -14,6 +14,7 @@
     "@docusaurus/plugin-ideal-image": "2.3.1",
     "@docusaurus/preset-classic": "2.3.1",
     "@docusaurus/theme-search-algolia": "2.3.1",
+    "@headlessui/react": "^1.7.18",
```
JKarlavige: Is the `headlessui` package being used anywhere within this PR? If not used feel free to remove the package for now, or if it is used feel free to disregard this comment! 😁

website/src/components/expandable/index.js
```diff
@@ -0,0 +1,50 @@
...
+    .replace(/[^\w\-]+/g, '') // Remove all non-word chars
+    .replace(/\-\-+/g, '-') // Replace multiple - with single -
+    .replace(/^-+/, '') // Trim - from start
+    .replace(/-+$/, ''); // Trim - from end
+}
+
+function expandable({ children, alt_header = null }) {
+  const [isOn, setOn] = useState(false);
+  // generate a slug from the alt_header
+  const anchorId = slugify(alt_header);
```
JKarlavige: If the `alt_header` prop is not passed into the component, it is being set to `null` by default. This causes `text` to be null within the slugify method. This means `text.toString()` will error out, as `toString` is not a method of `null`. 

image

It looks like this component relies on `alt_header` existing, otherwise there is no text to show next to the arrow.
image

If that's the case, my recommendation would be to add validation to ensure `alt_header` is set before doing anything else within the `expandable` component. If `alt_header` is not passed in, the component should not render any html on the page.

Let me know if you have any questions on how to approach that! My hint would be to focus on adding the validation within the `expandable` component function and not the `slugify` function. As long as the validation is done within the `expandable` function, we can be sure `alt_header` exists before trying to pass it into `slugify`

website/src/components/expandable/styles.module.css
```diff
@@ -0,0 +1,86 @@
...
+    border-color: white; /* White arrow in dark mode */
+    border-color: rgb(253, 153, 83);
+}
+
+:local(.body) {
+    margin-top: 10px;
+    margin-left: 1em;
+    padding: 10px;
+    background-color: transparent; /* Changed from a color to transparent */
+}
```
JKarlavige: The content within the toggle aligns slightly more inward than the `alt_header` text. Can we adjust the styles to align the content with the header, similar to stripe's docs?

Also, just a heads up if we change any styles here, they will affect all elements on the page which have these classes. That means if the FAQs component is used on the same page as this expandable component, there might be conflicting styles.

It would be best to scope these styles specifically to this expandable component. Since the `expandableContainer` class wraps all other elements. The easiest way to scope the styles to this component is by prefixing the classes in this css with `.expandableContainer`, for example: `.expandableContainer :local(.body)`. I'd recommend doing this for all css declarations in this file.

image

website/static/js/headerLinkCopy.js
JKarlavige, line 1: Confirmed i'm able to copy the link to the header!

JKarlavige (changes requested):
Great work on this @mirnawong1! Just have a few small requested edits and left comments in those sections. Let me know if you have any questions with those!

website/package.json
mirnawong1: oh shoot i was testing it and then forgot to remove it - thank you so much!

mirnawong1:
thanks so much @JKarlavige ! will review this 🙏

website/src/components/expandable/styles.module.css
mirnawong1: amazing, thank you so much for the guidance @JKarlavige !! Screenshot 2024-02-09 at 12 46 45

mirnawong1: thank you so much @JKarlavige ! made that change! 
Screenshot 2024-02-09 at 12 57 18

mirnawong1:
hey @JKarlavige and @nghi-ly - ready for rereview - thank you for the feedback!

website/package-lock.json
JKarlavige, line 1: It looks like the `headlessui` package is still hanging out in the `package-lock`. Re-running `npm install` locally and committing the new `package-lock` should get that removed!

mirnawong1: oh darn! thank you for flagging Jason!

website/src/components/expandable/index.js
```diff
@@ -0,0 +1,51 @@
...
+    .replace(/[\u0300-\u036f]/g, '') // Remove diacritics
+    .replace(/\s+/g, '-') // Replace spaces with -
+    .replace(/[^\w\-]+/g, '') // Remove all non-word chars
+    .replace(/\-\-+/g, '-') // Replace multiple - with single -
+    .replace(/^-+/, '') // Trim - from start
+    .replace(/-+$/, ''); // Trim - from end
+}
+
+function expandable({ children, alt_header = null }) {
+  if(!alt_header) { return null; }
```
JKarlavige, line 18: perfect! 😁

JKarlavige (approved):
This lgtm! Just left that one note regarding the `package-lock`, but all functionality looks to be working. Nice job adding the validation to prevent the component from showing unless the header is set!

:shipit: :shipit: :shipit:

nghi-ly (approved):
love love love that orange arrow!

mirnawong1:
.table-of-contents .lifecycle-badge {
  display: none;
}

mirnawong1:
hey @nghi-ly, updated the merge conflicts - are we good to merge this?

nghi-ly:
@mirnawong1 : feel free to merge!

mirnawong1:
i'm merging this now @nghi-ly and @JKarlavige ! will look to see if its possible to add the lifecycle pill to this next week

PR #4919, opened 2024-02-14
update lifecycle component to accept strings + commas

mirnawong1:
this pr quickly updates the lifcycle pill so that we can use the lifecycle pill without the curly brackets (`### test <Lifecycle status="team,enterprise"/>`)

and updated the lifecycle code to check if the status is a string and accept strings separated by commands (without curly brackets).

### DIFF
--- a/website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
+++ b/website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
@@ -8,8 +8,6 @@ meta:
   api_name: dbt Semantic Layer APIs
 ---
 
-### test <Lifecycle status="team,enterprise"/>
-
 import CreateModel from '/snippets/_sl-create-semanticmodel.md';
 import DefineMetrics from '/snippets/_sl-define-metrics.md';
 import ConfigMetric from '/snippets/_sl-configure-metricflow.md';
--- a/website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
+++ b/website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
@@ -8,6 +8,8 @@ meta:
   api_name: dbt Semantic Layer APIs
 ---
 
+### test <Lifecycle status="team,enterprise"/>
+
 import CreateModel from '/snippets/_sl-create-semanticmodel.md';
 import DefineMetrics from '/snippets/_sl-define-metrics.md';
 import ConfigMetric from '/snippets/_sl-configure-metricflow.md';
--- a/website/src/components/lifeCycle/index.js
+++ b/website/src/components/lifeCycle/index.js
@@ -22,8 +22,10 @@ export default function Lifecycle(props) {
   if (!props.status || (Array.isArray(props.status) && props.status.length === 0)) {
     return null;
   }
-  // Check if props.status is an array or a single value
-  const statuses = Array.isArray(props.status) ? props.status : [props.status];
+  // Check if props.status is an array or a single value amd to handle strings separated by commas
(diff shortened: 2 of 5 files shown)


website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
```diff
@@ -8,6 +8,8 @@ meta:
   api_name: dbt Semantic Layer APIs
 ---
 
+### test <Lifecycle status="team,enterprise"/>
```
mirnawong1: tessting to see if build comes out right. will remove

mirnawong1:
deploy preview works: https://docs-getdbt-com-git-test-dbt-labs.vercel.app/docs/use-dbt-semantic-layer/quickstart-sl#prerequisites

website/docs/docs/use-dbt-semantic-layer/quickstart-sl.md
mirnawong1:
```suggestion
```

website/src/components/lifeCycle/index.js
```diff
@@ -22,8 +22,10 @@ export default function Lifecycle(props) {
   if (!props.status || (Array.isArray(props.status) && props.status.length === 0)) {
     return null;
   }
-  // Check if props.status is an array or a single value
-  const statuses = Array.isArray(props.status) ? props.status : [props.status];
+  // Check if props.status is an array or a single value amd to handle strings separated by commas
+  const statuses = typeof props.status ==='string'
+  ?props.status.split(',').map(s => s.trim()) 
+  : Array.isArray(props.status) ? props.status : [props.status];
```
JKarlavige: Now that we're always passing in a string, there's a couple adjustments we can make:

You're on the right track with using `.split(',')`! I recommend updating this section and simply setting `const statuses = props?.status?.split(',')`. This will always create an array, even if one item is passed in without a comma.

For example, passing in `status="enterprise"` will set `props?.status?.split(',')` to `['enterprise']`. If the `status` prop isn't set on a component, `statuses` will become undefined.

I then recommend moving this line above the `if` check on line 22. We can then update the if check to: `if(!statuses?.length)`. This will check if `statuses` has a value and is an array. If not, return null. Otherwise apply the html.

JKarlavige (changes requested):
This is close! Left one comment walking through a few adjustments 😁

website/src/components/lifeCycle/index.js
mirnawong1: ok awesome! thank you so much @JKarlavige ! i made those changes. one thing i kept though was 
```
  if (!props.status || (Array.isArray(props.status) && props.status.length === 0)) {
    return null;
  }

```

because if i didn't add a status, the header would display a dash:
Screenshot 2024-02-14 at 18 29 14

JKarlavige: Nice catch, was able to recreate when passing an empty string into the status prop: `status=""`. I merged your if check into the updated one so we only have 1 if check. The `!props.status` is enough to verify if `status` is an empty string.

Added the component to a few headers to double-check the build one more time, if the build passes i'll get this merged!

JKarlavige:
Build completed successfully after the latest updates. Verified the following:

`<Lifecycle status="team,enterprise" />` shows the pricing plans
image

`<Lifecycle status="" />` does not render the component

`<Lifecycle />` does not render the component

`<Lifecycle status="custom" />` renders the component with default bg
image

Also verified the link copying works!

JKarlavige (approved):
:shipit:

PR #4927, opened 2024-02-15
update hover and remove scroll

mirnawong1:
this pr updates the lightbox component and removes the scroll function. currently, i'm seeing the hover even when scrolling and when stopping my scroll state, leading to a weird ux and the hovered images getting in the way of reading. 

this pr removes the scroll function and tries to simplify the component. it als adjusts the scale.


matthewshaver (approved):

mirnawong1:
Thanks @matthewshaver ! @JKarlavige happy to merge it but pls feel free to check if the code is ok 🙏

JKarlavige:
Sorry I wasn't able to get to this one! All looks good from checking out the code here!

PR #4932, opened 2024-02-15
clarify bq qs step after config block change 

mirnawong1:
this pr updates the quickstart snippet (for bigquery specifically) per user feedback [here](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1707955900758119). dbt run isn't needed after adding a model config block for materialization. instead, users shoudl run dbt run --full-refresh for the change to take place.

### DIFF
--- a/website/snippets/quickstarts/change-way-model-materialized.md
+++ b/website/snippets/quickstarts/change-way-model-materialized.md
@@ -54,10 +54,9 @@ By default, everything gets created as a view. You can override that at the dire
 
     </File>
 
-4. Enter the `dbt run` command. Your model, `customers` should now build as a view.
-5. Enter the `dbt run --full-refresh` command for this to take effect in your warehouse.
+4. Enter the `dbt run --full-refresh` command for this to take effect in your warehouse.
 
-#### FAQs
+### FAQs
 
 <FAQ path="Models/available-materializations" />
 <FAQ path="Project/which-materialization" />
--- a/website/snippets/quickstarts/change-way-model-materialized.md
+++ b/website/snippets/quickstarts/change-way-model-materialized.md
@@ -54,7 +54,9 @@ By default, everything gets created as a view. You can override that at the dire
 
     </File>
 
-4. Enter the `dbt run --full-refresh` command for this to take effect in your warehouse.
+4. Enter the `dbt run` command. Your model, customers should now build as a view. 
+   - BigQuery users should run `dbt run --full-refresh` (mentioned in step 5) to ensure the materialization changes are fully applied
+5. Enter the `dbt run --full-refresh` command for this to take effect in your warehouse.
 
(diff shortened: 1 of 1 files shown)


website/snippets/quickstarts/change-way-model-materialized.md
```diff
@@ -54,10 +54,11 @@ By default, everything gets created as a view. You can override that at the dire
 
     </File>
 
-4. Enter the `dbt run` command. Your model, `customers` should now build as a view.
+4. Enter the `dbt run` command. Your model, customers should now build as a view.
```
runleonarun: Not sire if we removed the backticks on purpose?
```suggestion
4. Enter the `dbt run` command. Your model, `customers`, should now build as a view. 
```

website/snippets/quickstarts/change-way-model-materialized.md
```diff
@@ -54,10 +54,11 @@ By default, everything gets created as a view. You can override that at the dire
 
     </File>
 
-4. Enter the `dbt run` command. Your model, `customers` should now build as a view.
+4. Enter the `dbt run` command. Your model, customers should now build as a view. 
+   - BigQuery users should run `dbt run --full-refresh` (mentioned in step 5) to ensure the materialization changes are fully applied
```
runleonarun: Are we saying they shouldn't `dbt run` and instead should `dbt run --full refresh`? If so, we might want to be more specific around that:
```suggestion
   - BigQuery users need to run `dbt run --full-refresh` instead of `dbt run` to full apply materialization changes.
```

runleonarun (commented):
@mirnawong1 some small clarifications. 

An aside, do you think this might feel out-of-place in other warehouse guides?  I think there's no elegant way to add it in just the BQ guide, so I'm not sure there's a better suggestion.

website/snippets/quickstarts/change-way-model-materialized.md
mirnawong1: oh i removed it first and then copied it from the page. thank you for catching!

mirnawong1:
> @mirnawong1 some small clarifications.
> 
> An aside, do you think this might feel out-of-place in other warehouse guides? I think there's no elegant way to add it in just the BQ guide, so I'm not sure there's a better suggestion.

thank you @mirnawong1 ! i do too, but also had a hard time adding it without messing up the snippet. if i add it before, it's too early. if i add it after, it might be missed. that's why i asked whether step 4 was necessary. unless i just remove the snippet for bq?

runleonarun (approved):
@mirnawong1 let's go with this and see if people comment about it. The benefits of a snippet is way higher than it being out of place. Let's use this.

Approved!

mirnawong1:
lol i commented to myself 🙈

thanks @runleonarun !

PR #4968, opened 2024-02-21
add a SL faq addressing data platforms

mirnawong1:
this pr adds an faq that explains what data platforms are supported by SL and which are not. this is due to various audiences asking.

### DIFF
--- a/website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
+++ b/website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
@@ -66,6 +66,13 @@ The dbt Semantic Layer measures usage in distinct 'Queried Metrics'. Refer to th
 
 ## Availability
 
+<expandable alt_header="What data platforms are supported by the dbt Semantic Layer?">
+
+The dbt Semantic Layer supports the following data platforms: Snowflake, BigQuery, Databricks, or Redshift.
+
+Support for other data platforms, such as Fabric and Trino, isn't available at this time. If you're interested in using the dbt Semantic Layer with a different data platform, please reach out to us [here](https://www.getdbt.com/get-started).
+</expandable>
+
 <expandable alt_header="Do I need to be on a specific version of dbt to use dbt Semantic Layer?">
 
 Yes, the dbt Semantic Layer is compatible with [dbt v1.6 or higher](/docs/dbt-versions/upgrade-dbt-version-in-cloud).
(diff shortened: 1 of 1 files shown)


website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
```diff
@@ -66,6 +66,13 @@ The dbt Semantic Layer measures usage in distinct 'Queried Metrics'. Refer to th
 
 ## Availability
 
+<expandable alt_header="What data platforms are supported by the dbt Semantic Layer?">
+
+The dbt Semantic Layer supports the following data platforms: Snowflake, BigQuery, Databricks, or Redshift.
```
matthewshaver: I think we would normally do something like this as a bulleted list, yeah?
```suggestion
The dbt Semantic Layer supports the following data platforms: 

- Snowflake 
- BigQuery 
- Databricks  
- Redshift
```

DevonFulcher (approved):

website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
```diff
@@ -66,6 +66,13 @@ The dbt Semantic Layer measures usage in distinct 'Queried Metrics'. Refer to th
...
+
+Support for other data platforms, such as Fabric and Trino, isn't available at this time. If you're interested in using the dbt Semantic Layer with a different data platform, please reach out to us [here](https://www.getdbt.com/get-started).
```
matthewshaver: Maybe we don't need to call out platforms we *don't* support.
I don't think that link is going to be helpful as a contact link. Is there something more direct we could use like and email address or a link to a page with contact info?

```suggestion
Support for other data platforms, such as Fabric and Trino, isn't available at this time. If you're interested in using the dbt Semantic Layer with a data platform not on the list, please [contact us](https://www.getdbt.com/get-started).
```

website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
mirnawong1: yea its not linking or steps. no strong feelings so pushing it forward and merging the pr 🙌

website/docs/docs/use-dbt-semantic-layer/sl-faqs.md
mirnawong1: i think it's helpful to be explicit and not encourage inferring, personally imho. esp since users/sa's asked about it. 

i like the mesh of:
> Support for other data platforms, such as Fabric and Trino, isn't available at this time. If you're interested in using the dbt Semantic Layer with a data platform not on the list, please [contact us](https://www.getdbt.com/get-started).

matthewshaver (approved):

PR #4971, opened 2024-02-22
Suggest >= consistently in incremental models

skion:

The current docs sometimes suggest to use `>` and sometimes `>=` when comparing last-update date(times) in incremental models. Using `>` could probably lead to gaps in the dataset in certain cases, but even if not, then still safer to always suggest using `>=` in the documentation.

## Checklist

N/A


mirnawong1 (approved):
hey @skion great suggestion and you're right about the consistency! really appreciate you opening this pr up and looking forward to future contributions ✨!

PR #4976, opened 2024-02-22
agg_time_dimenson required if the model contains measures

Jstein77:
Docs we're incorrect - agg_time_dimenson is required if the model contains measures, not dimensions.


mirnawong1 (approved):
approving this and thank you @Jstein77 !

PR #5079, opened 2024-03-14
Fix example for configuring a custom alias in YAML

dbeatty10:

The code example is not valid, so we should update it to align with the code example [here](https://docs.getdbt.com/reference/model-configs).

## Checklist
- [x] Test out the code example to make sure it works 😎


mirnawong1 (approved):
🚢

PR #5100, opened 2024-03-18
`conversion` is a valid metric `type`

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-3-dbt-labs.vercel.app/docs/build/metrics-overview)

Since [`conversion`](https://docs.getdbt.com/docs/build/conversion) is a valid `type`, we should add it as an option in the docs.

> [!NOTE]  
> I'm not sure if version blocks apply here, because I don't know if the `conversion` has always existed or if it was added after other types.

This is the section of the page being updated:

image

## Checklist

### DIFF
--- a/website/docs/docs/build/metrics-overview.md
+++ b/website/docs/docs/build/metrics-overview.md
@@ -19,7 +19,7 @@ The keys for metrics definitions are:
 | --------- | ----------- | ---- |
 | `name` | Provide the reference name for the metric. This name must be unique amongst all metrics.   | Required |
 | `description` | Describe your metric.   | Optional |
-| `type` | Define the type of metric, which can be `simple`, `ratio`, `cumulative`, or `derived`.  | Required |
+| `type` | Define the type of metric, which can be `simple`, `ratio`, `cumulative`, `conversion`, or `derived`.  | Required |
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/metrics-overview.md
```diff
@@ -19,7 +19,7 @@ The keys for metrics definitions are:
 | --------- | ----------- | ---- |
 | `name` | Provide the reference name for the metric. This name must be unique amongst all metrics.   | Required |
 | `description` | Describe your metric.   | Optional |
-| `type` | Define the type of metric, which can be `simple`, `ratio`, `cumulative`, or `derived`.  | Required |
+| `type` | Define the type of metric, which can be `simple`, `ratio`, `cumulative`, `conversion`, or `derived`.  | Required |
```
mirnawong1:
```suggestion
| `type` | Define the type of metric, which can be `conversion`, `cumulative`, `derived`, `ratio`, or `simple`. | Required |
```

mirnawong1 (approved):
amazing catch @dbeatty10 ! youre so right - also updated the order so it aligned with the right toc - thank you so much for catching this!!

PR #5102, opened 2024-03-18
add note about ip address allow list

mirnawong1:
this pr adds a sentence in the gsheets integration for users that have IP restrictions. they need to add google's IP address to their IP allowlist to allow the connection to flow through. otherwise it'll fail. 

[slack thread. and request](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1710770314300909)


DevonFulcher (approved):
Thank you!

Issue #5111, opened 2024-03-19
[Core] Fix content in table for available flags (global configs)


dbeatty10:
### Link to the page(s) on docs.getdbt.com requiring updates

https://docs.getdbt.com/reference/global-configs/about-global-configs#available-flags

### Tell us more about this update

Here's what it is currently:

|    Flag name   |  Type  |               Default               | Supported in project? | Environment variable | Command line option | Supported in Cloud CLI? |
|:--------------:|:------:|:-----------------------------------:|:---------------------:|:--------------------:|:-------------------:|:-----------------------:|
| `target_path`  | path   | none                                | ❌                     | `DBT_TARGET_PATH`    | `--target-path`     | ❌                       |
| `profiles_dir` | path   | None (current dir, then `HOME` dir) | ❌                     | `DBT_PROFILES_DIR`   | `--profiles-dir`    | ❌                       |
| `target`       | path   | None (uses `target/`)               | ❌                     | `DBT_TARGET`         | `--target`          | ❌                       |
| `profile`      | string | None                                | ✅ (as top-level key)  | `DBT_PROFILE`        | `--profile`         | ❌                       |

Five things are not accurate:
1. Default for `target_path`
1. Default for `target`
1. Type for `target`
1. `DBT_TARGET` won't be available until https://github.com/dbt-labs/dbt-core/pull/9081 is merged+released
1. `DBT_PROFILE` won't be available until https://github.com/dbt-labs/dbt-core/pull/9081 is merged+released

Here's what it should be instead (until https://github.com/dbt-labs/dbt-core/pull/9081 is merged+released):

|    Flag name   |  Type  |               Default               | Supported in project? | Environment variable | Command line option | Supported in Cloud CLI? |
|:--------------:|:------:|:-----------------------------------:|:---------------------:|:--------------------:|:-------------------:|:-----------------------:|
| `target_path`  | path   | None (uses `target/`)               | ❌                     | `DBT_TARGET_PATH`    | `--target-path`     | ❌                       |
| `profiles_dir` | path   | None (current dir, then `HOME` dir) | ❌                     | `DBT_PROFILES_DIR`   | `--profiles-dir`    | ❌                       |
| `target`       | path   | None                                | ❌                     |                      | `--target`          | ❌                       |
| `profile`      | string | None                                | ✅ (as top-level key)  |                      | `--profile`         | ❌                       |

### Reviewers/Stakeholders/SMEs

Doug E. Fresh 😎

[...]

PR #5112, opened 2024-03-19
Fix default for `target` and `target_path`

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/reference/global-configs/about-global-configs#available-flags)

See https://github.com/dbt-labs/docs.getdbt.com/issues/5111

These are not accurate:

1. Default for `target_path`
1. Default for `target`
1. Type for `target`

The problem for the first two is that the content is flip-flopped. So this PR just does that swap.

The problem for the last one is that `target` also should be `string`, not `path` (as mentioned by https://github.com/dbt-labs/docs.getdbt.com/issues/5111#issuecomment-2007409988).

## 🎩

Here's what it looks like after the fixes in this PR:
image

## Checklist


mirnawong1 (approved):
✨

PR #4980, opened 2024-02-23
update docs to reflect main git enhancements

mirnawong1:
this pr updates the following pages to account for the new and improved main branch workflow in the ide:

- IDE git button: https://arc.net/l/quote/kgbayfdq
- IDE UI: https://arc.net/l/quote/edletcth
- IDE linting: https://docs.getdbt.com/docs/cloud/dbt-cloud-ide/lint-format#format-sql
- quickstart workflow: https://arc.net/l/quote/okqvcakv
- possibly this faq; https://docs.getdbt.com/faqs/Environments/custom-branch-settings#development

[notion project](https://www.notion.so/dbtlabs/DevExp-Enhancements-CLI-IDE-Git-AI-9d72cc68bcc44a6bafe95eab793235fb)

PR #5129, opened 2024-03-21
fix qs code and remove snippet

mirnawong1:
this pr reverts changes from pr #4980.

the db varies from warehouse to warehouse and a snippet isn't appropriate for this. re-adding the code along with the correct database/dataset info.

also raised by user ticket and [discourse](https://discourse.getdbt.com/t/new-user-trying-dbt-building-customres-model-and-getting-error-the-project-raw-has-not-enabled-bigquery/12497)


mirnawong1:
- [preview for redshift](https://docs-getdbt-bmta56i5f-dbt-labs.vercel.app/guides/redshift?step=7)
- [preview for databricks](https://docs-getdbt-bmta56i5f-dbt-labs.vercel.app/guides/databricks?step=7)
- [preview for bigquery](https://docs-getdbt-bmta56i5f-dbt-labs.vercel.app/guides/bigquery?step=8)
- [preview for snowflake](https://docs-getdbt-bmta56i5f-dbt-labs.vercel.app/guides/snowflake?step=7)

mirnawong1:
tested and merging pr so users don't run into this while testing dbt or doing the quickstart.

website/docs/guides/bigquery-qs.md
```diff
@@ -111,9 +111,80 @@ Now that you have a repository configured, you can initialize your project and s
 
 ## Build your first model
 
-import BuildFirstModel from '/snippets/quickstarts/_build-your-first-model.md';
+You have two options for working with files in the dbt Cloud IDE:
 
-<BuildFirstModel/>
+- Edit in the protected primary branch &mdash; Edit, format, or lint files and execute dbt commands directly in your primary git branch. Since the dbt Cloud IDE prevents commits to the protected branch, it prompts you to commit those changes to a new branch.
+
+- Create a new branch before editing &mdash; If you prefer to keep the primary branch unchanged, create a new branch before starting your edits. Navigate to **Version Control** on the left sidebar and click **Create branch**.
```
matthewshaver: Editing in the primary branch is a nice-to-have feature that we implemented to address customer feedback, but I think we should be presenting this option as the primary (default) option that everyone should use rather than a "preference." It should probably be the other way around. Especially because editing in the primary branch will eventually result in creating a new one any way via extra steps.

mirnawong1: thanks Matt! im not sure i understand what you mean here. do you think we should only keep the first bullet as the only option? 

i left both options since this quickstart is an intro and not everyone is familiar with git.

matthewshaver: I think we have both, but the way this is worded sort of presents editing in the protected branch as the preferred method and creating a new branch as a secondary option. It should probably be the other way around.
"If you prefer to keep the primary branch unchanged" is a little misleading because they can't actually change the primary branch. They can "edit" on it but we'll force them to create a new branch before they push any changes.

So something like:

- Create a new branch (recommended) &mdash; Create a new branch to edit and commit your changes. Navigate to **Version Control** on the left sidebar and click **Create branch**.

- Edit in the protected primary branch &mdash; If you prefer to edit, format, or lint files and execute dbt commands directly in your primary git branch. The dbt Cloud IDE prevents commits to the protected branch, so you will be prompted to commit your changes to a new branch.

matthewshaver (approved):

mirnawong1:
previews all look good!

PR #5135, opened 2024-03-22
clarify dbt show is only for SQL models - not python

### LINKED ISSUE #5083: [Core] `dbt show` is only for SQL models (not Python)
### Link to the page(s) on docs.getdbt.com requiring updates

https://docs.getdbt.com/reference/commands/show

### Tell us more about this update

The docs do specify dbt-SQL in a couple places, but it would be nice to explicitly call out that [dbt-py](https://docs.getdbt.com/docs/build/python-models) (Python) models are not supported by `dbt show`.

### Reviewers/Stakeholders/SMEs

Kshitij

### Related GitHub issues

_No response_

### Additional information

_No response_

mirnawong1:
this pr is meant to explicitly say that dbt show supports dbt sql and not dbt-py python models.

Closes #5083


dbeatty10 (approved):
Looks great

PR #5152, opened 2024-03-25
Update the example commands for the "state" selector method

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-4-dbt-labs.vercel.app/reference/node-selection/methods#the-state-method)

Two things:
1. All the examples need a `--state` flag (similar to the "result" selector method)
1. One example in particular doesn't work because of an extra space

### More detail

Here's the error I get when the `--state` flag is not included:
```
01:02:43  Encountered an error:
Runtime Error
  Got a state selector method, but no comparison manifest
```

Here's the error I get with the extra space:
```
01:11:01  The selection criterion '' does not match any nodes
```

## Checklist

### DIFF
--- a/website/docs/reference/node-selection/methods.md
+++ b/website/docs/reference/node-selection/methods.md
@@ -222,7 +222,7 @@ The `state` method is used to select nodes by comparing them against a previous 
 
 
   ```bash
-dbt test --select "state:new "           # run all tests on new models + and new tests on old models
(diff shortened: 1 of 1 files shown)


mirnawong1 (approved):
oh that's a great catch, thank you for opening this up @dbeatty10 !

PR #5156, opened 2024-03-25
add example to redshift grants section

mirnawong1:
raised in the [slack communit](https://getdbt.slack.com/archives/C0441GSRU04/p1710879707806589), this PR sets to add info that grants roles/groups is supported for redshift. 

raised in [internal slack](https://dbt-labs.slack.com/archives/C067EUS2D7S/p1711037112812519) and also used [discourse](https://discourse.getdbt.com/t/how-to-grant-privilege-to-groups-using-grant-redshift/7507/3) as sourc.e


b-per:
I am not the expert here, but based on my quick testing I don't think that this is 100% correct.

My understanding is that

- for users we support the `+` syntax for `grants`, and in that case, we just need to list the users, e.g. `select: ["bper"]`
- for roles and groups
  - we **don't** support the `+` for `grants`, so we always need to provide the full list of roles/groups that need acess
  - and we have to prefix the role/group by role or group, e.g. `select: ["role my_role", "group my_group"]`

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,20 @@ models:
 
 <div warehouse="Redshift">
 
-* Granting to / revoking from is only fully supported for Redshift users (not groups or roles).
+For Redshift, you need to explicitly grant roles and groups access to your models, whether to single users, groups, or roles, and you can revoke access as well.
```
matthewshaver: Would this be better suited here with a link from the note https://docs.getdbt.com/reference/resource-configs/redshift-configs or maybe at least also there?

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,38 @@ models:
...
+    description: "My schema"
+    +grants:
+      select: ["gspider"]  # example of user access
+```
+
+**For roles and groups:**
+For roles and groups, the `+grants` syntax is not supported. You need to provide the complete list of roles or groups that require access each time you make a change. Additionally, you must prefix each role or group with 'role' or 'group':
+
+```
+yaml
```
mirnawong1:
```suggestion
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,37 @@ models:
...
+    schema: my_schema
+    description: "My schema"
+    +grants:
+      select: ["gspider"]  # example of user access
+```
+
+**For roles and groups:**
+For roles and groups, the `+grants` syntax is not supported. You need to provide the complete list of roles or groups that require access each time you make a change. Additionally, you must prefix each role or group with 'role' or 'group':
+
+```
```
mirnawong1:
```suggestion
```yaml
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,37 @@ models:
...
+models:
+  my_schema:
+    schema: my_schema
+    description: "My schema"
+    +grants:
+      select: ["gspider"]  # example of user access
+```
+
+**For roles and groups:**
+For roles and groups, the `+grants` syntax is not supported. You need to provide the complete list of roles or groups that require access each time you make a change. Additionally, you must prefix each role or group with 'role' or 'group':
```
dataders: I don't think that the `+` is what is not supported here, rather, what isn't supported is the [inheritance of grants](https://docs.getdbt.com/reference/resource-configs/grants#grant-config-inheritance) 

```suggestion
Inheritance of grants are not supported for groups and roles in Redshift. Accordingly, for every model, you need to provide the complete list of roles or groups that require access each time you make a change to a model.

Additionally, must prefix each role or group with 'role' or 'group':
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,39 @@ models:
...
+    schema: my_schema
+    description: "My schema"
+    +grants:
+      select: ["gspider"]  # example of user access
+```
+
+**For roles and groups:**
+Inheritance of grants are not supported for groups and roles in Redshift. Accordingly, for every model, you need to provide the complete list of roles or groups that require access each time you make a change to a model.
+
+Additionally, must prefix each role or group with 'role' or 'group':
```
mirnawong1:
```suggestion
Additionally, you must prefix each role or group with 'role' or 'group':
```

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,39 @@ models:
...
+    schema: my_schema
+    description: "Schema for reporting"
+    grants:
+      select: ["role my_role", "group my_group"]
+```
+
+Some things to note:
+
+- Ensure that any roles or groups mentioned must already exist in Redshift.
+- Use the keyword `+grants` to add to or modify users for existing grants (rather than replace), however this syntax does not apply to roles and groups.
```
dataders: I still don't think that the `+grants` does anything different than `grants`. The important caveat is really my edit. 

```suggestion
- Adding or modifying users for existing grants (rather than replace), is not supported for roles and groups.
```

dataders: edit -- I'm totally wrong here after reading the Notes section at the end of the 
[grants: Grant config inheritance](https://docs.getdbt.com/reference/resource-configs/grants#grant-config-inheritance)

Now, I think this is the right language.

```suggestion
- The merge strategy of grants (via the `+` prefix in front of `+grants`) is not supported for roles and groups.
```

dataders (changes requested):
only one small change!

mirnawong1:
thanks so much @dataders ! i think I've folded in all your suggestions now 🙏

website/docs/reference/resource-configs/grants.md
```diff
@@ -249,7 +249,39 @@ models:
...
+**For users:**
+When granting access to individual users, use the `+grants` syntax in your configuration to add or modify permissions for these users without replacing existing configurations. In the following example, you only need to list the users:
+
+```yaml
+models:
+  my_schema:
+    schema: my_schema
+    description: "My schema"
+    +grants:
+      select: ["gspider"]  # example of user access
```
mirnawong1, line 263: 👻 🕷️ 🕸️

dataders (approved):
great! let's merge this! though as a reminder, https://github.com/dbt-labs/dbt-redshift/pull/626 
will be worked on in the next month which will require another pass at this page. Hopefully though, the result is the docs become simpler with less caveats

mirnawong1:
as discussed in slack thread](https://dbt-labs.slack.com/archives/C03SK9W08BC/p1714506360050039), closing this pr as this is not the intended behavior. see https://github.com/dbt-labs/dbt-adapters/issues/647 for more info.

PR #5174, opened 2024-03-27
Update snowflake-setup.md

mirnawong1:
this pr updates the key pair auth section of the snowflake docs and updates it to the following

- private_key support added in 1.7 (not v1.5)
- v1.5 and 1.6 still need private_key_path (currently docs says you don't need it for v1.5+)

stemmed from slack post from Sid Patel [here](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1711394087902749)


sidpatel13 (approved):
Thank you,  this LGTM!

PR #5180, opened 2024-03-27
FIx spacing to avoid horizontal scroling

matthewshaver:

Reason: When guides have large photos, videos, tables, or code blocks they would sometimes go off screen, requiring the user to scroll horizontally. In some of the Guides this was more egregious than others, but an all around negative for the customer experience.

The following screenshots are cropped but they show the text running off-screen on the right hand side:

Screenshot 2024-03-27 at 4 10 07 PM

Screenshot 2024-03-27 at 4 10 00 PM

Screenshot 2024-03-27 at 4 09 38 PM

Solution: Adding a max width to each of the guides, regardless of whether or not they were impacted (future proofing them if iterations add those elements. The width set to 900px and brings the text to the edge but gives a little buffer. This also prevents code blocks and other elements from extending the beyond the border. Users will now scroll within the text block itself, rather than the whole screen. This auto-sizes pictures and videos so they fit on the screen.

Limitations:
- Doesn't have an impact on tables. For tables to resize, we're going to need some custom code for our site (which I believe is in the queue)
- The closing </div> bracket can't be placed before snippets with no headers (the header is in the snippet) or it causes a glitch where the text won't appear on the screen for those sections (despite them still appearing on the menu). This is resolved by moving the closing bracket above these sections. It only impacted adapter docs that use the same snippets at the end. 

## Checklist

- [x] For [docs versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#about-versioning), review how to [version a whole page](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) and [version a block of content](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#versioning-blocks-of-content).


mirnawong1 (approved):
lgtm, tested it and working well! thanks ya'll!

PR #5186, opened 2024-03-28
Use the full configuration name (`loaded_at_field`)

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/docs/core/connect-data-platform/about-core-connections)

The property is named [`loaded_at_field`](https://docs.getdbt.com/reference/resource-properties/freshness#complete-example) rather than `loaded_at`.

## Checklist
- [x] I have confirmed the code example works by trying it out


website/docs/docs/core/connect-data-platform/about-core-connections.md
```diff
@@ -39,20 +39,20 @@ The following table lists the features available for adapters:
...
 
 ### Source freshness
-You can measure source freshness using the warehouse metadata tables on supported adapters. This allows for calculating source freshness without using the `loaded_at` field and without querying the table directly. This is faster and more flexible (though it might sometimes be inaccurate, depending on how the warehouse tracks altered tables). You can override this with the `loaded_at` field in the model config. If the adapter doesn't support this, you can still use the `loaded_at` field.
+You can measure source freshness using the warehouse metadata tables on supported adapters. This allows for calculating source freshness without using the `loaded_at_field` and without querying the table directly. This is faster and more flexible (though it might sometimes be inaccurate, depending on how the warehouse tracks altered tables). You can override this with the `loaded_at_field` in the model config. If the adapter doesn't support this, you can still use the `loaded_at_field`.
```
mirnawong1: non blocker but wondering if we should link to it here? or wrong link?

```suggestion
You can measure source freshness using the warehouse metadata tables on supported adapters. This allows for calculating source freshness without using the [`loaded_at_field`](/reference/resource-properties/freshness#loaded_at_field) and without querying the table directly. This is faster and more flexible (though it might sometimes be inaccurate, depending on how the warehouse tracks altered tables). You can override this with the `loaded_at_field` in the model config. If the adapter doesn't support this, you can still use the `loaded_at_field`.
```

mirnawong1 (approved):
great catch, thank you @dbeatty10 !

matthewshaver (approved):

PR #5227, opened 2024-04-04
Fix syntax in SL / MF Dimensions page

gwenwindflower:
Noticed some more old syntax in the SL/MF docs, this should fix it (I think I've looked at every metricflow/SL page in the past 24 hours so this is hopefully it!)


mirnawong1 (approved):
🚢

PR #5236, opened 2024-04-05
Add "pill" property to Card component

john-rock:
- Adds `pills` prop to card component
- Renders pills on frontend in the same style we have on the guides page
- This will remove the need to pass pills as html within the `body` property like we currently are.
- This PR also contains the work to migrate the Trusted Adapter cards to use the new `pills` prop.
image

## How to use
Pills must be added as strings within an array on the card component.
The array must be wrapped in single quotes and each pill item must be wrapped in double quotes speerated by a comma

For example: `pills='["dbt Cloud", "dbt Core"]'`

## Preview
https://docs-getdbt-com-git-add-card-pills-dbt-labs.vercel.app/docs/trusted-adapters

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206847403325348


JKarlavige (approved):
lgtm :shipit:

nghi-ly (commented):
@john-rock : i just looked at the preview -- this looks great! i like the sizing and font changes you made :chefs-kiss:

happy to approve but want to double-check, is there anything specific you'd like me to test (besides looking at preview)?

nghi-ly (approved):
tysm for adding this ability! 

approving in case i'm blocking you! 

if there are tests you want me to do, happy to assist!

PR #5246, opened 2024-04-08
Beware custom `fail_calc` in tests that have the potential to return no rows at all

### LINKED ISSUE #918: `fail_calc` config docs should warn users not to use a sum() fail_calc in their test query if that query could conceivably return no rows at all.
## Steps to reproduce:

This is covered off in [this issue](https://github.com/dbt-labs/dbt-core/issues/4248).

Basically, write a test that returns no rows and uses a `sum()` function for its fail_calc. It will crash.

## Expected results:

When running a test, we generally expect it to `pass` if no bad records are found, and `fail/warn` if bad records are found. If we've used bad SQL or the database is in a bad/unusual state, we might sometimes get a `database error`.

## Actual results:

When a test is told to pass or fail based on the sum of something, it cannot either pass or fail if no rows are returned. It cannot do its job because it does not have what it needs in order to do its job (what it needs is a result set to speak of).

## Extra details:

This is a `wontfix` as far as the way `dbt` itself works, which makes a bunch of sense. Given that it's not really a bug, the best place to fix it is in the docs:

If users who are thinking of using a custom `fail_calc` consult the docs first, they'll see that they shouldn't use a sum() function when they could conceivably get no rows back. 

If they don't consult the docs first, and end up running into this error, then the fact that there is something explicit in the docs about it will help them to figure it out and fix it faster than they may otherwise.

[The docs](https://docs.getdbt.com/reference/resource-configs/fail_calc) currently tell users that they can use a sum() function:

```For instance, you can configure a unique test to return sum(n_records) instead of count(*) as the failure calculation: that is, the number of rows in the model containing a duplicated column value, rather than the number of distinct column values that are duplicated.```

I think there should be a paragraph after that -- something explicit and to the point -- along the lines of (but not necessarily word-for-word):

```
Beware that a sum() function should not be used as a fail_calc in any test that has the potential to return no rows at all.

Such a test would break (neither pass nor fail), and return the following error, because it cannot sum(null): 

None is not of type 'integer'

Failed validating 'type' in schema['properties']['failures']:
    {'type': 'integer'}

On instance['failures']:
    None
```

### PR DESCRIPTION
**Pull request by dbeatty10**
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/reference/resource-configs/fail_calc)

Follow-up to #5221

resolves #918

See also: https://github.com/dbt-labs/dbt-core/issues/4248

## Checklist

### DIFF
--- a/website/docs/reference/resource-configs/fail_calc.md
+++ b/website/docs/reference/resource-configs/fail_calc.md
@@ -14,7 +14,7 @@ For instance, you can configure a `unique` test to return `sum(n_records)` inste
 :::tip Tip
 Beware using functions like `sum()` for `fail_calc` in any test that has the potential to return no rows at all.
 
-Such a test would break (neither pass nor fail), and return the following error: 
+If no rows are returned, the test won't pass or fail but will return the following error: 
 
 ```
 None is not of type 'integer'
--- a/website/docs/reference/resource-configs/fail_calc.md
+++ b/website/docs/reference/resource-configs/fail_calc.md
@@ -11,6 +11,29 @@ Most tests do not use the `fail_calc` config, preferring to return a count of fa
 
 For instance, you can configure a `unique` test to return `sum(n_records)` instead of `count(*)` as the failure calculation: that is, the number of rows in the model containing a duplicated column value, rather than the number of distinct column values that are duplicated.
 
+:::tip Tip
+Beware using functions like `sum()` for `fail_calc` in any test that has the potential to return no rows at all.
+
+Such a test would break (neither pass nor fail), and return the following error: 
+
+```
+None is not of type 'integer'
+
+Failed validating 'type' in schema['properties']['failures']:
+    {'type': 'integer'}
+
+On instance['failures']:
+    None
+```
+
+This scenario can be handled with a case statement that returns `0` when there are no rows:
+
+```yaml
+fail_calc: "case when count(*) > 0 then sum(n_records) else 0 end"
+```
+
+:::
+
 <Tabs
   defaultValue="specific"
   values={[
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: website/docs/reference/resource-configs/fail_calc.md [hunk 1]
```diff
@@ -11,6 +11,29 @@ Most tests do not use the `fail_calc` config, preferring to return a count of fa
...
+None is not of type 'integer'
+
+Failed validating 'type' in schema['properties']['failures']:
+    {'type': 'integer'}
+
+On instance['failures']:
+    None
+```
+
+This scenario can be handled with a case statement that returns `0` when there are no rows:
```
mirnawong1:
```suggestion
To avoid this issue, use a case statement to ensure that `0` is returned when no rows exist:
```

website/docs/reference/resource-configs/fail_calc.md
```diff
@@ -11,6 +11,29 @@ Most tests do not use the `fail_calc` config, preferring to return a count of fa
 
 For instance, you can configure a `unique` test to return `sum(n_records)` instead of `count(*)` as the failure calculation: that is, the number of rows in the model containing a duplicated column value, rather than the number of distinct column values that are duplicated.
 
+:::tip Tip
+Beware using functions like `sum()` for `fail_calc` in any test that has the potential to return no rows at all.
+
+Such a test would break (neither pass nor fail), and return the following error:
```
mirnawong1:
```suggestion
If no rows are returned, the test won't pass or fail but will return the following error: 
```

mirnawong1 (approved):
small tweaks and approved for merged! thank you again @dbeatty10 !

PR #5255, opened 2024-04-09
Update databricks-configs.md so that only 1.7 content shows on 1.7

benc-db:
Minor change: 1.6 and 1.7 tables were both showing on 1.7


mirnawong1 (approved):
great catch, thanks for adding this @benc-db ! ✨

PR #5265, opened 2024-04-11
Update exports.md

mirnawong1:
add add'l info about exports and mf server data flow per [slack thread](https://dbt-labs.slack.com/archives/C03KHQRQUBX/p1712795602018859)

### DIFF
--- a/website/docs/docs/use-dbt-semantic-layer/exports.md
+++ b/website/docs/docs/use-dbt-semantic-layer/exports.md
@@ -192,7 +192,7 @@ When you run a build job, any saved queries downstream of the dbt models in that
 
 </VersionBlock>
 
-2. After dbt finishes building the models, the MetricFlow Server processes the exports, compiles the necessary SQL, and executes this SQL against your data platform.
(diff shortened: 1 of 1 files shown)


website/docs/docs/use-dbt-semantic-layer/exports.md
```diff
@@ -192,7 +192,7 @@ When you run a build job, any saved queries downstream of the dbt models in that
 
 </VersionBlock>
 
-2. After dbt finishes building the models, the MetricFlow Server processes the exports, compiles the necessary SQL, and executes this SQL against your data platform.
+2. After dbt finishes building the models, the MetricFlow Server processes the exports, compiles the necessary SQL, and executes this SQL against your data platform. It directly executes a "create table" statement, so the data stays within your data platform.
```
nghi-ly: sugg: comma is unnecessary in this case

```suggestion
2. After dbt finishes building the models, the MetricFlow Server processes the exports, compiles the necessary SQL, and executes this SQL against your data platform. It directly executes a "create table" statement so the data stays within your data platform.
```

nghi-ly (approved):

PR #5287, opened 2024-04-16
add more detailed steps and warnings to Databricks PL doc

dhaworth:

Adding additional detail to the Databricks PrivateLink doc around setup on the Databricks side. A customer had a bad time setting this up because of some unknown limitations on how to configure PL within Databricks. We're trying to do a better job of illuminating those here to avoid future customers running into the same limitations.

The bullets and warning message are what's new. I used an `:::information` instead of a `:::caution` because the quoted text shows up white and wasn't readable against the light yellow caution box
Screenshot 2024-04-19 at 9 00 01 AM


website/docs/docs/cloud/secure/databricks-privatelink.md
```diff
@@ -28,7 +28,17 @@ import PrivateLinkSLA from '/snippets/_PrivateLink-SLA.md';
...
+  - [Create a Private Access Settings object](https://docs.databricks.com/en/security/network/classic/private-access-settings.html) &mdash; Create a Private Access Settings (PAS) object with your desired public access settings, and setting Private Access Level to **Endpoint**. Choose the registered endpoint created in the previous step.
+  - [Create or update your workspace](https://docs.databricks.com/en/security/network/classic/privatelink.html#step-3d-create-or-update-the-workspace-front-end-back-end-or-both) &mdash; Create a workspace, or update your an existing workspace. Under **Advanced configurations → Private Link** choose the private access settings object created in the previous step.
+
+:::note Warning
+If using an existing Databricks workspace, all workloads running in the workspace will need to be stopped to enable Private Link! Workloads also cannot be started for another 20 minutes after changes are made. From the [Databricks documentation](https://docs.databricks.com/en/security/network/classic/privatelink.html#step-3d-create-or-update-the-workspace-front-end-back-end-or-both):
```
matthewshaver:
```suggestion
If using an existing Databricks workspace, all workloads running in the workspace will need to be stopped to enable Private Link! Workloads also cannot be started for another 20 minutes after making changes. From the [Databricks documentation](https://docs.databricks.com/en/security/network/classic/privatelink.html#step-3d-create-or-update-the-workspace-front-end-back-end-or-both):
```

matthewshaver (approved):

PR #5293, opened 2024-04-17
Adding profile and target

### LINKED ISSUE #5134: New environment variables: `DBT_PROFILE` and `DBT_TARGET`
# Docs Changes Needed from dbt-core PR #9081

### Pull request

https://github.com/dbt-labs/dbt-core/pull/9081

### Feature request

https://github.com/dbt-labs/dbt-core/issues/7798

### Page to update

https://docs.getdbt.com/reference/global-configs/about-global-configs#available-flags

### What's new

Starting in v1.8 (but not earlier), there are two new environment variables for two flags:
1. `DBT_PROFILE` (corresponding to `--profile`)
1. `DBT_TARGET ` (corresponding to `--target`)

### How to document this feature

Essentially re-add the contents that were removed in this PR: #5113

No additional prose or text is needed -- just an update to the table (with appropriate versioning blocks as-needed / desired).

### Side notes

We don't need to add any additional prose or call-out for the following because they are covered by the content in the table. But just to be share further detail here as a side note:

- `profile` _can_ be specified as a top-level in [`dbt_project.yml`](https://docs.getdbt.com/reference/dbt_project.yml) (but is [last in the precedence order](https://docs.getdbt.com/reference/global-configs/about-global-configs#setting-flags) behind the CLI flag and environment variable)
- `target` can **not** be specified as a top-level in [`dbt_project.yml`](https://docs.getdbt.com/reference/dbt_project.yml)

matthewshaver:

Closes https://github.com/dbt-labs/docs.getdbt.com/issues/5134


mirnawong1 (approved):
🚢 @matthewshaver

mirnawong1 (approved):

PR #5309, opened 2024-04-19
[DO NOT MERGE YET] Add metrics as dimensions docs

Jstein77:
[ ] waiting on eng to confirm PR

This PR adds docs for the metrics as dimension feature. I also need to update the filter docs to show they syntax for referencing metrics in a filter as follows. 

| filter | You can optionally add a filter string to any metric type, applying filters to dimensions, entities, time dimensions, or other metrics during metric computation. Consider it as your WHERE clause. | Optional |
| --- | --- | --- |

```yaml
metrics:
  - name: metric name                     ## Required
    description: same as always           ## Optional
    type: the type of the metric          ## Required
    type_params:                          ## Required
      - specific properties for the metric type
    config: here for `enabled`            ## Optional
    meta:
        my_meta_direct: 'direct'           ## Optional
    label: The display name for your metric. This value will be shown in downstream tools. ## Required
    filter: |                             ## Optional            
      {{  Dimension('entity__name') }} > 0 and {{ Dimension(' entity__another_name') }} is not
      null and {{ Metric('metric_name', group_by=['entity']) }} > 5
```

## Filters[](https://docs.getdbt.com/docs/build/metrics-overview#filters)

A filter is configured using Jinja templating. Use the following syntax to reference entities, dimensions, time dimensions , or Metrics in filters ( Link to metrics page):

`filter: |  {{ Entity('entity_name') }}` 

`filter: |  {{ Dimension('primary_entity__dimension_name') }}`

`filter: |  {{ TimeDimension('time_dimension', 'granularity') }}`

`filter: |  {{ Metric('metric_name', group_by=['Entity']) }}`


website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,163 @@
...
+First, the filter `{{ Metric('data_models', group_by=['account']) }}` will generate SQL similar to the `data_models_per_user` sub query shown above.
+
+```sql
+select
+  sum(1) as data_model_runs
+  , account
+ from data_model_runs
+ group by 
+	 account
+```
```
mirnawong1: should this be `{{ ref('data_model_runs') }}`?

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,163 @@
...
+
+```sql
+select
+	sum(1) as activated_accounts
+from accounts
+left join (
+select
+    sum(1) as data_model_runs
+  , account
+ from data_model_runs
```
mirnawong1: should this be `{{ ref('data_model_runs') }}`?

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,163 @@
...
+   type_params:
+	   measure: accounts
+	 filter: |
+		 {{ Metric('data_model_runs', group_by=['account']) }} > 5
+```
+
+Let’s break down the SQL that will be generated from this metric definition when I run the following query from the CLI `dbt sl query --metrics activated_accounts` 
+
+First, the filter `{{ Metric('data_models', group_by=['account']) }}` will generate SQL similar to the `data_models_per_user` sub query shown above.
+
```
mirnawong1: should this be `{{ Metric('data_model_runs', group_by=['account']) }} > 5`?

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,163 @@
...
+| account | data_model runs |
+| --- | --- |
+| 1 | 4 |
+| 2 | 7 |
+| 3 | 9 |
+| 4 | 1 |
+
+We’ll filter the above table for  accounts with >5 data model runs, then count the number of accounts who meet this criteria.
+
```
mirnawong1: Should we explicity say, "Then, using the `{{ Metric('data_model_runs', group_by=['account']) }}` filter, you canfilter this table to accounts with more than 5...."?

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,163 @@
...
+You can recreate this pattern in MetricFlow by referencing a metric in the `where` filter for another metric using the `Metric()` object syntax. The function for referencing a metric takes a metric name and exactly one entity:
+
+```yaml
+{{ Metric('metric_name', group_by=['Entity']) }}
+```
+
+Using the same `activated_accounts` metric as an example:
+- Create two semantic models, `model_runs` and `accounts`. 
+- Then create a `measure` and `metric` to count data model runs, and another measure to count users.
+- Specify the foreign entity `account` in the `model_runs` semantic model.
```
mirnawong1, line 71: is the below yaml missing a metric calculating data model runs?

website/docs/docs/build/metrics-overview.md
```diff
@@ -75,7 +75,7 @@ metrics:
     label: The display name for your metric. This value will be shown in downstream tools. ## Required
     filter: |                             ## Optional            
       {{  Dimension('entity__name') }} > 0 and {{ Dimension(' entity__another_name') }} is not
-      null
+      null and {{ Metric('metric_name', group_by=['entity']) }} > 5
```
mirnawong1:
```suggestion
      null AND {{ Metric('metric_name', group_by=['entity']) }} > 5
```

website/docs/docs/build/metrics-overview.md
```diff
@@ -227,15 +227,21 @@ metrics:
...
+
+filter: |  
   {{ Dimension('primary_entity__dimension_name') }}
-filter: |
+
+filter: |  
   {{ TimeDimension('time_dimension', 'granularity') }}
+
+filter: |  
+  {{ Metric('metric_name', group_by=['Entity']) }}
```
courtneyholcomb: I would update this to say `entity_name` instead of `Entity` - currently it looks like we might want the `Entity(...)` object syntax, but we just want the name.

website/docs/docs/build/metrics-overview.md
courtneyholcomb: Would also update this to say `entity_name` instead of `entity` (for consistency with what's below)

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+The previous SQL query calculates the number of `activated_accounts` by using the `data_model_runs` metric as a dimension for the user entity. It filters based on the metric value scoped to the account entity. You can express this logic natively in the MetricFlow specification.
+
+## Reference a metric in a filter
+
+Use the `Metric()` object syntax to reference a metric in the `where` filter for another metric. The function for referencing a metric accepts a metric name and exactly one entity:
+
+```yaml
+{{ Metric('metric_name', group_by=['Entity']) }}
```
courtneyholcomb: same here re: `Entity -> entity_name`

website/docs/docs/build/ref-metrics-in-filters.md
courtneyholcomb: It exists! The measure `data_model_runs` has `create_metric: true` so that metric will be created automatically.
Though if we think users aren't aware of that attribute, maybe it would be helpful to add the metric YAML explicitly.

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,164 @@
...
+	```
+
+- MetricFlow joins this query to the query generated by the `accounts` measure on the group by elements and applies the filter conditions:
+
+	```sql
+	select
+		sum(1) as activated_accounts
+	from accounts
+	left join (
+	select
```
courtneyholcomb: Can we indent this subquery?

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,164 @@
...
+sidebar_label: "Metrics as dimensions"
+---
+
+[Metrics](/docs/build/metrics-overview) provide users with valuable insights into their data, such as the number of active users. They also are used to track performance, trend, and help businesses make important decisions. [Dimensions](/docs/build/dimensions), on the other hand, are attributes that help categorize data, such as user type or number of orders placed by a customer.
+
+To make informed business decisions, some metrics need the value of another metric as part of the metric definition, leading us to "metrics as dimensions".
+
+This document explains how you can use metrics as dimensions with metric filters, enabling you to create more complex metrics and gain more insights.
+
+## Use cases
```
courtneyholcomb: Just a note that all of these use cases are not currently supported. Currently we only allow one entity in the group by for the metric filter. The next iteration of this feature will enable these use cases - users will be able to pass multiple entities OR dimensions in the group by for the metric filter. Should we remove these use cases from the docs until they are supported? Or maybe add them to the bottom of the page with a note saying this is coming soon?

courtneyholcomb: (this only applies to the `Use cases` section prior to the `Example` section)

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,164 @@
...
+
+Then, filter this table to accounts with more than 5 data model runs and count the number of accounts that meet this criteria:
+
+| activated_accounts |
+| --- |
+| 2 |
+
+## Considerations when using metrics in filter
+
+- When using a metric filter, ensure the sub-query can join to the outer query without fanning out the result ( unexpectedly increasing the number of rows).
```
courtneyholcomb: nit: remove space after `(`

website/docs/docs/build/ref-metrics-in-filters.md
courtneyholcomb: Maybe we say basically that but referencing `MetricFlow` instead of `you` to maintain consistency with the rest of the section, since this is a continuation. i.e.,
"MetricFlow then filters this table to accounts with more than 5 data model runs."

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,164 @@
...
+
+| activated_accounts |
+| --- |
+| 2 |
+
+## Considerations when using metrics in filter
+
+- When using a metric filter, ensure the sub-query can join to the outer query without fanning out the result ( unexpectedly increasing the number of rows). 
+  - The example that filters the accounts measure using `{{ Metric('data_model_runs', group_by=['account']) }}` is valid because it aggregates the model runs to the account level.
+  - However, filtering by `{{ Metric('data_model_runs', group_by=['model']) }}` isn't valid due to a one-to-many relationship between accounts and model runs, leading to duplicate data.
```
courtneyholcomb: I would update this to say `However, filtering the 'accounts' measure by...`
Currently it's hard to remember what we're filtering and that's crucial to understanding the one-to-many relationship mentioned here.

website/docs/docs/build/ref-metrics-in-filters.md
courtneyholcomb: Hmm not sure it makes sense here since the the SQL generated by MetricFlow will reference a table directly, not a `ref()`.

website/docs/docs/build/ref-metrics-in-filters.md
mirnawong1: oh wow, `create_metric: true` is VERY cool! should we add this in the metrics overview doc? and does it apply to other metrics?

courtneyholcomb (approved):
Looks awesome! Thank you!! 🚀

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,160 @@
...
+| 3 | 9 |
+| 4 | 1 |
+
+MetricFlow then filters this table to accounts with more than 5 data model runs and counts the number of accounts that meet this criteria:
+
+| activated_accounts |
+| --- |
+| 2 |
+
+## Considerations when using metrics in filter
```
mirnawong1:
```suggestion
## Considerations
```

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,166 @@
+---
+title: "Metrics as dimensions with metric filters"
+id: "ref-metrics-in-filters"
+description: "Add metrics as dimensions to your metric filters to create more complex metrics and gain more insights."
+sidebar_label: "Metrics as dimensions"
+---
+
+[Metrics](/docs/build/metrics-overview) provider users with valuable insights into their data, like number of active users and overall performance trends to inform business decisions. [Dimensions](/docs/build/dimensions), on the other hand, help categorize data through attributes, like user type or number of orders placed by a customer.
```
mirnawong1:
```suggestion
[Metrics](/docs/build/metrics-overview) provide users with valuable insights into their data, like number of active users and overall performance trends to inform business decisions. [Dimensions](/docs/build/dimensions), on the other hand, help categorize data through attributes, like user type or number of orders placed by a customer.
```

website/docs/docs/build/ref-metrics-in-filters.md
```diff
@@ -0,0 +1,166 @@
...
+          sum(1) as data_model_runs, 
+		      account
+	    from data_model_runs
+	    group by 
+		      account
+  ) as subq on accounts.account = subq.account
+  where data_model_runs > 5
+	```
+
+  The intermediate tables used to create this metric are: Accounts with the `data_model_runs` dimension
```
mirnawong1:
```suggestion
  The intermediate tables used to create this metric is: Accounts with the `data_model_runs` dimension
```

PR #5325, opened 2024-04-23
update sl guide per feedback

mirnawong1:
This PR adds/updates the example yaml configs per @faith's feedback [in slack](https://dbt-labs.slack.com/archives/C06R0B7UCBW/p1713821380963739)


website/docs/guides/sl-snowflake-qs.md
```diff
@@ -818,7 +818,7 @@ After setting up your orders model:
 
 ```yaml
 semantic_models:
-  - name: customers
+  - name: dim_customers
```
faithebear: @mirnawong1 can't add a comment on the right line for some reason! But in line 826 of this file the model: key should say model: ref('dim_customers') and it currently says ref('customers') I think line 821 can stay as customers if we like, it's just that it has to ref dim_customers.

faithebear (changes requested):
left a comment on line 826--model should ref dim_customers not customers. otherwise this looks great, thanks :)

website/docs/guides/sl-snowflake-qs.md
mirnawong1:
```suggestion
  - name: customers
```

PR #5261, opened 2024-04-10
[Feature] Add 'Configuring unique connections' blog post

gwenwindflower:
## What are you changing in this pull request and why?

This PR adds a blog post covering Extended Attributes and some other related features for configuring more unique Environment and User connections in dbt Cloud.

PR #5328, opened 2024-04-23
[Fix]  Remove root package-lock.json and update correct one

gwenwindflower:
This is a hotfix to remove the mistakenly add package-lock.json at the root of the project and update the correct one in the `website/package-lock.json` path just to be sure everything is in sync correctly.


dbeatty10 (approved):
This looks like the right move to undo the `package-lock.json`-specific pieces of https://github.com/dbt-labs/docs.getdbt.com/pull/5261.

PR #5349, opened 2024-04-24
update label parameter

mirnawong1:
this pr updates the following per feedback from Amy C:

- `label` field isn’t required for measures 
- however `label` field IS required for metrics config 

this pr adds clarifying detail in all the metrics docs (simple, ration, derived, etc), and removes it from semantic model docs. 

[slack thread](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1713885759185999)


runleonarun (approved):
Looks good and can be shipped

PR #5438, opened 2024-05-02
Update SL Streamlit app blog

joellabes:
The article talks about "why build this" and "what this thing does" before it actually describes what Doug's built! This moves the Loom demo video to the top of the article and explains briefly what it does.


amychen1776 (approved):
Ahh this makes sense now.

PR #5503, opened 2024-05-14
Make clear that KoLV includes the current core version

joellabes:
We got some feedback that Keep on Latest Version in dbt Cloud doesn't include 1.8. It actually does - we need to make that clearer!


jtcohen6:
Thank you for this @joellabes!

TODO: also document that in dbt Cloud APIs / Terraform the value is `'versionless'` (as opposed to e.g. `'1.7.0-latest'`)

mirnawong1:
Am linking to [slack thread](https://dbt-labs.slack.com/archives/C05FWBP9X1U/p1715782793330549) which has some additional context

matthewshaver (approved):

website/docs/docs/dbt-versions/upgrade-dbt-version-in-cloud.md
```diff
@@ -13,7 +13,7 @@ Navigate to the settings page of an environment, then click **Edit**. Click the
 
 ### Keep on latest version
 
-By choosing to **Keep on latest version**, you always get the latest fixes and early access to new functionality for your dbt project. dbt Labs will handle upgrades for you, as part of testing and redeploying the dbt Cloud SaaS application.
+By choosing to **Keep on latest version**, you always get the latest fixes and early access to new functionality for your dbt project. dbt Labs will handle upgrades for you, as part of testing and redeploying the dbt Cloud SaaS application. Keep on latest version always includes the most recent version of dbt Core, and more.
```
mirnawong1, line 16: do we think this statement might benefit from an example? so: Keep on latest version always includes the most recent version of dbt Core, and more. For example, if the latest version is 1.8, then selecting "Keep on latest version" mean your environment is automatically updated to dbt v1.8.

PR #5508, opened 2024-05-14
Redshift unit test limitation

### LINKED ISSUE #5430: [Core] Compute-node only functions in Redshift like MEDIAN, etc. won't work with unit tests
### Link to the page(s) on docs.getdbt.com requiring updates

Wherever we are documenting platform-specific caveats.

Possibly this page:
https://docs.getdbt.com/docs/build/unit-tests

Possibly this section:
https://docs.getdbt.com/docs/build/unit-tests#before-you-begin

### Tell us more about this update

As noted in https://github.com/dbt-labs/dbt-redshift/issues/807, certain functions in Redshift (`MEDIAN`, `LISTAGG`, etc.) don't currently work with dbt's built-in unit testing capabilities being introduced in v1.8.

The reason is that functions like MEDIAN are compute-node only functions and return an error if the query doesn't reference a user-defined table or Amazon Redshift system table.

Since dbt unit testing relies on CTE (rather than tables), models that contain any of these functions will fail with an error like:
```
[XX000] ERROR: One or more of the used functions must be applied on at least one user created tables. Examples of user table only functions are LISTAGG, MEDIAN, PERCENTILE_CONT, etc
```

### Reviewers/Stakeholders/SMEs

.

### Related GitHub issues

https://github.com/dbt-labs/dbt-redshift/issues/807

### Additional information

See internal Slack thread here:
https://dbt-labs.slack.com/archives/C05LRDBSCNL/p1714653471861229

matthewshaver:

Closes: https://github.com/dbt-labs/docs.getdbt.com/issues/5430

## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) so my content adheres to these guidelines.
- [ ] For [docs versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#about-versioning), review how to [version a whole page](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) and [version a block of content](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#versioning-blocks-of-content).

Adding or removing pages (delete if not applicable):
- [ ] Add/remove page in `website/sidebars.js`
- [ ] Provide a unique filename for new pages
- [ ] Add an entry for deleted pages in `website/vercel.json`
- [ ] Run link testing locally with `npm run build` to update the links that point to deleted pages


website/docs/docs/build/unit-tests.md
```diff
@@ -26,8 +26,8 @@ With dbt Core v1.8 and dbt Cloud environments that opt to "Keep on latest versio
 - You must specify all fields in a BigQuery STRUCT in a unit test. You cannot use only a subset of fields in a STRUCT.
 - If your model has multiple versions, by default the unit test will run on *all* versions of your model. Read [unit testing versioned models](#unit-testing-versioned-models) for more information.
 - Unit tests must be defined in a YML file in your `models/` directory.
-- Available to dbt Cloud customers who have selected ["Keep on latest version"](/docs/dbt-versions/upgrade-dbt-version-in-cloud#keep-on-latest-version) and  dbt Core v1.8.0 or later.
```
runleonarun, line 29: Are you removing this because it repeats info in the note at the top of the page?

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,88 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
 </VersionBlock>
 
 </VersionBlock>
+
+<VersionBlock firstVersion="1.8">
+
+## Unit test limitations
+
+Unit tests aren't supported on Redshift if the SQL in the common table expression (CTE) contains functions such as `LISTAGG`, `MEDIAN`, `PERCENTILE_CONT`, etc. Those functions must be executed against a user-created table. dbt combines given rows to be part of the CTE, which is unsupported by Redshift. You can try the following SQL:
```
runleonarun:
```suggestion
Redshift doesn't support Unit tests when the SQL in the common table expression (CTE) contains functions such as `LISTAGG`, `MEDIAN`, `PERCENTILE_CONT`, etc. These functions must be executed against a user-created table. dbt combines given rows to be part of the CTE, which Redshift does not support. 

The following query illustrates this limitation:
```

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,88 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
...
+         LISTAGG(name || ' x ' || quantity, ',') AS option_name_list,
+         id
+      FROM test_fixture
+      GROUP BY id
+   )
+   select * from agg
+);
+
+```
+This results in the error:
```
runleonarun:
```suggestion
The previous code results in the error: 
```

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,88 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
...
+      LISTAGG(name || ' x ' || quantity, ',') AS option_name_list,
+      id
+   FROM test_tmp1234
+   GROUP BY id
+)
+select * from agg;
+
+```
+
+If all given rows are created as a temporary table first, then running the test by referring to it will result in a successful run.
```
runleonarun:
```suggestion
When all given rows are created as a temporary table first, then running the test by referring to the temporary tables results in a successful run.
```

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,88 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
...
+
+In short, separate the unit tests into two steps:
```
runleonarun, line 341: I feel like these should be moved up to the top of this section so we start with how to do this right.

runleonarun (changes requested):
@matthewshaver Just have some suggestion to smooth this message out.

website/docs/docs/build/unit-tests.md
matthewshaver: Yeah, this shouldn't have been there. There was previously the exact same note at the top of this list, so apparently it was there twice. We removed it in favor of the banner because folks were missing it

website/docs/reference/resource-configs/redshift-configs.md
matthewshaver: Could going from "The following query" to the code example, to "the previous code" be a little redundant?

website/docs/reference/resource-configs/redshift-configs.md
matthewshaver: We do cover these points at the top. Does this work well as a summary, or should we switch the top and the bottom? I think it's beneficial to have it in both places in some fashion

website/docs/reference/resource-configs/redshift-configs.md
runleonarun: @matthewshaver  how about "This query results in the error:"

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,90 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
 </VersionBlock>
 
 </VersionBlock>
+
+<VersionBlock firstVersion="1.8">
+
+## Unit test limitations
+
+Redshift doesn't support Unit tests when the SQL in the common table expression (CTE) contains functions such as `LISTAGG`, `MEDIAN`, `PERCENTILE_CONT`, etc. These functions must be executed against a user-created table. dbt combines given rows to be part of the CTE, which Redshift does not support.
```
runleonarun: @matthewshaver What I like about how the content is [framed below](https://github.com/dbt-labs/docs.getdbt.com/pull/5508/files#diff-5d2cee8c48606c33956458e7b89eacbda29c62349ad49d68ea9380173acda697R339-R343) is that it's clearer than this paragraph and it's framed in the positive. I think we should rewrite this paragraph so we: 
1. Communicate that unit test are limited by actually saying what they need to do to get unblocked.  like "Avoid using SQL in CTE functions such as.." and "Unit tests can be used when you.."
2. Explain how to overcome this limitation. Instead of the passive language of "These functions must be executed..." Use more active language like "Execute these functions against..."

website/docs/reference/resource-configs/redshift-configs.md
runleonarun: I see your point about keeping the summary so instead I [made some suggestions](https://github.com/dbt-labs/docs.getdbt.com/pull/5508/files#r1601956520) to clear up the content at the top.

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,90 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
...
+         LISTAGG(name || ' x ' || quantity, ',') AS option_name_list,
+         id
+      FROM test_fixture
+      GROUP BY id
+   )
+   select * from agg
+);
+
+```
+This results in the error:
```
matthewshaver:
```suggestion
This query results in the error: 
```

website/docs/reference/resource-configs/redshift-configs.md
matthewshaver: Works for me!

website/docs/reference/resource-configs/redshift-configs.md
matthewshaver:
```suggestion
Redshift doesn't support Unit tests when the SQL in the common table expression (CTE) contains functions such as `LISTAGG`, `MEDIAN`, `PERCENTILE_CONT`, etc. These functions must be executed against a user-created table. dbt combines given rows to be part of the CTE, which Redshift does not support. For unit tests to function properly in this scenario, creating temporary tables for the unit tests to reference is a good workaround. 
```

website/docs/reference/resource-configs/redshift-configs.md
```diff
@@ -257,3 +257,90 @@ The workaround is to execute `DROP MATERIALIZED VIEW my_mv CASCADE` on the data
...
 
 </VersionBlock>
+
+<VersionBlock firstVersion="1.8">
+
+## Unit test limitations
+
+Redshift doesn't support Unit tests when the SQL in the common table expression (CTE) contains functions such as `LISTAGG`, `MEDIAN`, `PERCENTILE_CONT`, etc. These functions must be executed against a user-created table. dbt combines given rows to be part of the CTE, which Redshift does not support. For unit tests to function properly in this scenario, creating temporary tables for the unit tests to reference is a good workaround. 
+
+The following query illustrates this limitation:
```
matthewshaver:
```suggestion
The following query illustrates the limitation:
```

runleonarun (approved):
This works

PR #5530, opened 2024-05-20
clarify where docs serve is supported to ensure

mirnawong1:
per [slack thread](https://dbt-labs.slack.com/archives/C06FWKP5V28/p1715952639527989), this pr sets to clarify where you can use the docs serve command as users previously weren't aware. this pr tries to make it more explicit.


jp-dbt (approved):

mirnawong1:
thanks soo much @jp-dbt - really appreciate! 🙏

PR #5550, opened 2024-05-22
Add `is_incremental()` as another case when you will need to force dependencies

### LINKED ISSUE #5547: add is_incremental as additional example of when you will need to force dependencies
### Contributions

- [ ] I have read the contribution docs, and understand what's expected of me.

https://docs.getdbt.com/reference/dbt-jinja-functions/ref#forcing-dependencies

We already have an example for when you need to force dependencies - a model that only has a macro.

Let's also add an example for when references are wrapped in the `is_incremental()` macro - see [slack thread](https://dbt-labs.slack.com/archives/C05FWBP9X1U/p1716406069131559?thread_ts=1716344292.510149&cid=C05FWBP9X1U). 

Dependencies nested in the `is_incremental` macro will not show up in lineage, so will need to force dependencies in that case.

### Example

If you have an incremental model that includes a snippet like below, then you will not be able to see the `source('raw', 'orders')` upstream dependency in the lineage.

```sql
{% if is_incremental() %}
select * from {{ source('raw', 'orders') }}
{% endif %}
```

This is because the `is_incremental` macro will always return false at parse time, so anything buried inside that will not be available to parsing, so will not be added to `depends_on`.

So in this case, you'd need to do [force a dependency](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#forcing-dependencies) _outside_ of the `is_incremental()` conditional like this instead:

```sql
 -- depends_on: {{ source('raw', 'orders') }}

{% if is_incremental() %}
select * from {{ source('raw', 'orders') }}
{% endif %}
```

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/reference/dbt-jinja-functions/ref#forcing-dependencies)

Resolves https://github.com/dbt-labs/docs.getdbt.com/issues/5547

## Checklist

### DIFF
--- a/website/docs/reference/dbt-jinja-functions/ref.md
+++ b/website/docs/reference/dbt-jinja-functions/ref.md
@@ -102,3 +102,17 @@ In normal usage, dbt knows the proper order to run all models based on the usage
 ```
 
 dbt will see the `ref` and build this model after the specified reference.
+
+Another example is when a reference appears within an [`is_incremental`](/docs/build/incremental-models#understand-the-is_incremental-macro) conditional block.
+
+This is because the `is_incremental` macro will always return `false` at parse time, so any references within it can't be inferred.
+
+To handle this, you can use a SQL comment outside of the `is_incremental()` conditional:
+
+```sql
+-- depends_on: {{ source('raw', 'orders') }}
+
+{% if is_incremental() %}
+select * from {{ source('raw', 'orders') }}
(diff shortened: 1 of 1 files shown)


graciegoheen (approved):
MVP Doug !

runleonarun (approved):
:shipit:

PR #5556, opened 2024-05-23
Revert "updates the expandable component"

mirnawong1:
the expandables aren't appearing correctly on the live site so this reverts this pr

https://docs.getdbt.com/docs/collaborate/dbt-explorer-faqs

Reverts dbt-labs/docs.getdbt.com#5531
Screenshot 2024-05-23 at 15 40 45

PR #5531, opened 2024-05-20
updates the expandable component

mirnawong1:
this pr iterates on the expandable component to improve the user experience based on some feedback [slack thread](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1714405513625989):

- separated the expandable component from the headerLinkCopy.js file as i think that was conflicting with the behavior.
- allows copying and text highlighting inside the expandable
- added separate code for a copy link pop up/close behavior. `showCopyPopup` and `closeButton` with timeout.
- added css info so expandable alt_header text is black when using details theme. 
- also adds a `useEffect` behavior so that the toggle is expanded when the user goes to the copied link so they know where to start reading

[asana task](https://app.asana.com/0/1200099998847559/1207358280519061/f) requesting PR review

PR #5561, opened 2024-05-23
update expandables component v2

mirnawong1:
this pr is a second attempt to update the expandable component. the first pr (https://github.com/dbt-labs/docs.getdbt.com/pull/5531) was [reverted](https://github.com/dbt-labs/docs.getdbt.com/pull/5556) quickly and now this PR correctly updated the component to behavior as described below AND replaces the all the components the use`<expandable>` to `<Expandable>` to abide by react's best practices.

### description
iterates on the expandable component to improve the user experience based on some feedback slack thread:

separated the expandable component from the headerLinkCopy.js file as i think that was conflicting with the behavior.
allows copying and text highlighting inside the expandable
added separate code for a copy link pop up/close behavior. showCopyPopup and closeButton with timeout.
added css info so expandable alt_header text is black when using details theme.
also adds a useEffect behavior so that the toggle is expanded when the user goes to the copied link so they know where to start reading
asana task requesting PR rev

### DIFF
--- a/website/docs/docs/build/custom-schemas.md
+++ b/website/docs/docs/build/custom-schemas.md
@@ -98,7 +98,7 @@ To customize this macro, copy the example code in the section [How does dbt gene
 
 Be careful. dbt will ignore any custom `generate_schema_name` macros included in installed packages.
 
-<expandable alt_header="Warning: Don't replace `default_schema` in the macro">
+<Expandable alt_header="Warning: Don't replace `default_schema` in the macro">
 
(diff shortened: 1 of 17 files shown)


john-rock (approved):
Hey @mirnawong1 looks like we are all good now! Confirming the expandable is working as expected after updated the case where the component is being used.

image

mirnawong1:
life is good @john-rock ! let's go and thank you so much!

PR #5585, opened 2024-05-31
update unit test versioned models link

mirnawong1:
this pr updates the link to the correct unit test versioned models link. previously it was linking to a non existent anchor link: `Read [unit testing versioned models](#unit-testing-versioned-models) for more information)` and not going anywhere. 

this pr sets to add a redirect as well so the dbt learn course links to the correct place and we avoid user confusion. 

raised in slack by @b-per: https://dbt-labs.slack.com/archives/C02NCQ9483C/p1717152369804939


nataliefiann:
Hiya @mirnawong1 this looks really good. I've tested the redirect links and they didn't redirect to the page as it was an anchor link but the training and front end teams have been tagged to make them aware.

nataliefiann (approved):

PR #5586, opened 2024-05-31
Update exports.md

mirnawong1:
remove multi tenant as exports supports both mt and st.


DevonFulcher (approved):
Awesome thank you!

mirnawong1:
amazing thank you so much @DevonFulcher !

mirnawong1:
@DevonFulcher this might be a silly question - can i just confirm that exports also supports multi cell?

DevonFulcher:
@mirnawong1 Good question! As far as I know, there is nothing preventing users from using Exports in MC

PR #5590, opened 2024-06-02
Update connect-snowflake.md

martydertz:
link to key-pair main docs


mirnawong1 (approved):
hey @martydertz , great catch! thank you for opening this pr up to update the link. it should take users to the right location now. ✨

PR #5599, opened 2024-06-04
Update resource-type.md

mirnawong1:
add a note that the exclude-resource-type flag is for 1.8 and higher


nataliefiann (approved):
Hiya @mirnawong1 

This looks really great! I've had the look at the page on vercel and they layout is perfect and reads really well. I have no suggestions so I'll approve this for you. Please let me know if there's anything else I can do to help.

Kind Regards
Natalie

PR #5621, opened 2024-06-06
Update metrics-overview.md

mirnawong1:
add clarification that metric filter is available in 1.8+ or versionless per user slack feedback: https://getdbt.slack.com/archives/C046L0VTVR6/p1716496151560399


Jstein77 (approved):
LGTM!

PR #5707, opened 2024-06-27
Add docs for using `psycopg2` with `dbt-postgres`

### LINKED ISSUE #96: [Regression] Forced psycopg2 dependency on Linux causes sudden build failures
### Is this a regression?

- [X] I believe this is a regression in functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

My existing build system now breaks due to introduction of hard `psycopg2` dependency into pyproject.toml in 1.8.0 in #60 , which forces `psycopg2` compilation from source on Linux, which hasn't got the dependencies installed in my docker image, which breaks the build. My desired production setup is psycopg2-binary (for consistency), and I do not wish to build psycopg2 from source (because I don't want the dev dependencies).

### Expected/Previous Behavior

I expect a minor release to not break an existing build system, and I expect the same psycopg2 variant to be used across dev and prod for quality reasons (replicable build).

### Steps To Reproduce

1. On an existing psycopg2-binary setup, update dbt-postgres to 1.8.0.

### Relevant log output

```shell
#0 21.05   ChefBuildError
#0 21.05 
#0 21.05   Backend subprocess exited when trying to invoke get_requires_for_build_wheel
#0 21.05   
#0 21.05   running egg_info
#0 21.05   writing psycopg2.egg-info/PKG-INFO
#0 21.05   writing dependency_links to psycopg2.egg-info/dependency_links.txt
#0 21.05   writing top-level names to psycopg2.egg-info/top_level.txt
#0 21.05   
#0 21.05   Error: pg_config executable not found.
#0 21.05   
#0 21.05   pg_config is required to build psycopg2 from source.  Please add the directory
#0 21.05   containing pg_config to the $PATH or specify the full executable path with the
#0 21.05   option:
#0 21.05   
#0 21.05       python setup.py build_ext --pg-config /path/to/pg_config build ...
#0 21.05   
#0 21.05   or with the pg_config option in 'setup.cfg'.
#0 21.05   
#0 21.05   If you prefer to avoid building psycopg2 from source, please install the PyPI
#0 21.05   'psycopg2-binary' package instead.
#0 21.05   
#0 21.05   For further information please check the 'doc/src/install.rst' file (also at
#0 21.05   <https://www.psycopg.org/docs/install.html>).
```

### Environment

```markdown
- OS: Mac
- Python: 3.9
- dbt-postgres (working version): 1.7.14
- dbt-postgres (regression version): 1.8.0
```

### Additional Context

There is also a [comment here](https://github.com/dbt-labs/dbt-postgres/commit/32c8547ae42169f8147c5fbec2712c22b7f9db33#r142012438) from @andy-clapson challenging Linux=prod assumption.

mikealfare:

We changed how `dbt-postgres` is installed with regards to `psycopg2` versus `psycopg2-binary`.
In prior versions you could configure it with an environment variable. In 1.8, that's not an option because we changed our build tooling to a more modern stack. We initially tried an OS-driven approach, but that had issues mainly due to CI workflows that normally run on linux images. So in 1.8.2 we reverted back to installing `psycopg2-binary` by default and prescribed the manual workaround of uninstalling `psycopg2-binary` and reinstalling `psycopg2`.

## Checklist
- [x] For [docs versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#about-versioning), review how to [version a whole page](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) and [version a block of content](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#versioning-blocks-of-content).


website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+```
+You should research which OS dependencies are needed for your particular scenario.
+
+<VersionBlock firstVersion="1.8">
+
+#### Limitations
+
+In versions 1.8.0 and 1.8.1, `psycopg2-binary` is installed on MacOS and Windows operating systems and `psycopg2` is installed on Linux operating systems.
+This has the side effect of requiring the OS dependencies identified above to install `dbt-postgres` on Linux.
+Users will either need to update their workflows to install these dependencies, or upgrade to 1.8.2.
```
jtcohen6, line 152: 👍 Given that we recommend users always install the latest (patch) version, no one _should_ be installing v1.8.0 or v1.8.1 (instead of v1.8.2+) — but given you've already written it, I'm happy to have it in here :)

jtcohen6 (approved):
@mikealfare Thank you for this - looks great!

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+`psycopg2-binary` is installed by default when installing `dbt-postgres`.
+This is ideal for development and testing workflows where performance is less of a concern.
+However, production environments will benefit from `psycopg2`, which is built from source for that particular operating system and archtecture.
+
+<VersionBlock firstVersion="1.8">
+
+To use `psycopg2`, simply uninstall `psycopg2-binary` and install the equivalent version of `psycopg2` after installing `dbt-postgres`:
+```bash
+pip install dbt-postgres
+if [[ $(pip show psycopg2-binary) ]]; then
```
VersusFacit, line 113: Nit: `if pip show psycopg2-binary > /dev/null 2>&1; then` is my preferred way to achieving this behavior (and also suppressing unnecessary error information which `$()` does not suppress.

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+    PSYCOPG2_VERSION=$(pip show psycopg2-binary | grep Version | cut -d " " -f 2)
+    pip uninstall -y psycopg2-binary
```
VersusFacit: Maybe: 

`pip uninstall -y psycopg2-binary &&  pip install psycopg2==$PSYCOPG2_VERSION` would be nice because then we chain those commands together rather than trying to install on an unsuccessful uninstall.

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
 
 If `dbt-postgres` encounters an operational error or timeout when opening a new connection, it will retry up to the number of times configured by `retries`. The default value is 1 retry. If set to 2+ retries, dbt will wait 1 second before retrying. If set to 0, dbt will not retry at all.
 
+
+### `psycopg2` vs `psycopg2-binary`
+
+`psycopg2-binary` is installed by default when installing `dbt-postgres`.
+This is ideal for development and testing workflows where performance is less of a concern.
```
VersusFacit: How is using a binary less efficient? A few words would help color this for non-SMEs

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+However, production environments will benefit from `psycopg2`, which is built from source for that particular operating system and archtecture.
+
+<VersionBlock firstVersion="1.8">
+
+To use `psycopg2`, simply uninstall `psycopg2-binary` and install the equivalent version of `psycopg2` after installing `dbt-postgres`:
```
VersusFacit:
```suggestion
To use `psycopg2`:
1. install `dbt-postgres`
2. uninstall `psycopg2-binary`
3. install the equivalent version of `psycopg2`
```

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+    pip uninstall -y psycopg2-binary
+    pip install psycopg2==$PSYCOPG2_VERSION
+fi
+```
+
+</VersionBlock>
+
+<VersionBlock lastVersion="1.7">
+
+To use `psycopg2`, ensure that `DBT_PSYCOPG2_NAME=psycopg2` when installing `dbt-postgres`.
```
VersusFacit: Maybe this is worse?
```suggestion
To ensure your dbt installation uses `psycopg2`, prefix all `dbt-postgres` installation commands with `DBT_PSYCOPG2_NAME=psycopg2`.
```

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+The easiest way to do this is to set it inline while installing:
```
VersusFacit: if you take my above suggestion
```suggestion
For example:
```

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+```bash
+DBT_PSYCOPG2_NAME=psycopg2 pip install dbt-postgres
+```
+
+</VersionBlock>
+
+Installing `psycopg2` requires OS level dependencies, which is one of the reasons users opt for `psycopg2-binary` (the other reason being a much longer installation time).
```
VersusFacit: I like to group reasons together rather than using parenthetical. 
```suggestion
Some systems prefer `psycopg2-binary` since use of `psycopg2` requires OS level dependencies and a much longer installation time.
```

website/docs/docs/core/connect-data-platform/postgres-setup.md
```diff
@@ -93,3 +93,57 @@ If the database closes its connection while dbt is waiting for data, you may see
...
+```bash
+sudo apt-get update
+sudo apt-get install libpq-dev python-dev
+```
+whereas on Mac, you need to install `postgresql`:
+```bash
+brew install postgresql
+pip install psycopg2
+```
+You should research which OS dependencies are needed for your particular scenario.
```
VersusFacit: Can we provide any idea of what this research should be? Maybe it might be better just to say any OS may have it's own dependencies to discover.

VersusFacit (approved):
Little nits and thoughts for final revision.

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: I added a little more verbiage. Let me know what you think.

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: I don't disagree with the approach, but I'd rather have something that's easier for end users to read and understand. I don't think this is worth the trade off in this case.

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: That's fair.

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: Reworded to remove "research".

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: Agreed on the recommendation. I'm trying to be cognizant of the fact that `dbt-postgres` tends to be used more as a base adapter than other adapters (e.g. what we used to do with `dbt-redshift`). I thought it might be more likely in those scenarios that folks are hard pinning the version.

website/docs/docs/core/connect-data-platform/postgres-setup.md
mikealfare: I wouldn't say that the system prefers it, rather the user may prefer it on some systems. Though I agree this is a bit wordy; I'll try rewording it a bit.

mikealfare (commented):
@VersusFacit I incorporated most of your feedback. Let me know what you think of the updated version.

VersusFacit (approved):

colin-rogers-dbt:
@matthewshaver is there a blocker to merging this?

mirnawong1:
Happy to merge this for you tomorrow @mikealfare if there are no issues. This isn't tied to v1.9 right?

mikealfare:
> Happy to merge this for you tomorrow @mikealfare if there are no issues. This isn't tied to v1.9 right?

I'm not sure what is meant by "tied to v1.9". This takes effect in `1.8.2`, and then would continue into `1.9.0` and beyond.

mirnawong1:
great thanks @mikealfare, was just making sure this wasn't an upcoming 1.9 feature but you're right - it's very clear this is live already.  ! will merge this pr for yoU!

PR #5712, opened 2024-06-28
Updates query param handling on /guides

john-rock:
[Asana Task](https://app.asana.com/0/1202136277772113/1207686648073613/f)
This PR allows additional query params to be used on the `/guides` page. Additional query params are required for activating Mutiny preview mode. Currently we can not activate Mutiny previews on this route.

[Slack thread](https://dbt-labs.slack.com/archives/C06M99V657T/p1718991855866109) with Mutiny support for more context.

### Old Behavior
Only `tags` and `level` were allowed as query params. Any additional query param would be removed.
Access https://docs.getdbt.com/guides?mutiny=1 and see the `mutiny` param get automatically removed on page load.

### New Behavior
Additional query params can be used while keeping the `tags` and `level` params only available to be filtered on.

## Testing Steps
1. Access https://docs-getdbt-com-git-fix-params-on-guides-dbt-labs.vercel.app/guides?mutiny=1 and confirm `mutiny` param does not get removed.
2. Confirm filtering by URL param still works https://docs-getdbt-com-git-fix-params-on-guides-dbt-labs.vercel.app/guides?mutiny=1&tags=Best+practices and that the `mutiny` param does not get applied as a filter.

### DIFF
--- a/website/src/components/quickstartGuideList/index.js
+++ b/website/src/components/quickstartGuideList/index.js
@@ -48,11 +48,20 @@ function QuickstartList({ quickstartData }) {
   }, [quickstartData]);
 
   const updateUrlParams = (selectedTags, selectedLevel) => {
-    const params = new URLSearchParams();
+    // Create a new URLSearchParams object from the current URL search string
+    const params = new URLSearchParams(location.search);
+
+    // Remove existing 'tags' and 'level' parameters to avoid duplicates
+    params.delete('tags');
+    params.delete('level');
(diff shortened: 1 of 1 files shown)


JKarlavige (approved):
lgtm! verified i'm able to:
- load the page with `mutiny` param without it being deleted
- able to add `level` & `tags` params without `mutiny` param being deleted
- able to update levels & tags without other params being affected

PR #5720, opened 2024-07-02
fix(redirects): update redirects for sl guide update

gwenwindflower:
Quick fix for missing redirects! Thank you @T-Dunlap and @mirnawong1 for finding and getting the missing pages right on the first try respectively ❤️ !


mirnawong1 (approved):
approved and will test with the preview @gwenwindflower ! to be fair, i think you had one for the marts doc - i just added the forward slash at the beginning

mirnawong1:
redirects work in previews!

PR #5723, opened 2024-07-03
fix(metrics overview): clean up typo in example

gwenwindflower:
This example looks like it contains invalid YAML, I believe the correct YAML is this. Will confirm tomorrow, don't merge this until confirmed please, thank you!


mirnawong1 (approved):
thank you for opening this up, winnie!!! confirmed with winnie how it should be and merging this ✨

PR #5755, opened 2024-07-09
Update mesh-qs.md

mirnawong1:
removed dbt_utils from the sql as it's no longer needed and returns an error. raised by support @jennlw


mirnawong1:
confirmed that dbt_utils isn't needed: https://dbt-labs.slack.com/archives/C017GDLAF7D/p1720532313714379

PR #5306, opened 2024-04-18
Update incremental-models.md

zvijayakumar:
If event_time is NULL or the table is truncated, the condition will always be true and load all records.  This is an edge case. 

If a final table is truncated it will return the NULL as the result 

( select * from table where event_time > NULL ) --> return no results 

It will not have any data in it and dbt considers it as incremental load. then the model will fail, to avoid that situation. In dbt we don't have anything to handle this situation. 

## Checklist

- [ ] Review the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) so my content adheres to these guidelines.
- [ ] For [docs versioning](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#about-versioning), review how to [version a whole page](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#adding-a-new-version) and [version a block of content](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/single-sourcing-content.md#versioning-blocks-of-content).
- [ ] Add a checklist item for anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

Adding or removing pages (delete if not applicable):
- [ ] Add/remove page in `website/sidebars.js`
- [ ] Provide a unique filename for new pages
- [ ] Add an entry for deleted pages in `website/vercel.json`
- [ ] Run link testing locally with `npm run build` to update the links that point to deleted pages

PR #5785, opened 2024-07-15
Restoring cross-database compatibility of example code

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/docs/build/incremental-models#filtering-rows-on-an-incremental-run)

The code change in https://github.com/dbt-labs/docs.getdbt.com/pull/5326 was tested against a variety of dbt adapters. The update in https://github.com/dbt-labs/docs.getdbt.com/pull/5306 modified that code example, and it appears to have database-specific cast syntax (`::`) as well as a data type that means different things in different databases (i.e., the `TIMESTAMP` data type has different semantics in BigQuery vs. Snowflake).

So we should restore the code example in https://github.com/dbt-labs/docs.getdbt.com/pull/5326. If there is a better example that has been tested across dbt adapters, then we can consider making another update in the future.

## Checklist


mirnawong1 (approved):
thanks for clarifying that @dbeatty10 ! apologies for that!

PR #5816, opened 2024-07-20
Fix rendering of snippet for run results JSON file

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-2-1-dbt-labs.vercel.app/reference/artifacts/run-results-json)

It currently looks like [this](https://docs.getdbt.com/reference/artifacts/run-results-json):

image

When it should look more like [this](https://docs.getdbt.com/reference/dbt-classes):

image

> [!NOTE]  
> There is an extra newline between bullet points due to the snippet, but I wasn't able to make it go away after a [couple attempts](https://github.com/dbt-labs/docs.getdbt.com/pull/5816/commits).
>
> If we want to eliminate the extra newline, we'd either need to go away from using a snippet or figure out how to make the snippet place nice.

## Checklist


mirnawong1 (approved):
oh wow, great catch and thanks for always improving our docs @dbeatty10 ! merging this for you!

PR #5835, opened 2024-07-24
add clarifying faq and info on permissions

mirnawong1:
this pr adds a couple of sentences clarifying users need the right permission/access to create schemas when running store_failures config. 

raised in slack thread by @wusanny : https://dbt-labs.slack.com/archives/C02NCQ9483C/p1721798716164609


mirnawong1:
hey @wusanny , can i double check with you if the context is technically accurate please?

website/docs/reference/resource-configs/store_failures.md
```diff
@@ -110,3 +108,20 @@ tests:
...
 </Tabs>
+
+## FAQs
+
+<detailsToggle alt_header="Receiving a 'permissions denied for schema' error">
+
+If you're receiving a `Adapter name adapter: Adapter_name error: permission denied for schema dev_usernamea_dbt_test__audit`, this is most likely due to your user not having permission to create new schemas, despite having owner access to your own development schema.
+
+To resolve this, you need proper authorization to create or access custom schemas. Run the following SQL command in the data platform environment:
+
```
nataliefiann, line 119: Hiya @mirnawong1 

(non-blocker suggestion) would it be worthwhile adding an image to show users where they can add these credentials?

Kind Regards
Natalie

nataliefiann (approved):
Hiya @mirnawong1 

I love this! I've left a non-blocker suggestion but other than that, I have no real suggestions or recommendations. I've approved this for you.

Kind Regards
Natalie

website/docs/reference/resource-configs/store_failures.md
mirnawong1: thats a great question and i did think about this but i think the user has to do this in their own data platform settings. I'm going to loop @wusanny to confirm that though

wusanny: Thank you @mirnawong1 for writing these up. The authorization needs to be done in their respective data platform, that is right. 

Hopefully am responding to these correctly, just couple things:
- tiny typo on "dev_usernamea_dbt_test__audit" in line 116 - there is an extra 'a'
- the exact authorization query may differ from one data platform to another, not sure if it's something worth mentioning?

mirnawong1: brilliant, thanks so much @wusanny ! will update and thanks for those catches! 🦅👀

mirnawong1: folded in those changes, thank you again!

PR #5856, opened 2024-07-29
Example of using the "config" selection method for `meta`

dbeatty10:
[Preview](https://docs-getdbt-com-git-dbeatty10-patch-1-dbt-labs.vercel.app/reference/node-selection/methods#the-config-method)

The ["config" selector method](https://docs.getdbt.com/reference/node-selection/methods#the-config-method) can be used to select models that match a specified [node config](https://docs.getdbt.com/reference/configs-and-properties).

There's a [`meta` config](https://docs.getdbt.com/reference/resource-configs/meta) for the following resource types, which can be selected via the "config" selector method:
- Models
- Sources
- Seeds
- Snapshots
- Singular Data Tests
- Semantic Models
- Metrics
- Saved queries

The "config" selector method doesn't currently work `meta` within the following resource types currently though (either because `meta` isn't supported at all or because it is not underneath `config`):
- doesn't support `meta` at all:
  - Generic Data Tests
  - Unit Tests
  - Analyses
- `meta` not underneath `config`
  - Macros
  - Exposures

## Checklist


mirnawong1 (approved):
looks great, thank you @dbeatty10 !

PR #5950, opened 2024-08-21
Mention that Account Viewer also has access to Audit Logs

b-per:
Account Viewer also has access to audit logs since recently

Updating page to mention that Audit logs are accessible by people with Account Viewer access as well


mirnawong1 (approved):
thanks for opening this up @b-per ! merging this for you now!

PR #5957, opened 2024-08-22
Fix whitespace and callout

nghi-ly:

The blockquote in the callout in this section https://docs.getdbt.com/docs/cloud/secure/databricks-privatelink#configure-aws-privatelink is not showing up in light mode.

- Updated callout to be of type `warning` and used double quotes instead of blockquote to be more readable. Filed Asana ticket re: the blockquote bug https://app.asana.com/0/1200099998847559/1208117747990951/f
- Fixed whitespacing so indentations look prettier


mirnawong1 (approved):
lgtm!

PR #5973, opened 2024-08-24
Threads config update in teradata-setup.md

VarunSharma15:

Multiple threads are supported by dbt-teradata adapter but the documentation doesn't mention it.
This can confuse users about whether threads are supported or not.
Made the change that mentions it.


matthewshaver (approved):
Thank you @VarunSharma15 !

PR #5974, opened 2024-08-26
[Roadmap] Custom version name functionality

JKarlavige:

[Asana task](https://app.asana.com/0/1200792801231972/1208103052992612/f)

Adds a `customDisplay` property to the `versions` object to provide the option of setting a custom display name for a specific version.

For example, we can show `dbt Cloud` rather than `v1.9` by setting  the version in `website/dbt-versions.js`:
```
{
    version: "1.9",
    isPrerelease: true, 
    customDisplay: "dbt Cloud"
},
```

## Preview

https://docs-getdbt-com-git-custom-version-name-dbt-labs.vercel.app

- Select a different version on both desktop & mobile and verify the version updates
- Refresh the page and verify the version persists after page reload
- Verify the correct version banners appear after updating the version on docs pages

## Notes

Disregard the `v1.9` banner, this only shows because I set the `customDisplay` property on a version which is using `isPrerelease`. Confirmed with Matt they will create a separate version to use the `customDisplay` property which doesn't use the `isPrerelease` flag in prod, and the banner will not show in that scenario.

image

## TODO Before Merge
- [ ] Remove demo "dbt Cloud" version 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208103052992612


matthewshaver (approved):

john-rock (approved):
lgtm!

PR #5980, opened 2024-08-27
Better description in slack unfurls for full_refersh config

joellabes:
I pasted https://docs.getdbt.com/reference/resource-configs/full_refresh into Slack and it unfurled saying to read the guide:

image

Updating the text to instead provide information about the config. There's a bunch of these spread across the docs site but I'm cleaning up as I find them instead of doing a widespread fix.


mirnawong1 (approved):
great PR @joellabes and thanks for opening this up to add a clearer description!

PR #5992, opened 2024-08-29
add versionblocks to cumulative metrics doc

mirnawong1:
this pr adds versionblocks (1.8 and lower, and 1.9+) to the cumulative metrics doc to differentiate the available parameters.

`cumulative_type_params` are only available in “versionless” for dbt-cloud users, which should be mentioned in the documentation.
If you’re a dbt-core user, not dbt-cloud, then you can still use the `grain_to_date` & `window` params, but nested under `type_params` instead of `cumulative_type_params`. The `period_agg` param is not available to dbt-core users yet.

raised in this internal slack thread and by user in dbt community: https://dbt-labs.slack.com/archives/C02NCQ9483C/p1724944680040759

### DIFF
--- a/website/docs/docs/build/cumulative-metrics.md
+++ b/website/docs/docs/build/cumulative-metrics.md
@@ -16,6 +16,8 @@ Note that we use the double colon (::) to indicate whether a parameter is nested
 
 ## Parameters
 
+<VersionBlock firstVersion="1.9">
+
 | Parameter | <div style={{width:'350px'}}>Description</div> | Type |
 | --------- | ----------- | ---- |
 | `name` | The name of the metric. | Required |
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/cumulative-metrics.md
```diff
@@ -32,11 +34,33 @@ Note that we use the double colon (::) to indicate whether a parameter is nested
...
+| --------- | ----------- | ---- |
+| `name` | The name of the metric. | Required |
+| `description` | The description of the metric. | Optional |
+| `type` | The type of the metric (cumulative, derived, ratio, or simple). | Required |
+| `label` | Required string that defines the display value in downstream tools. Accepts plain text, spaces, and quotes (such as `orders_total` or `"orders_total"`). | Required |
+| `type_params` | The type parameters of the metric. Supports nested parameters indicated by the double colon, such as `type_params::measure`. | Required |
+| `window` | The accumulation window, such as 1 month, 7 days, 1 year. This can't be used with `grain_to_date`. | Optional  |
+| `grain_to_date` |Sets the accumulation grain, such as `month`, which will accumulate data for one month and then restart at the beginning of the next. This can't be used with `window`. | Optional |
+| `type_params::measure` | A list of measure inputs | Required |
+| `measure:name` | TThe measure you are referencing. | Optional  |
```
courtneyholcomb: nit: `TThe`

website/docs/docs/build/cumulative-metrics.md
```diff
@@ -101,8 +160,44 @@ metrics:
       cumulative_type_params:
         grain_to_date: month
 ```
+</VersionBlock>
+
+<VersionBlock lastVersion="1.9">
```
courtneyholcomb: Should this be 1.8 instead of 1.9?

courtneyholcomb (approved):
Left a couple of notes inline, but otherwise looks great! Thank you!

PR #6083, opened 2024-09-18
add search to localhost

mirnawong1:
this pr adds search to our localhost env for easier testing

[asana task](https://app.asana.com/0/1200099998847559/1208270609081652/f)

tested and it works!
Screenshot 2024-09-18 at 10 45 42


JKarlavige:
Hey @mirnawong1, `.env` files are intended on being used locally without being pushed to Git. In most cases they are used for storing private API keys locally. Then once the code is being served in production, it pulls the env variables from the hosting provider (Vercel in our case). 

Although we don't have any sensitive API keys here, I would prefer we continue to keep the `.env` files out of Git.

`.env` files are often marked in the [.gitignore file](https://github.com/dbt-labs/docs.getdbt.com/blob/b2d3302ca7a4e0b217d4d466f42c93b101f86663/website/.gitignore#L17), which prevents certain files from accidentally being pushed to Git.

PR #6092, opened 2024-09-18
Add behavior change flag docs for adapter maintainers

### LINKED ISSUE #6047: Update adapter maintainer docs for Behavior Flags
### Contributions

https://docs.getdbt.com/guides/adapter-creation?step=3

The Behavior Flags framework is new functionality that adapter maintainers can use to safely deprecate existing behavior. These changes are generally breaking changes.

I could see this content going near the bottom as a separate section.

### Additional information

See also: https://github.com/dbt-labs/docs.getdbt.com/issues/6048

mikealfare:

We added behavior flags for adapter maintainers to use on first- and third-party adapters. This PR adds documentation explaining how to configure and use behavior flags.

## Checklist
- [x] I have added checklist item(s) to this list for anything anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."
- [x] Approval by Product (@amychen1776)

### DIFF
--- a/website/docs/guides/adapter-creation.md
+++ b/website/docs/guides/adapter-creation.md
@@ -558,11 +558,11 @@ See [this GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/54
 
 ### Behavior change flags
 
-Starting in `dbt-adapters==1.5.0` and `dbt-core==1.8.7`, adapter maintainers have the ability to implement their own behavior change flags.
-For more information on what a behavior change is, please refer to [Behavior changes](https://docs.getdbt.com/reference/global-configs/behavior-changes). Behavior Flags are not intended to be long living feature flags and should be implemented with the expectation that the behavior will be default within an expected period of time. 
-To implement a behavior change flag, you will need to provide a name for the flag, a default setting (`True` / `False`), optional source, and  a description and/or a link to the flag's documentation on docs.getdbt.com. We recommend having a description and documentation link whenever possible. 
-The description and/or docs should provide end users with context for why the flag exists, why they may see a warning, and why they may want to utilize the behavior flag. 
-Behavior change flags can be implemented by overwriting `_behavior_flags()` on the adapter in `impl.py`:
+Starting in `dbt-adapters==1.5.0` and `dbt-core==1.8.7`, adapter maintainers can implement their own behavior change flags. Refer to [Behavior changes](https://docs.getdbt.com/reference/global-configs/behavior-changes)for more information. 
+
+Behavior Flags are not intended to be long-living feature flags. They should be implemented with the expectation that the behavior will be the default within an expected period of time. To implement a behavior change flag, you must provide a name for the flag, a default setting (`True` / `False`), an optional source, and a description and/or a link to the flag's documentation on docs.getdbt.com. 
+
+We recommend having a description and documentation link whenever possible. The description and/or docs should provide end users context for why the flag exists, why they may see a warning, and why they may want to utilize the behavior flag. Behavior change flags can be implemented by overwriting `_behavior_flags()` on the adapter in `impl.py`:
 
 <File name='impl.py'>
 
@@ -589,7 +589,7 @@ class ABCAdapter(BaseAdapter):
 
 </File>
 
-Once a behavior change flag has been implemented, it can be referenced on the adapter both in `impl.py` and in jinja macros:
+Once a behavior change flag has been implemented, it can be referenced on the adapter both in `impl.py` and in Jinja macros:
 
 <File name='impl.py'>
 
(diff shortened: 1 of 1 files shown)


website/docs/guides/adapter-creation.md
```diff
@@ -556,6 +556,106 @@ While much of dbt's adapter-specific functionality can be modified in adapter ma
 
 See [this GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/5468) for information on the macros required for `GRANT` statements:
 
+### Behavior change flags
+
+Starting in `dbt-adapters==1.5.0` and `dbt-core==1.8.7`, adapter maintainers have the ability to implement their own behavior change flags.
+For more information on what a behavior change is, please refer to [Behavior changes](https://docs.getdbt.com/reference/global-configs/behavior-changes).
+To implement a behavior change flag, provide a name, a default setting (`True` / `False`), and optional source, and either a description or a link to the flag's documentation on docs.getdbt.com.
```
amychen1776:
```suggestion
To implement a behavior change flag, you will need to provide a name for the flag, a default setting (`True` / `False`), optional source, and  a description and/or a link to the flag's documentation on docs.getdbt.com. We recommend having a description and documentation link whenever possible. 
```

amychen1776: Also what is source?

website/docs/guides/adapter-creation.md
```diff
@@ -556,6 +556,106 @@ While much of dbt's adapter-specific functionality can be modified in adapter ma
...
+The description and/or docs should provide end users with context for why the flag exists, why they may see a warning, and why they may want to override the default.
```
amychen1776:
```suggestion
The description and/or docs should provide end users with context for why the flag exists, why they may see a warning, and why they may want to utilize the behavior flag. 
```

website/docs/guides/adapter-creation.md
```diff
@@ -556,6 +556,106 @@ While much of dbt's adapter-specific functionality can be modified in adapter ma
...
+    @property
+    def _behavior_flags(self) -> List[BehaviorFlag]:
+        return [
+            {
+                "name": "enable_new_functionality_requiring_higher_permissions",
+                "default": False,
+                "source": "dbt-abc",
+                "description": (
+                    "The dbt-abc adapter is implementing a new method for sourcing metadata. "
+                    "While we feel this is a better way to source metadata, it does require higher permissions on the platform. "
```
amychen1776:
```suggestion
                    "This is a more performant way for dbt to source metadata but requires higher permissions on the platform. "
```

website/docs/guides/adapter-creation.md
```diff
@@ -556,6 +556,106 @@ While much of dbt's adapter-specific functionality can be modified in adapter ma
...
+        {# do the new thing #}
+    {% else %}
+        {# do the old thing #}
+    {% endif %}
+{% endmacro %}
+```
+
+</File>
+
+It's best practice to evaluate a behavior flag as few times as possible. This will make it easier to remove once the behavior change has matured.
```
amychen1776, line 657: >It's best practice to evaluate a behavior flag as few times as possible.

what does this mean?

website/docs/guides/adapter-creation.md
```diff
@@ -556,6 +556,106 @@ While much of dbt's adapter-specific functionality can be modified in adapter ma
 
 See [this GitHub discussion](https://github.com/dbt-labs/dbt-core/discussions/5468) for information on the macros required for `GRANT` statements:
 
+### Behavior change flags
+
+Starting in `dbt-adapters==1.5.0` and `dbt-core==1.8.7`, adapter maintainers have the ability to implement their own behavior change flags.
+For more information on what a behavior change is, please refer to [Behavior changes](https://docs.getdbt.com/reference/global-configs/behavior-changes).
```
amychen1776:
```suggestion
For more information on what a behavior change is, please refer to [Behavior changes](https://docs.getdbt.com/reference/global-configs/behavior-changes). Behavior Flags are not intended to be long living feature flags and should be implemented with the expectation that the behavior will be default within an expected period of time. 
```

amychen1776 (approved):
:shipit:

website/docs/guides/adapter-creation.md
mikealfare: Don't pepper the code with a bunch individual checks like it's a configuration. It should truly represent "a legacy way" and "a novel way".

The `dbt-redshift` example is a good one. Depending on whether we use the SDK or not, we either call the macro, or we call the SDK. We check the flag once.

In more complicated scenarios this may not be possible. Take Iceberg support for example. We need to check the flag in `list relations` and we also need to check it when creating a table in the `create table` macro. So we need to check it twice, there's really no way around that. But let's look at the `create table` macro in particular. We should just have two versions of that macro, the legacy version without Iceberg and the novel way with Iceberg; these are gated by the flag. That's good.

Using two completely different versions of the code will result in a lot of duplicated code though. Both versions need a name, they need to put the sql at the end, they have shared options like `warehouse`, etc. But there are multiple points where they differ as well. There are different config options like `external_volume`. One uses `create table my_table as...` and the other uses `create iceberg table my_table as...`. So there will be an intuitive push towards DRY code where you can embed a check inside the macro at each point where it differs. That's not good. The behavior flag is becoming part of the code, and can more easily contribute to bugs in the code, in particular in the legacy code which was supposed to be gated from this whole change in the first place.

website/docs/guides/adapter-creation.md
mikealfare: `source` is basically a namespace. I would list the adapter as the source for adapter behavior flags. But the framework can be used elsewhere, so I don't force that. I also provide a default that takes the calling module, so that even if the maintainer does not add a source, there's still something that we can use to differentiate flags with the same name across multiple adapters (or not in an adapter at all). I wanted to try and automate it since it would feel silly to need to specify the adapter for every flag. We could probably be smarter than this and set it in `BaseAdapter` somehow when they get registered.

matthewshaver (approved):

matthewshaver:
Thank you @mikealfare !

PR #6099, opened 2024-09-19
Make it clearer that the redirect url needs to be updated

b-per:

Right now, the docs are not really clear that the `OAUTH_REDIRECT_URI` parameter needs to be updated with the value gotten from before. 

We list `'LOCATED_REDIRECT_URI'` (which people need to update), like we list `'CONFIDENTIAL'` which is a fixed string.

## Checklist


Lana-hl (approved):

mirnawong1:
hey @b-per and @Lana-hl - thanks for working together on this! I'm going to tweak this to make it further clear bc you're totally right! and i also think that tags `<>` aren't always understood as 'replace'.

also, in the previous text we say: In the following example create or replace security integration query, replace YOUR_ACCESS_URL with the [appropriate Access URL](https://docs.getdbt.com/docs/cloud/about-cloud/access-regions-ip-addresses) for your region and plan.' 

but i don't see any reference to access URL?

```
CREATE OR REPLACE SECURITY INTEGRATION DBT_CLOUD
  TYPE = OAUTH
  ENABLED = TRUE
  OAUTH_CLIENT = CUSTOM
  OAUTH_CLIENT_TYPE = 'CONFIDENTIAL'
  OAUTH_REDIRECT_URI = 'LOCATED_REDIRECT_URI'
  OAUTH_ISSUE_REFRESH_TOKENS = TRUE
  OAUTH_REFRESH_TOKEN_VALIDITY = 7776000;
```

Lana-hl:
great point @mirnawong1 ! I do think it will make it clearer if we add back `the [appropriate Access URL](https://docs.getdbt.com/docs/cloud/about-cloud/access-regions-ip-addresses) for your region and plan` part

mirnawong1:
hey @b-per and @Lana-hl , how's this looking for you both: [preview link](https://docs-getdbt-lonedskzn-dbt-labs.vercel.app/docs/cloud/manage-access/set-up-snowflake-oauth#create-a-security-integration). we could remove the comment in the code if it's tmi

Screenshot 2024-09-19 at 14 03 50

b-per:
I might recommend removing the comment because I don't know if the statement would work with it.

Nitpicking a bit, I think that `LOCATED_REDIRECT_URI` is an odd name to the parameter as well, but I don't really mind if we don't change it.

mirnawong1:
ok i've updated it to this and added a screenshot that @Lana-hl provided:
Screenshot 2024-09-19 at 15 21 22

mirnawong1:
hey @b-per @Lana-hl , going to merge this pr now. let me know if you have any issues or other thoughts - will be happy to address in another pr! thank you again for the help!

mirnawong1 (approved):

PR #6106, opened 2024-09-20
add sl env var support

mirnawong1:
you can now use ea or en vars to connect to SL. 

this pr creates a snippet to add this info in various places. 

[linear](https://linear.app/dbt-labs/issue/SL-1377/support-for-value-overrides-in-data-platform-credentials#comment-f464e63a) and [slack thread](https://dbt-labs.slack.com/archives/C047UU1H7R9/p1726690127659349)


website/snippets/_sl-env-vars.md
```diff
@@ -0,0 +1 @@
+Use [Extended Attributes](/docs/dbt-cloud-environments#extended-attributes) and [Environment Variables](/docs/build/environment-variables) when connecting to the Semantic Layer. If you set a value directly in the Semantic Layer Credentials, it will have a higher priority than Extended Attributes. When using environment variables, the default value for the environment will be used.
```
mirnawong1, line 1: @aiguofer just double checking that the user would be adding this in their env settings and then adding `DBT_ENV_SECRET` in the SL credentials page?

aiguofer: correct.

For example, using something like `{{env_var('DBT_WAREHOUSE')}}` to set the `warehouse` value in your SL credentials would work.

Similarly, if you set your `account` value in your connection using Extended Attributes with a value like `{{env_var('DBT_ACCOUNT')}}`, we will now be able to do that look up for the Extended Attributes AND the env var.

nataliefiann (approved):
Hiya @mirnawong1 

Thanks for creating this PR. I have no suggestions or recommendation. I've approved this one for you and please let me know if there's anything else I can help with.

Kind Regards
Natalie

PR #6116, opened 2024-09-23
Remove Stemma from the list

b-per:
Stemma seems to not exist anymore. Removing it from the list.


mirnawong1 (approved):
thanks @b-per and great catch! i've also updated the data catalog page too and replaced stemma with secoda [per request from @bobbybirstockdbtlabs ](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1726876624599309)

PR #6178, opened 2024-09-30
Experiment-driven Update: Quicklinks in header 

john-rock:
This pull request adds support for dynamic CTAs in the Guides hero component, extending functionality introduced during [this experiment](https://www.notion.so/dbtlabs/Experiment-docs-Guides-Guides-landing-page-d9d5aa4cba5c4e26be1fda8fd45289da?pvs=4) experiment.

The updated component allows either CTAs that are added to include a link or (in the case of the experiment) an on click event. For this version the on click applies a filter to the guides in the same manner as the experiment.

### Key Changes:
- CTA title can now be set in the Hero.
- CTAs can now be added to the Hero and support two actions: 
   - **Linking:** Direct users to a URL.
   - **On-Click Events:** Trigger JavaScript functions, such as applying a filter to the guides list (as done in the experiment).
- The filtering behavior has been aligned with the experimental version to ensure consistency in user interaction.

## Preview Link
https://docs-getdbt-com-git-experiment-apply-quicklink-hero-dbt-labs.vercel.app/guides

## How to Use the Updated Guides Hero Component

To use the new CTA functionality, follow these steps:

1. **Locate the `heroCTAs` Array:**
   - In the `quickstartGuideList/index.js` file, you will find a `heroCTAs` array.
   - This array consists of objects, each containing two properties:
     - **`title`**: This will be the displayed text of the CTA.
     - **`value`**: This corresponds to the URL query parameter that will be appended when the CTA is clicked, applying the filter.

2. **Ensure Proper `value` Mapping:**
   - The `value` property must match an existing tag in the **Topic** dropdown for the filter to work correctly.
   - When a CTA is clicked, the `value` is passed as a query parameter, filtering the guides in the same way as manually selecting the topic from the filter dropdown.

#### Example using `onClick`:

```javascript
const heroCTAs = [
  {
    title: 'Quickstart Guides',
    value: 'Quickstart' // Must match a tag in the Topic dropdown
  },
  {
    title: 'Use Jinja to improve your SQL code',
    value: 'Jinja' // Must match a tag in the Topic dropdown
  }
];
```

#### Example using `href`:

```javascript
const heroCTAs = [
    {
      title: 'Sign up for dbt Cloud',
      href: 'https://www.getdbt.com/signup',
      newTab: true
    }
];
```

## Why this change?
This update enhances the flexibility of the Guides hero component, allowing for more interactive CTAs that either navigate users to external pages or trigger in-page filtering, improving the overall UX based on lessons learned from the experiment.

### Before merging
- [ ] Confirm if we are launching with the buttons set in this PR


john-rock:
Hi @runleonarun 👋 

Quick question for you – this PR implements the values we used in the previous experiment that added the ability to filter via buttons in the Guides page hero. Since we’re not currently running that experiment, would you prefer I:
1. Merge this PR without any CTAs in the hero, or
2. Merge it with the current CTAs that are already set in the hero?

Let me know which you prefer!

website/src/components/hero/index.js
```diff
@@ -12,6 +12,22 @@ function Hero({ heading, subheading, showGraphic = false, customStyles = {}, cla
...
+              <div className={styles.callToActions}>
+                {callToActions.map((c, index) => (
+                  c.href ? (
+                    <a key={index} href={c.href} className={`${styles.callToAction}`}>
+                      {c.label}
+                    </a>
+                  ) : (
+                    <a key={index} onClick={c.onClick} className={`${styles.callToAction}`}>
+                      {c.title}
+                    </a>
```
JKarlavige: I see `c.href` and `c.label` here which looks intended for being able to link out to urls from the ctas. However in the `quickstartGuideList` index file, it looks like only `title` and `value` are ever passed into the Hero component, into the `title` and `onClick` properties.

```
featuredGuides.map(guide => ({
    title: guide.title,
    onClick: () => handleCallToActionClick(guide.value)
})
```
I tried adding a new object to the `featuredGuides` array with `href` & `label` properties. With this combo the cta appears blank and still tries to run `onClick`. If i switch `label` to `title`, the cta text will show, but `onClick` is still run. 
image

Should it be possible to pass the `href` property into the Hero component to allow linking to urls? Or should this only be limited to activating the guide filters?

website/src/components/quickstartGuideList/index.js
```diff
@@ -111,6 +135,11 @@ function QuickstartList({ quickstartData }) {
         showGraphic={false}
         customStyles={{ marginBottom: 0 }}
         classNames={styles.quickstartHero}
+        callToActions={featuredGuides.map(guide => ({
+          title: guide.title,
+          onClick: () => handleCallToActionClick(guide.value)
+        }))}
+        callToActionsTitle={'Popular guides'}
```
JKarlavige: Here is where `href` and `label` is never passed in

JKarlavige (changes requested):
Really digging this feature! Just have one note on how URLs should be handled when trying to link to from the CTAs

website/src/components/hero/index.js
john-rock: Ope good catch. Just pushed an update allowing the hero to accept the `href` as well. Added an example in there too, lmk what you think.

JKarlavige (approved):
lgtm!! 🚢

website/src/components/quickstartGuideList/index.js
```diff
@@ -98,6 +98,30 @@ function QuickstartList({ quickstartData }) {
     handleDataFilter();
   }, [selectedTags, selectedLevel, searchInput]); // Added searchInput to dependency array
 
+  // Set the featured guides that will show as CTAs in the hero section
+  // The value of the tag must match a tag in the frontmatter of the guides in order for the filter to apply after clicking
+  const heroCTAs = [
+    {
+      title: 'Quickstart Guides',
```
mirnawong1: sugg sentence case
```suggestion
      title: 'Quickstart guides',
```

mirnawong1 (approved):
lgtm!

PR #6209, opened 2024-10-02
add additional ratio metric example

mirnawong1:
this pr adds an additional ratio metric exmaple and explains the advanced keys for numerator/demonimator 

[slack thread](https://dbt-labs.slack.com/archives/C02NCQ9483C/p1727797401568429)


website/docs/docs/build/ratio-metrics.md
```diff
@@ -55,6 +65,30 @@ metrics:
       numerator: food_orders
       denominator: orders
 ```
+</File>
+
+#### Example 2 
+This eample is a ratio metric that calculates the ratio of food orders to total orders, with a filter and alias applied to the numerator:
```
courtneyholcomb: I would just add a note at the end here to make explicit how this is different, something like:
"...with a filter and alias applied to the numerator. Note that in order to add these attributes you'll need to use an explicit key for the `name` attribute, too."

courtneyholcomb (approved):
The example looks great!! I gave one small recommendation inline

PR #6276, opened 2024-10-10
Referenced node in defer

dbeatty10:

The docs [here](https://docs.getdbt.com/reference/node-selection/defer#usage) currently say:

> 1. Is the referenced node included in the model selection criteria of the current run?
> 2. Does the reference node exist as a database object in the current environment?

Since it seems like both items should have the same language, I'm guessing the latter was intended to say "reference**d** node".

## Checklist


mirnawong1 (approved):
great catch, thank you @dbeatty10 !

PR #6411, opened 2024-11-04
add truncate to vale

mirnawong1:
this pr adds a yaml config that checks for  in blog prs.


mirnawong1:
will address this end of november - closing to restart

PR #6451, opened 2024-11-10
fix: to_columns must be an array

mnahkies:
when providing a single column as a string, instead of an array, the compiled sql ends up like this:
```
some_id bigint not null references "database"."schema"."table" (e, x, t, e, r, n, a, l, _, i, d),
```

adjust the example to show an array with a single element instead.

## Checklist
- [x] I have added checklist item(s) to this list for anything anything that needs to happen before this PR is merged, such as "needs technical review" or "change base branch."

### DIFF
--- a/website/docs/reference/resource-properties/constraints.md
+++ b/website/docs/reference/resource-properties/constraints.md
@@ -65,7 +65,7 @@ models:
           - type: unique
           - type: foreign_key
             to: ref('other_model_name')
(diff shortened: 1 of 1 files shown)


mirnawong1 (approved):
hey @mnahkies great catch and thanks so much for opening this pr up to add this! you're absolutely right, going to merge this for you!

PR #6494, opened 2024-11-15
Update behavior-changes.md

jtcohen6:
Given user/customer findings around edge cases in the `restrict_direct_pg_catalog_access` behavior flag (really feature flag) — we do not have any near-term plans to turn this flag ON by default, rather we will be working with the AWS Redshift to understand their plans for addressing those edge cases

@amychen1776 to confirm


mirnawong1:
hey @amychen1776 wanted to check if this pr was ok to review?

amychen1776:
Sorry I missed this one in my emails 😓  @jtcohen6 we have actually ripped this out in later versions due to the user experience and understanding how much more we need to do in order to support data sharing. The plan is to rather create a new behavior flag when time comes to support data sharing fully. So I guess in this table, we should find a way to call this out

amychen1776:
Closing this out because we have removed this behavior flag with https://github.com/dbt-labs/dbt-redshift/pull/940

PR #6540, opened 2024-11-26
Update python-models.md to reflect changed default behaviour

joellabes:
The Python docs mentioned that we intended to change to anonymous sprocs in 1.4: 
>We plan to switch this on for all dbt + Snowpark Python models starting with the release of dbt Core version 1.4.

I have verified that the default is now different: 
https://github.com/dbt-labs/dbt-snowflake/blob/8e027d8fcfbf3958cbc97dfe38fba153d1077a6d/dbt/adapters/snowflake/impl.py#L401 

So am updating the docs accordingly. Since it goes back to 1.4, I haven't bothered with version blocks.

---

- https://docs-getdbt-com-git-joellabes-patch-4-dbt-labs.vercel.app/docs/build/python-models

### DIFF
--- a/website/docs/docs/build/python-models.md
+++ b/website/docs/docs/build/python-models.md
@@ -673,18 +673,7 @@ def model(dbt, session: snowpark.Session):
 
 </VersionBlock>
 
-**About "sprocs":** dbt submits Python models to run as _stored procedures_, which some people call _sprocs_ for short. By default, dbt will create a named sproc containing your model's compiled Python code, and then _call_ it to execute. Snowpark has an Open Preview feature for _temporary_ or _anonymous_ stored procedures ([docs](https://docs.snowflake.com/en/sql-reference/sql/call-with.html)), which are faster and leave a cleaner query history. You can switch this feature on for your models by configuring `use_anonymous_sproc: True`. We plan to switch this on for all dbt + Snowpark Python models starting with the release of dbt Core version 1.4.
-
-<File name='dbt_project.yml'>
-
-```yml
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/python-models.md
```diff
@@ -673,18 +673,7 @@ def model(dbt, session: snowpark.Session):
...
-
-```yml
-# I asked Snowflake Support to enable this Private Preview feature,
-# and now my dbt-py models run even faster!
-models:
-  use_anonymous_sproc: True
-```
-
-</File>
+**About "sprocs":** dbt submits Python models to run as _stored procedures_, which some people call _sprocs_ for short. By default, dbt will use Snowpark's _temporary_ or _anonymous_ stored procedures ([docs](https://docs.snowflake.com/en/sql-reference/sql/call-with.html)), which are faster and leave a cleaner query history than creating and then calling a named sproc containing your model's compiled Python code. You can switch this feature off for your models by configuring `use_anonymous_sproc: False`.
```
mirnawong1:
```suggestion
**About "sprocs":** dbt submits Python models to run as _stored procedures_, which some people call _sprocs_ for short. By default, dbt will use Snowpark's _temporary_ or _anonymous_ stored procedures ([docs](https://docs.snowflake.com/en/sql-reference/sql/call-with.html)), which are faster and keep query history cleaner than named sprocs containing your model's compiled Python code. To disable this feature, set `use_anonymous_sproc: False` in your model configuration. 
```

mirnawong1 (approved):
lgtm @joellabes !

PR #6566, opened 2024-11-29
Update snapshots.md

### LINKED ISSUE #6561: 1.9+ snapshots docs should probably keep the legacy syntax as well
### Contributions

https://docs.getdbt.com/docs/build/snapshots

When viewing docs for v1.9 or Versionless, the "Configure snapshots in versions 1.8 and earlier" section says 

> The latest versions use an updated snapshot configuration syntax that optimizes performance.

This isn't accurate - the code that it generates + runs against the warehouse is going to be the same, but _configuring_ the snapshot in YAML can be more pleasant because there's less fiddly bits. _Edit: I've just seen that the 1.8 docs explain that it's because multiple nodes in a single file is trickier to parse. I guess that is true, but in the grand scheme of things it's not a huge deal._

Separate to that, my preference would be to have _both_ styles of configuration provided for new versions of dbt. New versions of dbt have not removed the classic snapshot syntax and people may still need to see how to configure them. Right now, I think that the classic syntax would just get deleted from the docs site once v1.8 goes EOL, which isn't right. 

### Additional information

_No response_

mirnawong1:
clarify the benefits of the latest snapshot yaml spec

Resolves #6561

---

- https://docs-getdbt-com-git-mirnawong1-patch-26-dbt-labs.vercel.app/docs/build/snapshots

### DIFF
--- a/website/docs/docs/build/snapshots.md
+++ b/website/docs/docs/build/snapshots.md
@@ -487,7 +487,7 @@ Snapshot results:
 
 For information about configuring snapshots in dbt versions 1.8 and earlier, select **1.8** from the documentation version picker, and it will appear in this section.
 
-To configure snapshots in versions 1.9 and later, refer to [Configuring snapshots](#configuring-snapshots). The latest versions use an updated snapshot configuration syntax that optimizes performance.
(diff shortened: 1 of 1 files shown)


nataliefiann (approved):
Hiya @mirnawong1 

This is perfect! I've approved this for you.

Kind Regards
Natalie

PR #6593, opened 2024-12-05
Adding tab borders

### LINKED ISSUE #3889: Suggestion: Add UI distinction to tell tabs are tabs
### Contributions

### What page(s) or areas on docs.getdbt.com are affected?

https://docs.getdbt.com/reference/resource-configs/meta#designate-a-source-column-as-containing-pii is an easy example page because it contains an example at the top of the page that has a lot of "tab" options.

Screenshot 2023-08-08 at 10 33 28 AM

This is in Chrome on Mac Ventura.

### What changes are you suggesting?

As a first time user it's hard to tell that Models, Sources, Seeds, etc. are tabs and that I should click on the tab to see a specific implementation. As I've continued to use the docs over time I still miss that there are other tabs I could refer to. I think adding some kind of box around the title of each tab would help visually identify that those words indicate tabs and that you should click on them.

### Additional information

_No response_

matthewshaver:

Enhancing tabs:
- Places borders around individual tabs
- Sets a unique background color for clicked and hovered tabs
- Puts a border around the entire table area so it's clear where tabbed content ends

See https://www.notion.so/dbtlabs/Documentation-plan-18cbb38ebda7808584e0c13abe8c91d9

Closes https://github.com/dbt-labs/docs.getdbt.com/issues/3889

## Checklist
- [ ] I have reviewed the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) so my content adheres to these guidelines.


mirnawong1 (commented):
looks good to me @matthewshaver !

website/src/css/custom.css
JKarlavige, line 1: Rather than using hard-coded hex values, can we choose from the existing variables available at the top of this `custom.css` file? That will help us stay within our brand colors, while also making it easier to update colors across the board if needed down the road.

website/src/css/custom.css
```diff
@@ -2143,4 +2143,66 @@ table th {
...
+}
+
+.tabs {
+  background-color: transparent;
+  padding: 10px;
+}
+
+/* Default color and border for tabs */
+.tabs__item {
+  background-color: #f9fbfc; /* Sets the tab color the same as the file borders */
```
JKarlavige: In an effort to reduce the amount of custom colors used, would we be able to swap this with the `var(--color-white)` variable. That color is close enough to white where it shouldn't be too noticeable.

website/src/css/custom.css
```diff
@@ -2143,4 +2143,66 @@ table th {
...
+  border: 2px solid #ccc;
```
JKarlavige: Same note about using available color variables here, can we swap all uses of `#ccc` with a similar color currently available as a variable?

website/src/css/custom.css
```diff
@@ -2143,4 +2143,66 @@ table th {
...
+  padding: 5px 10px;
+  margin-right: 5px;
+  border-radius: 4px;
+  cursor: pointer;
+  transition: background-color 0.3s;
+}
+
+/* Default color and border for tabs in dark mode */
+[data-theme='dark'] .tabs__item {
+  background-color: #000000;
```
JKarlavige: To avoid using black (not a brand color), can we swap this and anywhere #000 is used with `--color-primary-blue`, which is set to `#262a38`?

website/src/css/custom.css
```diff
@@ -2143,4 +2143,66 @@ table th {
...
+  padding: 5px 10px;
+  margin-right: 5px;
+  border-radius: 4px;
+  cursor: pointer;
+  transition: background-color 0.3s;
+}
+
+/*Color when hovering over tabs */
+.tabs__item:hover {
+  background-color: #D95B3B;
```
JKarlavige: `#d95b3b` is not an available brand color. Could we swap this with `#DE5D43` or `#ff694a`? `#ff694a` has a variable already, but a new variable would need to be created for `#DE5D43` if you go with that one.

website/src/css/custom.css
```diff
@@ -2143,4 +2143,66 @@ table th {
...
+[data-theme='dark'] .tabs-container{
+  border: 1px solid #ccc;
+  padding: 15px;
+  border-radius: 10px;
+  margin-top: 10px;
+  background-color: transparent;
+}
+
+[data-theme='light'] .tabs-container{
+  border: 1px solid #000000;
```
JKarlavige: In light mode, can we use a lighter color for the border? The `var(--ifm-toc-border-color)` variable might be a good option, which is the current border color for the table of contents.

website/src/css/custom.css
JKarlavige: Also, on hover, the lighter text is still set on the tab while the orange background is applied. This doesn't pass the WCAG accessibility check for contrast [link here](https://webaim.org/resources/contrastchecker/). The text color may need to be updated on hover to match the color when the tab is active.
image

JKarlavige: And on hover, can we have the border color update to match the background color as well? Similar to how the border updates for the active tab.

JKarlavige (changes requested):
Thanks for taking this on Matt. Left some comments with some recommended adjustments. To try and summarize them i'll list them below:
- Swapping hard-coded hex values for available color variables at the top of this file
- Adjusting tab text color on hover
- Adjusting tab border color on hover
- Adjusting section border color in light mode 

JKarlavige:
Hey @matthewshaver, hope it's okay but I pushed a commit swapping the background color for the tabs from orange to teal. The orange wasn't passing accessibility tests with white text (when in dark mode), and I found the contrast between orange background and dark text a bit much.

New preview is building now, if all looks good i'll give it the approval. Let me know if you have any thoughts/concerns with that background color adjustment!

image

JKarlavige (approved):
lgtm! 🚢

matthewshaver:
@JKarlavige I like that color swap and we of course want to stay accessible to the widest audience possible. Thank you for making that change. I'll implement it elsewhere we used the orange (which was a choice to stay with our orange branding, but we're not committed to it).

PR #6655, opened 2024-12-12
add new page about node selection process

### LINKED ISSUE #514: Node selection mental model
It feels like there's endless appetite to provide explanation and example for when resource selection feels counterintuitive. I think we have a lot to gain from laying out exactly the way that node selection _does_ work:

1. dbt selects any resource that is matched by one or more of the `--models`/`--select` criteria, in the order of selection methods (e.g. `tag:`), then graph operators (e.g. `+`), then set operators (`,`). The selected resources may be models, sources, seeds, snapshots, tests. _IFF the command is `dbt test`, dbt also selects any test that directly depends on a selected resource._
2. dbt de-selects any resource that is matched by one or more of the `--exclude` criteria. _IFF the command is `dbt test`, dbt also de-selects any test that directly depends on an excluded resource._
3. dbt now has a list of still-selected resources of varying types. As a final step, it tosses away any resource that does not match the resource type of the current invocation. (Only seeds are kept for `dbt seed`, only models for `dbt run`, only tests for `dbt test`, and so on.)

Things like test selection and the `source:` selection method feel tricky because they intersect with this process is ways that can feel quite different from `dbt run -m my_model`. Rather than write more about each edge case, I think we need to build the mental model, so that they seem less like "exceptions" to rote-memorize and more like tricky cases that are still consistent with the overall schematic.

(Note that step 2 works a bit different for YAML selectors because the `exclude:` key subtly differs from `--exclude`, in order to power more complex set logic. That's briefly documented [here](https://docs.getdbt.com/reference/node-selection/yaml-selectors/#exclude).)

mirnawong1:
this pr adds a new page meant to explain how node selection works for resources. 

Resolves #514

---

- https://docs-getdbt-com-git-add-node-selection-dbt-labs.vercel.app/reference/node-selection/node-selection
- https://docs-getdbt-com-git-add-node-selection-dbt-labs.vercel.app/reference/node-selection/syntax

### DIFF
--- a/website/docs/reference/node-selection/node-selection.md
+++ b/website/docs/reference/node-selection/node-selection.md
@@ -1,0 +1,85 @@
+---
+title: "Node selection"
+description: "Learn how to select and exclude resources in dbt."
+sidebar_label: "About node selection"
+id: node-selection
+---
+
+dbt's resource selection process is consistent across commands but can feel counterintuitive in certain cases. To clarify this, let's outline the steps dbt follows when selecting resources.
+
+To understand dbt resource selection:
+
+- Think of selection and exclusion as two filtering steps applied sequentially.
+- Recognize that the type of command determines the final resources executed.
+- Understand that edge cases like `dbt test` or `source:` aren't exceptions but logical outcomes of this schematic.
+
+### Initial resource selection
+
+dbt selects any resource that matches one or more `--models` or `--select` criteria in the order of:
+
+- Selection methods: Methods like `tag:`, `source:`, `model:`.
+- Graph operators: Operators like `+` (which selects upstream/downstream dependencies).
+- Set operators: Combining resources with `,` or other logical operators.
+- Selected resources may include:
+  - models
+  - sources
+  - seeds
+  - snapshots
+  - 
+  - tests
+
+### Applying --exclude criteria
+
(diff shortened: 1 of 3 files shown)


website/docs/reference/node-selection/syntax.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: "Syntax overview"
+sidebar_label: "Syntax overview"
+pagination_prev: "docs/reference/node-selection/node-selection"
```
mirnawong1:
```suggestion
```

website/docs/reference/node-selection/node-selection.md
```diff
@@ -0,0 +1,82 @@
+---
+title: "Node selection"
+description: "Learn how to select and exclude resources in dbt."
+sidebar_label: "About node selection"
+---
+
+This page outlines the steps dbt follows when selecting resources. To understand dbt resource selection:
```
mirnawong1, line 7: i'd love to clarify if we're talking about nodes here or resources?

mirnawong1: if we're talking commands and flags, it seems like we mean resources adn node nodes?

mirnawong1:
closing issue adn linked pr as content exists here: https://docs.getdbt.com/reference/node-selection/syntax#how-does-selection-work

PR #6700, opened 2024-12-20
clarify full-refresh behavior

### LINKED ISSUE #6698: Clarify full_refresh=true behavior
### Contributions

https://docs.getdbt.com/reference/resource-configs/full_refresh

in the dbt community slack, a [community member raised](https://getdbt.slack.com/archives/C0441GSRU04/p1734689780016539) that the full_refresh doc doesn't explicitly detail what happens when you use the `full_refresh:true` config and what it does.  

we should make it clearer that:
-`full_refresh=true` always refreshes, 
- clarify its not just a boolean string, 
- that it overrides the `--full-refresh` command flag.

User's feedback:
> Hi! I've spotted this doc which I feel could be a little clearer: https://docs.getdbt.com/reference/resource-configs/full_refresh
The full_refresh config effectively acts like an override for the --full-refesh tag in the run command, so when you set to:
False: It prevents all full refreshes - this is clearly documented and we understood this behaviour
True: It always full refreshes - this was the unexpected behaviour
The way that the doc is currently written only focuses on the case where full_refresh=false, and only on careful reading or looking at the source code is it clear what will happen if full_refresh=true vs when it is omitted.
It would be very helpful to update this doc to explicitly state what happens when full_refresh=true and to make it more obvious that this config is not a boolean setting, but rather is an override for the --full-refesh tag.

### Additional information

_No response_

mirnawong1:
this pr clarifies full_refresh config behavior when you set it to `false` and `true`.

Resolves #6698

---

- https://docs-getdbt-com-git-update-full-refresh-dbt-labs.vercel.app/docs/build/incremental-microbatch
- https://docs-getdbt-com-git-update-full-refresh-dbt-labs.vercel.app/reference/resource-configs/full_refresh


runleonarun (approved):
Looks good and ready to merge

PR #6701, opened 2024-12-20
clarify email mismatch in external oauth

mirnawong1:
this pr adds clarifying content/callouts about external oauth doc. this was raised by Sean McIntyre, who helped a customer navigate this as they encountered this issue. 

[refer to internal slack thread](https://dbt-labs.slack.com/archives/C07V8NCQ54K/p1734707506518529)

---

- https://docs-getdbt-com-git-update-external-oauth-dbt-labs.vercel.app/docs/cloud/manage-access/external-oauth


website/docs/docs/cloud/manage-access/external-oauth.md
```diff
@@ -29,6 +29,11 @@ The process of setting up external OAuth will require a little bit of back-and-f
 
 If the admins that handle these products are all different people, it’s better to have them coordinating simultaneously to reduce friction.
 
+:::info Username and credential matching required
```
runleonarun: Should we title this, `Snowflake and idP username matching required` because it sounds like the credentials are not required to match. Or are they? If they are, we might want to specify that below.

runleonarun (approved):
Approved with one minor question/clarification!  Thanks for addressing this @mirnawong1!

website/docs/docs/cloud/manage-access/external-oauth.md
mirnawong1: yep, the email address must match. made it a bit clearer and folded in feedback:

> :::info Snowflake and IdP username matching required
Ensure that the username/email address entered by the IdP admin matches the Snowflake credentials username. If the email address used in the dbt Cloud setup is different from the Snowflake email address, the connection will fail or you may run into issues.
:::

PR #6754, opened 2025-01-10
Update databricks daggers

joellabes:
The databricks table configuration options table used `+` (and `++`) to indicate disclaimers, but [the plus is meaningful](https://docs.getdbt.com/reference/resource-configs/plus-prefix) in configs so this is confusing. I've swapped them out for the more idiomatic dagger and double dagger.

Before and after:
image

I would have liked to be able to superscript these (or even better, have nice tooltips that display inline), but I don't think docusaurus supports that. 

I also clarified that the reason tblproperties can't be set at table creation is that there isn't a PySpark API for it - the "we do not have" was ambiguous as to whether dbt doesn't support it or it doesn't exist.  

## Checklist

---

- https://docs-getdbt-com-git-update-databricks-daggers-dbt-labs.vercel.app/reference/resource-configs/databricks-configs


mirnawong1 (approved):
thanks @joellabes ! good flags and i think we can try to make a custom tooltip so we can communicate disclaimers. this has come up a couple of times.

PR #6789, opened 2025-01-17
Note that dbt parse doesn't put compiled code in its manifest

joellabes:
I was confused about the difference between dbt parse and dbt compile, because since 1.4 they both create a manifest. Turns out that dbt parse doesn't include compiled code, so I added that caveat. 

## Checklist

---

- https://docs-getdbt-com-git-joellabes-patch-4-dbt-labs.vercel.app/reference/commands/parse


joellabes:
Arguably, since 1.4 is an unsupported version, we could just remove the commentary about how it doesn't make a manifest in older versoons? Whoever reviews this is welcome to edit that segment out if they agree it's redundant.

mirnawong1 (approved):
thanks @joellabes !

PR #6845, opened 2025-01-29
add links to lifecycle pill

mirnawong1:
this pr adds appropriate links to the lifecycle pill and ux elements to make it a bit clearer that they're linkable.

it also centralized links using the lifecycle-url.js file and const so we can update links in one place shoudl we need to change them. this change will also apply to new and existing lifecycle pills so there aren't different UX experiences for users. 

[docs notion project](https://www.notion.so/dbtlabs/GBB-docs-changes-18abb38ebda780f08215db2bb5b5b7c4?pvs=4)

requested by @runleonarun  and [asana task for PR review requested](https://app.asana.com/0/1200099998847559/1209270472361332/f)

---

- https://docs-getdbt-com-git-clickable-lifecycle-dbt-labs.vercel.app/docs/introduction


website/docs/docs/introduction.md
```diff
@@ -5,9 +5,19 @@ pagination_next: null
 pagination_prev: null
 ---
```
mirnawong1: will have to remove these test lifecycle pills before pr is merged

website/src/components/expandable/index.js
```diff
@@ -39,43 +39,46 @@ function Expandable({ children, alt_header = null, lifecycle }) {
...
   return (
-    <div id={anchorId} className={`${styles.expandableContainer} `}>
+    <div id={anchorId} className={`${styles.expandableContainer}`}>
       <div className={styles.header} onClick={handleToggleClick}>
         <span className={`${styles.toggle} ${isOn ? styles.toggleDown : styles.toggleRight}`}></span>
         &nbsp;
         <span className={styles.headerText}>
-          {alt_header}<Lifecycle status={lifecycle} />
+          {alt_header}
+          <span onClick={(e) => e.stopPropagation()}>
```
mirnawong1, line 79: this stops the copy link propagation so that users can click the pill for the link or click on the icon to copy the url

john-rock (commented):
Hey @mirnawong1 this is looking good and seems to function as expected.

Just a quick question for my own curiosity.
What is the difference between `Legacy` and `Deprecated`? Just thinking in terms of a user I almost expected `Legacy` to be clickable and listed in the [product lifecycle page ](https://docs.getdbt.com/docs/dbt-versions/product-lifecycles).

mirnawong1:
> Hey @mirnawong1 this is looking good and seems to function as expected.
> 
> Just a quick question for my own curiosity. What is the difference between `Legacy` and `Deprecated`? Just thinking in terms of a user I almost expected `Legacy` to be clickable and listed in the [product lifecycle page ](https://docs.getdbt.com/docs/dbt-versions/product-lifecycles).

hey @john-rock ! hm that's a good question! we have some features that are considered legacy but you can still use them if you want and the documentation isn't going anywhere -- so they're not entirely deprecated (e.g. configuring snapshots using sql config block is legacy since the new way to do it is via YAML config). whereas deprecated means its not supported and docs will be removed. 

i wonder if we should add 'legacy' to the product lifecycle doc? i'll raise that with the PM team! thanks for asking such a great question!

john-rock:
> > Hey @mirnawong1 this is looking good and seems to function as expected.
> > Just a quick question for my own curiosity. What is the difference between `Legacy` and `Deprecated`? Just thinking in terms of a user I almost expected `Legacy` to be clickable and listed in the [product lifecycle page ](https://docs.getdbt.com/docs/dbt-versions/product-lifecycles).
> 
> hey @john-rock ! hm that's a good question! we have some features that are considered legacy but you can still use them if you want and the documentation isn't going anywhere -- so they're not entirely deprecated (e.g. configuring snapshots using sql config block is legacy since the new way to do it is via YAML config). whereas deprecated means its not supported and docs will be removed.
> 
> i wonder if we should add 'legacy' to the product lifecycle doc? i'll raise that with the PM team! thanks for asking such a great question!

Ah I see, thank you for the context on that!

john-rock (approved):
lgtm!

mirnawong1:
omg thank you so much @john-rock ! i really appreciate the quick review! i was a bit nervous about this one and thank you for validating!

PR #6859, opened 2025-02-03
remove repeated word

mirnawong1:
this pr removes repeated word 'are' from the callout


nataliefiann (approved):
Hiya @mirnawong1 

Thanks for creating this. I've approved this for you.

Kind Regards
Natalie

PR #6866, opened 2025-02-04
Update snapshots.md

marcelobour:
This PR suggests changing the example recently added to this document as the result of this [Slack thread](https://getdbt.slack.com/archives/CBSQTAPLG/p1737393637257859).

The key point of the [original question](https://discourse.getdbt.com/t/snapshot-check-strategy-with-valid-from-date/1795) is: "...there is a parent entity that has a “last updated” date that I need to use for new rows instead of the current timestamp. This column is present in the source data, although as it comes from the parent entity instead it can’t be used reliably to indicate changes in the source entity, but when there are changes then it can be used as the start date for those changes."

So we need to check some columns, without including anything related to update dates, and when a change is detected, we want to use our database update-related column instead of the timestamp corresponding to the snapshot execution.

The recently added example isn't incorrect, but doesn't capture what was discussed and solved in the thread, which is a more probable use case situation. Moreover, using a timestamp strategy with a properly defined updated_at field and a check strategy that only checks a column corresponding to the updated time is basically the same. I'd venture to say it could even be less efficient.

In addition to modifying the example, titles, and described logic, a summary of this use case was added to the clarifications section to serve as a reference.

## Checklist

### DIFF
--- a/website/docs/docs/build/snapshots.md
+++ b/website/docs/docs/build/snapshots.md
@@ -371,7 +371,7 @@ snapshots:
 
 In this example:
 
-- If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used.
+- If at least one of the specified `check_cols `changes, the snapshot creates a new row. If the `updated_at` column has a value (is not null), the snapshot uses it; otherwise, it defaults to the timestamp.
 - If `updated_at` isn’t set, then dbt automatically falls back to [using the current timestamp](#sample-results-for-the-check-strategy) to track changes.
 - Use this approach when your `updated_at` column isn't reliable for tracking record updates, but you still want to use it &mdash; rather than the snapshot's execution time &mdash; whenever row changes are detected.
 
--- a/website/docs/docs/build/snapshots.md
+++ b/website/docs/docs/build/snapshots.md
@@ -373,7 +373,7 @@ In this example:
 
 - If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used.
 - If `updated_at` isn’t set, then dbt automatically falls back to [using the current timestamp](#sample-results-for-the-check-strategy) to track changes.
-- Use this approach when your `updated_at` column isn't reliable for tracking record updates, but you still want to use it - rather than the snapshot's execution time - whenever row changes are detected.
+- Use this approach when your `updated_at` column isn't reliable for tracking record updates, but you still want to use it &mdash; rather than the snapshot's execution time &mdash; whenever row changes are detected.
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/build/snapshots.md
```diff
@@ -364,13 +364,16 @@ snapshots:
...
-        - updated_at
+        - status
+        - is_cancelled
+      updated_at: updated_at
 ```
 
 In this example:
 
-- `check_cols: updated_at` makes sure that only the `updated_at` column triggers new snapshots.
+- If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used"
```
mirnawong1:
```suggestion
- If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used.
```

website/docs/docs/build/snapshots.md
```diff
@@ -364,13 +364,16 @@ snapshots:
...
+        - is_cancelled
+      updated_at: updated_at
 ```
 
 In this example:
 
-- `check_cols: updated_at` makes sure that only the `updated_at` column triggers new snapshots.
+- If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used.
 - If `updated_at` isn’t set, then dbt automatically falls back to [using the current timestamp](#sample-results-for-the-check-strategy) to track changes.
+- Use this approach when your `updated_at` column isn't reliable for tracking record updates, but you still want to use it - rather than the snapshot's execution time - whenever row changes are detected.
```
mirnawong1:
```suggestion
- Use this approach when your `updated_at` column isn't reliable for tracking record updates, but you still want to use it &mdash; rather than the snapshot's execution time &mdash; whenever row changes are detected.
```

website/docs/docs/build/snapshots.md
```diff
@@ -364,13 +364,16 @@ snapshots:
...
-        - updated_at
+        - status
+        - is_cancelled
+      updated_at: updated_at
 ```
 
 In this example:
 
-- `check_cols: updated_at` makes sure that only the `updated_at` column triggers new snapshots.
+- If there is any change in at least one of the specified `check_cols`, then a new row is created in the snapshot. If the provided `updated_at` column value is not null, it will be used; otherwise, the timestamp will be used.
```
mirnawong1:
```suggestion
- If at least one of the specified `check_cols `changes, the snapshot creates a new row. If the `updated_at` column has a value (is not null), the snapshot uses it; otherwise, it defaults to the timestamp.
```

mirnawong1 (approved):
thanks so much for clarifying this and correcting @marcelobour ! really appreciate it -- approved!

PR #6874, opened 2025-02-05
Update Cache Control

JKarlavige:

This is an effort to improve caching of our docs site to reduce Vercel overages. The strategy here is to cache all images to the browser for up to 31 days. This should reduce network requests on images, thus reducing our Fast Data Transfer and Edge Requests overages.

One thing i'm not 100% clear on is whether retrieving an item from the CDN cache counts towards our Fast Data Transfer. I believe it does, as I see the Request Header `x-vercel-cache: HIT` consistently across network requests and we are still going over our limit.

CDN requests count towards Edge Requests. And assets loaded from the CDN (such as images), count towards Edge Requests ([docs here](https://vercel.com/docs/pricing/networking#edge-requests)). So this should be able to help with both overages potentially.
image

### Cache settings
Images: Cache images in browser for 31 days, on CDN for 7 days.
Everything else: Cache everything else on CDN for 1 day, & don't cache on the browser.

## Preview

This page has a good amount of images to verify the cache:
https://docs-getdbt-com-git-update-cache-headers-dbt-labs.vercel.app/reference/node-selection/test-selection-examples

In the network tab, reload the page and verify images are now being cached by the browser. You may have to select the `Disable cache` option within the network tab.
image

Select an image in the network tab, and verify the `Cache-Control` matches what is set in the `vercel.json` within this PR.

Verify that other file types such as css, js, etc. use the fallback `Cache-Control` of `max-age=0, s-maxage=86400 stale-while-revalidate`, which caches the item in Vercel's CDN for 1 day.

## Notes

I will give the docs team a heads up when replacing images, the new image should have a unique name. If we ever need to bust the browser image cache, we can use the existing [image-cache-wrapper](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/functions/image-cache-wrapper.js). We can update the `CACHE_VERSION` to 3.

Note on the CDN cache - New Vercel deployments invalidate the CDN cache ([see docs here](https://vercel.com/docs/edge-network/headers/cache-control-headers#recommended-settings)), so it should be safe to cache everything on the CDN without seeing stale content after updates.

---

- https://docs-getdbt-com-git-update-cache-headers-dbt-labs.vercel.app/reference/node-selection/test-selection-examples

## TODO
- [x] Remove test content before merge
- [x] Revert image before merge


breezyfasano (approved):
lgtm! verified all images loaded from the memory cache, and did not when 'Disable cache' was disabled woo. Hoping this helps with the overages

PR #6900, opened 2025-02-12
add intro text component

mirnawong1:
created a component and added it as a frontmatter property to better handle how we introduce intro texts in our docs. (discussed in our docs sync last month)

purpose:  better handle how we introduce intro texts in our docs. gives users a quick snippet of what the page is about without overwhelming content. this should also ensure a consistent look  and feel for intros, and keep the docs cleaner and easier to maintain.

- added new component: website/src/components/IntroText/index.js
- added to the DocItem theme so it uses component as frontmatter: website/src/theme/DocItem/Content/index.js

---

- https://docs-getdbt-com-git-add-intro-text-component-dbt-labs.vercel.app/docs/collaborate/explore-projects
- https://docs-getdbt-com-git-add-intro-text-component-dbt-labs.vercel.app/docs/introduction
- https://docs-getdbt-com-git-add-intro-text-component-dbt-labs.vercel.app/reference/resource-configs/alias


website/src/components/IntroText/styles.module.css
```diff
@@ -0,0 +1,17 @@
...
+  html[data-theme='light'] .introText {
+    color: #4B5563; /* dark text in light mode approved by design (Madelyn)  */
+    font-size: 1.15em;
+  }
+  
+/* Adjusting for Dark Mode */
+html[data-theme='dark'] .introText {
+    color: #9CA3AF; /* white text in dark mode approved by design (Madelyn) */
+    font-size: 1.15em;
+  }
```
JKarlavige: Since light-mode is the default, we can shorten up the css here a bit. The `data-theme='light'` block can be removed entirely, and light-mode will rely on the `.introText` styles. And since the `font-size` is already set within the default `.introText`, it can be removed from the `data-theme='dark'` block.

And if other styles override the `.introText` styles, you can add `html` before `.introText` to give it more specificity.

The updated css would look something like:
```suggestion
/* default styles (assuming light mode) */
html .introText {
  color: #4B5563;
  font-size: 1.15em;
}

/* Adjusting for Dark Mode */
html[data-theme='dark'] .introText {
  color: #9CA3AF;
}
```

website/src/theme/DocItem/Content/index.js
```diff
@@ -52,6 +53,9 @@ export default function DocItemContent({ children }) {
         </header>
       )}
 
+      {/* Render IntroText if frontMatter.intro_text exists */}
+      {frontMatter.intro_text && <IntroText>{frontMatter.intro_text}</IntroText>}
```
JKarlavige, line 63: Great idea adding the option to set the text via frontmatter!

JKarlavige (approved):
Nice addition, all looks good here! I left one comment on shortening the css a bit, but that is optional and not a blocker.

Great work @mirnawong1! 🚀

PR #6975, opened 2025-03-05
Clarify that `accepted_values` test ignores `null` values

owenprough-sift:
The `accepted_values` test ignores `null` values, but that behavior is not apparent when reading the documentation, and could cause folks to incorrectly assume that an `accepted_values` test contains an implicit `not_null` test.

I've made a minor clarification to the description of `accepted_values` to explicitly call out that the test ignores `null` values. If there's a better phrasing or way to do this, let me know.

## Checklist


matthewshaver (approved):

matthewshaver:
Thank you for the clarification @owenprough-sift

PR #6978, opened 2025-03-05
The sample and empty flag

matthewshaver:

- Adds a new section "Optimize development" do the sidebar under "Build"
- Adds the sample flag feature new to v1.10 and updates the upgrade guide
- Gives the empty flag it's own page and versions it back to v1.8 when it was introduced

## Checklist
- [ ] I have reviewed the [Content style guide](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) so my content adheres to these guidelines.

---

- https://docs-getdbt-com-git-sample-dbt-labs.vercel.app/docs/build/empty-flag
- https://docs-getdbt-com-git-sample-dbt-labs.vercel.app/docs/build/sample-flag
- https://docs-getdbt-com-git-sample-dbt-labs.vercel.app/docs/dbt-versions/core-upgrade/05-upgrading-to-v1.10


website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,35 @@
...
+
+Sample mode builds upon the [`--empty` flag](/docs/build/empty-flag) by taking the validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies. 
+
+The `--sample` flag will become more robust over time, but it only supports time-based sampling for now.
+
+## Using the `--sample` flag
+
+The `--sample` flag is available for the [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands, and when used, sample mode generates filtered refs and sources. Since it's using time-based sampling, if you have refs like {{ ref('some_model') }} being sampled, you need to set [`event_time`](/reference/resource-configs/event-time) for `some_model` to the field that will be used as the timestamp.
```
matthewshaver:
```suggestion
The `--sample` flag is available for the [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands, and when used, sample mode generates filtered refs and sources. Since it's using time-based sampling, if you have refs like `{{ ref('some_model') }}` being sampled, you need to set [`event_time`](/reference/resource-configs/event-time) for `some_model` to the field that will be used as the timestamp. 
```

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,35 @@
...
+dbt run --select path/to/stg_customers --sample --time="3 days"
+```
+
+If you have an even larger model, for example, `stg_orders` you can set sample mode to hours:
+
+```
+dbt run --select path/to/stg_customers --sample --time="6 hours"
+```
+
+dbt will then execute the model SQL against the target data warehouse and build the tables with data from the sample size.
```
QMalcolm: Hey sorry for giving you no guidance on this, and thank you for putting this together despite that! ❤️ 

One thing we need to fix is that there is no `--time` flag now, instead all one needs to do is `--sample="6 hours"`.

There are additionally some details we should add. In the example `--sample="6 hours"`, the `"6 hours"` part is what we call the "sample specification". Presently there are two time based sample specifications supported, relative time specs and static time specs. 

The one we've already seen is a "relative time spec", that comes in the form `<int> <granularity>`, and makes your sample filter from `now` to `now - (<int> * <granularity>)`. Supported granularities currently are `hours`, `days`, `months`, and `years`.

Static time specs look like `--sample="{'start': '2025-01-03', 'end': '2025-01-07 12:34:56'}"`. With them you can define a specific `start` and `end` of your sample time filter. Note that the start is "inclusive" and the end is "exlcusive".

QMalcolm (changes requested):

graciegoheen:
Let's add a callout that python models are not currently supported for either `--empty` or `--sample`

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `--sample` flag is available for the [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands, and when used, sample mode generates filtered refs and sources. Since it's using time-based sampling, if you have refs like `{{ ref('some_model') }}` being sampled, you need to set [`event_time`](/reference/resource-configs/event-time) for `some_model` to the field that will be used as the timestamp. 
+
+There are two time-based sample specifications supported for sample mode:
+- **Relative time specs:** Filters sampled data from the time the command is run back to a specified integer and granularity. Supported granularities are:
+    - Hours
+    - Days
+    - Months
+    - Years
+- **Static time specs:** Filters your data between a defined start and end period
```
matthewshaver:
```suggestion
- **Static time specs:** Filters your data between a defined start and end period using date and/or timestamp.
```

website/docs/docs/build/empty-flag.md
```diff
@@ -0,0 +1,34 @@
...
+dbt run --empty
+```
+
+Build a specific model:
+
+```
+dbt run --select path/to/your_model --empty
+```
+
+dbt will then build and execute the SQL, resulting in an empty schema in the data warehouse.
```
runleonarun:
```suggestion
dbt will build and execute the SQL, resulting in an empty schema in the data warehouse.
```

website/docs/docs/build/empty-flag.md
```diff
@@ -0,0 +1,34 @@
+---
+title: "About the empty flag"
+description: "Use the empty flag to test your code and build your tables withoug populating data."
```
runleonarun:
```suggestion
description: "Use the empty flag to test your code and build your tables without populating data."
```

website/docs/docs/build/empty-flag.md
```diff
@@ -0,0 +1,34 @@
...
+
+# About the `--empty` flag
+
+:::note
+
+The `--empty` flag is not currently available for Python models. If the flag is used with a Python model, it will be ignored.
+
+:::
+
+There are cases during dbt development where you'll want to validate that your models are semantically correct without the time-consuming cost of building the entire model in the data warehouse. The [`run`](/reference/commands/run) and [`build`](/reference/commands/run) commands support the `--empty` flag for building schema-only dry runs. The `--empty` flag limits the refs and sources to zero rows. dbt will still execute the model SQL against the target data warehouse but will avoid expensive reads of input data. This validates dependencies and ensures your models will build properly.
```
runleonarun:
```suggestion
During dbt development, you might want to validate that your models are semantically correct without the time-consuming cost of building the entire model in the data warehouse. The [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands support the `--empty` flag for building schema-only dry runs. The `--empty` flag limits the refs and sources to zero rows. dbt will still execute the model SQL against the target data warehouse but will avoid expensive reads of input data. This validates dependencies and ensures your models will build properly.
```

website/docs/docs/build/empty-flag.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Examples
+
+Run the entire project while building only the schemas in your development environment:
```
runleonarun: Just wondering if we should specifically call out "all models" here. Unless you wanted to use a `build` command here where we're building all resources in a project?
```suggestion
Run all models in a project while building only the schemas in your development environment:
```

website/docs/docs/build/empty-flag.md
```diff
@@ -0,0 +1,34 @@
...
+
+```
+dbt run --empty
+```
+
+Build a specific model:
```
runleonarun:
```suggestion
Run a specific model:
```

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+# About the `--sample` flag
+
+:::note
+
+The `--sample` flag is not currently available for Python models. If the flag is used with a Python model, it will be ignored.
+
+:::
+
+Large data sets can drastically increase build times and reduce how quickly dbt developers can build and test new code. The dbt `--sample` flag can help to reduce build times when developing and warehouse spend by running dbt in sample mode. Sample mode enables you to address cases where you don't need to build the entire model during the development or CI cycle but include enough data to validate the outputs.
```
runleonarun:
```suggestion
Large data sets can drastically increase build times and reduce how quickly dbt developers can build and test new code. The dbt `--sample` flag can help to reduce build times and warehouse spend by running dbt in sample mode. Sample mode enables you to address cases where you don't need to build the entire model during the development or CI cycle but include enough data to validate the outputs. 
```

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+:::note
+
+The `--sample` flag is not currently available for Python models. If the flag is used with a Python model, it will be ignored.
+
+:::
+
+Large data sets can drastically increase build times and reduce how quickly dbt developers can build and test new code. The dbt `--sample` flag can help to reduce build times when developing and warehouse spend by running dbt in sample mode. Sample mode enables you to address cases where you don't need to build the entire model during the development or CI cycle but include enough data to validate the outputs. 
+
+Sample mode builds upon the [`--empty` flag](/docs/build/empty-flag) by taking the validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies.
```
runleonarun: Replacing "builds" as it's used a lot in our docs 😬 

Should we also mention when joins will not be populated? which cases?

```suggestion
Sample mode refines the [`--empty` flag](/docs/build/empty-flag), taking the validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies. 
```

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+Sample mode builds upon the [`--empty` flag](/docs/build/empty-flag) by taking the validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies. 
+
+The `--sample` flag will become more robust over time, but it only supports time-based sampling for now.
+
+## Using the `--sample` flag
+
+The `--sample` flag is available for the [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands, and when used, sample mode generates filtered refs and sources. Since it's using time-based sampling, if you have refs like `{{ ref('some_model') }}` being sampled, you need to set [`event_time`](/reference/resource-configs/event-time) for `some_model` to the field that will be used as the timestamp.
```
runleonarun:
```suggestion
The `--sample` flag is available for the [`run`](/reference/commands/run) and [`build`](/reference/commands/build) commands. When used, sample mode generates filtered refs and sources. Since it's using time-based sampling, if you have refs like `{{ ref('some_model') }}` being sampled, you need to set [`event_time`](/reference/resource-configs/event-time) for `some_model` to the field that will be used as the timestamp. 
```

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+There are two time-based sample specifications supported for sample mode:
+- **Relative time specs:** Filters sampled data from the time the command is run back to a specified integer and granularity. Supported granularities are:
+    - Hours
+    - Days
+    - Months
+    - Years
+- **Static time specs:** Filters your data between a defined start and end period using date and/or timestamp.
+
+
+### Example
```
runleonarun:
```suggestion
### Examples
```

runleonarun (commented):
Looks great @matthewshaver! I had just a few questions and suggestions!!

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+:::note
+
+The `--sample` flag is not currently available for Python models. If the flag is used with a Python model, it will be ignored.
+
+:::
+
+Large data sets can drastically increase build times and reduce how quickly dbt developers can build and test new code. The dbt `--sample` flag can help to reduce build times and warehouse spend by running dbt in sample mode. Sample mode enables you to address cases where you don't need to build the entire model during the development or CI cycle but include enough data to validate the outputs. 
+
+Sample mode builds upon the [`--empty` flag](/docs/build/empty-flag) by taking the validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies.
```
matthewshaver: I'm not sure I'd use "refines" here as that would indicate the empty flag is inferior or incomplete when they just serve two different purposes. How about:

```suggestion
Sample mode takes the [`--empty` flag's](/docs/build/empty-flag) validation of semantic results a step further by including a sampling of data from the model(s) in your development schema. It won't solve every scenario; for example, there are cases where not all joins will be populated. However, it presents a viable solution for faster building, testing, and validating many strategies. 
```

website/docs/docs/build/sample-flag.md
matthewshaver: I'm not sure I'd use "refines" here as that would indicate the empty flag is inferior or incomplete when they just serve two different purposes. How about:

website/docs/docs/build/sample-flag.md
runleonarun: Looks great matt!

runleonarun (approved):
🚢 it!

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Next, let's say you want to validate data for your entire business from a sample size further in the past - your busiest week in July, from the first until closing time on the eighth. You can run the following: 
+
+```
+dbt run --sample="{'start': '2024-07-01', 'end': '2024-07-08 18:00:00'}"
+```
+
+dbt will then execute the model SQL against the target data warehouse and build the tables with data from the sample sizes.
+
```
graciegoheen, line 75: @QMalcolm thoughts on adding a note about using .render() to make it so a given ref/source will _not_ be sampled?

website/docs/docs/build/sample-flag.md
```diff
@@ -0,0 +1,57 @@
...
+sidebar_label: "The sample flag"
+pagination_next: null
+pagination_prev: "docs/build/empty-flag"
+---
+
+# About the `--sample` flag
+
+:::note
+
+The `--sample` flag is not currently available for Python models. If the flag is used with a Python model, it will be ignored.
```
graciegoheen, line 13: @QMalcolm confirming this is also the case for snapshots ... and seeds?

website/docs/docs/build/sample-flag.md
QMalcolm: Yea that would be a good thing to include 🤔

website/docs/docs/build/sample-flag.md
QMalcolm: Seeds can be sampled. I.e. `ref(my_seed)` will sample the seed. A seed itself cannot be created as a sample of its inputs because it doesn't actually have any inputs to sample, it is the root node.

For snapshots, during the creation of a snapshot, its inputs will be sampled if running in sample mode. Also, a snapshot _can be_ sampled as an input to another node.

matthewshaver: Do we want to specify that out in the body of the page or does the callout need to be used to address it?

website/docs/docs/build/sample-flag.md
matthewshaver: Would you mind a little guidance on how we want to address this?

QMalcolm: @matthewshaver Something along the lines of
> If you would like a ref to never be sampled when running with `--sample` then append `.render()` to the `ref`, e.g. `{{ ref("my_model").render() }}`

website/docs/docs/build/sample-flag.md
QMalcolm: @matthewshaver I don't think we need to callout anything with regard to snapshots, because they'll just work. @graciegoheen does that sound correct?

As for seeds, I think a callout of something like the following would be nice:
> Seeds will be created as normal, but they will sampled when referenced but down stream nodes

website/docs/docs/build/sample-flag.md
matthewshaver: Thanks @QMalcolm - Added to the doc!

graciegoheen:
Can we make sure the `event_time` config page is updated as well -> https://docs.getdbt.com/reference/resource-configs/event-time

QMalcolm (approved):

PR #7006, opened 2025-03-12
Add info for `package:this` selector

b-per:

Adding `package:this` to the docs as `this` is a special name that can be used to select nodes from the current project.

This was implemented [here](https://github.com/dbt-labs/dbt-core/issues/6891) but was never added to the docs.

## Checklist

---

- https://docs-getdbt-com-git-b-per-patch-1-dbt-labs.vercel.app/reference/node-selection/methods

### DIFF
--- a/website/docs/reference/node-selection/methods.md
+++ b/website/docs/reference/node-selection/methods.md
@@ -143,6 +143,8 @@ selectors unambiguous.
   dbt run --select "snowplow.*"
 ```
 
+The package `this` is also a special one that would select the ndoes of the current project. With the example above, running `dbt run --select "package:this"` from the `snowplow` project would run the exact same set of models as any of the three other selectors.
+
 ### path
 The `path` method is used to select models/sources defined at or under a specific path.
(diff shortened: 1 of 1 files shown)


website/docs/reference/node-selection/methods.md
```diff
@@ -143,6 +143,8 @@ selectors unambiguous.
   dbt run --select "snowplow.*"
 ```
 
+The package `this` is also a special one that would select the ndoes of the current project. With the example above, running `dbt run --select "package:this"` from the `snowplow` project would run the exact same set of models as any of the three other selectors.
```
mirnawong1: fix typo and making it more direct
```suggestion
Use the `this` package to select nodes from the current project. From the example, running `dbt run --select "package:this"` from the `snowplow` project runs the exact same set of models as the other three selectors.

Since `this` always refers to the current project, using `package:this` ensures that you're only selecting models from the project you're working in.
```

mirnawong1 (approved):
this is a super cool feature, thanks so much for capturing it in the docs @b-per - very helpful for our users! ✨

PR #7007, opened 2025-03-12
Add script to check for Qualified cookie to hide Forethought widget

JKarlavige:

[Asana task](https://app.asana.com/0/1200792801231972/1209653584238369)

This PR adds a script which checks for a `QualifiedConversationStarted` cookie which is set from WWW within GTM. This cookie signals that a conversation has been started within the Qualified widget.

If a conversation is active, and the user is linked to the docs site from the widget, the Qualified widget will open on Docs to continue the conversation. If the Qualified widget is open, we should hide the Forethought widget to prevent the two widgets from overlapping.

## Preview
https://docs-getdbt-com-git-hide-forethought-dbt-labs.vercel.app/

## Testing Steps

1. Open the preview above
2. Verify you see the Forethought widget in the bottom-right corner
3. In `DevTools > Application (or Storage if using Firefox)`, create a new cookie called `QualifiedConversationStarted`. The value could be anything, I'm setting it to `true`
4. With the cookie created, refresh the page and verify the Forethought widget hides. The console logs should also confirm the element was found and hidden

## TODO Before Merge
- [x] Update selector to get `forethought-chat` element by ID 
- [x] Clear console logs

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -294,6 +294,7 @@ var siteSettings = {
     "/js/gtm.js",
     "/js/onetrust.js",
     "/js/mutiny.js",
+    "/js/hide-forethought.js",
   ],
   stylesheets: [
     "/css/fonts.css",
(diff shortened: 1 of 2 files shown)


john-rock (approved):
lgtm. Confirmed forethought widget is hidden after setting `QualifiedConversationStarted=true`

image

PR #7080, opened 2025-03-24
Update utils for dbt-fabric to tsql-utils 

marzaccaro:
Added a link to the tsql-utils, so it's easier to dispatch to the right fabric macros the dbt-utils macros, or even macros from other packages

I am editing this because looks like the tsql-utils package is supporting not only dbt-utils but other packages as well.

When using dbt-codegen in combination with the dbt-fabric adapter the `generate_source macro` is erroring because rather than using the `fabric__get_tables_by_pattern_sql.sql` the `default__get_tables_by_pattern_sql.sql` is used by dbt-codegen.

The way to solve this would be to follow the installation instructions for tsql-utils [here](https://github.com/dbt-msft/tsql-utils?tab=readme-ov-file#installation-instructions) and:

- install this package
- edit the dispatch setting of the project to check first tsql-utils and then the dbt-utils package

Also, if we link this package here rather than the utils folder within the package, it's easier for fabric adapter users to start using the dbt-utils and other packages, and maybe contribute to the package as well.

Hope it makes sense.
Long explanation for a tiny change!

## Checklist

- [x] The content in this PR requires a dbt release note, so I added one to the [release notes page](https://docs.getdbt.com/docs/dbt-versions/dbt-cloud-release-notes).).


mirnawong1 (approved):
hey @marzaccaro thanks so much for opening this up! going to cc @amychen1776 so they're aware too!

PR #7174, opened 2025-04-09
clarify exclude and include

### LINKED ISSUE #3665: Model versions: include/exclude only relevant if defining (a lot of) `columns`
### Contributions

https://docs.getdbt.com/docs/collaborate/govern/model-versions
https://docs.getdbt.com/reference/resource-properties/include-exclude

Copying from [internal slack](https://dbt-labs.slack.com/archives/C02HE19D51R/p1688130403322139?thread_ts=1688113653.925449&cid=C02HE19D51R):

> If none of your model versions define any columns — you don't actually need to define `columns` at all, and can totally skip the include/exclude situation. This isn't clear in the docs, because I was thinking about versioned models as also being contracted (= all `columns` defined).

While here: Should also try to disambiguate from the `include`/`exclude` syntax used for [`--warn-error-options`](https://docs.getdbt.com/reference/global-configs/warnings) (same syntax, different use case); and the `--exclude` syntax for node selection (totally different).

### Additional information

_No response_

mirnawong1:
this pr clarifies that exclude/include are only needed if you have columns. if you don't, you don't need to define them at all using the column/include/exclude keys. 

it also addes a callout to distinguish difference ways the include/exclude syntax is used.

Resolves #3665

---

- https://docs-getdbt-com-git-clarify-include-exclude-dbt-labs.vercel.app/docs/collaborate/govern/model-versions
- https://docs-getdbt-com-git-clarify-include-exclude-dbt-labs.vercel.app/reference/resource-properties/versions

### DIFF
--- a/website/docs/docs/collaborate/govern/model-versions.md
+++ b/website/docs/docs/collaborate/govern/model-versions.md
@@ -281,6 +281,8 @@ models:
 
 </Tabs>
 
+Note: If none of your model versions specify columns, you don't need to define columns at all and can omit the `include`/`exclude` keys from the versioned model. In this case, dbt will automatically use all top-level columns for all versions. 
+
 The configuration above says: Instead of two unrelated models, I have two versioned definitions of the same model: `dim_customers_v1` and `dim_customers_v2`.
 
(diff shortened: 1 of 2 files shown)


nataliefiann (approved):
Hiya @mirnawong1 

Thanks for creating this PR. No suggestions from me on this. This has been approved.

Kind Regards
Natalie

PR #7250, opened 2025-04-28
updates cloud parity table

mirnawong1:
updates st parity for SL since SL is avail in all tenancies. also fixes table formatting so its easier to read


nataliefiann (approved):
Hiya @mirnawong1 

Thanks for creating this PR! No suggestions or recommendations from me. That's all been approved for you.

Kind Regards
Natalie

PR #7268, opened 2025-04-30
Update databricks-setup.md

Siete-F:
I tried this `ansi_mode: true` configuration in my profiles.yml file, but (at least for me) it doesn't work.
See my DBT question [here](https://discourse.getdbt.com/t/ansi-mode-does-not-seem-to-work/19223) also.
The solution was using a different variable.

Might the "ansi_mode: true" setting only be suitable for a specific Databricks version? I am using a rather recent one 16.3.

After some trying (see link above) I resorted to using `spark.sql.ansi.enabled: true` which works.

## Checklist


amychen1776:
Hello @Siete-F - I'm curious what version of dbt-databricks you're using?

Siete-F:
Hi @amychen1776 , that must have been dbt-databricks 1.10.

As a check, does this result in an overflow error on your databricks? assuming you have an environment available?

```
SET ansi_mode = true;

SELECT cast(12345 AS TINYINT);
```
For me, this just overflows without error.

amychen1776:
Hm did it work on 1.9?
Tagging in the dbt-databricks adapter maintainer to get more clarification: 
@benc-db since all of your 1.10 major re-architecting was under a behavior flag - is this a regression?

benc-db:
I think the original comment is correct, that this is a DBR change, rather than a dbt change.  I don't see any reference to ansi_mode in the dbt-databricks repo.

amychen1776:
Oh so it's on the driver.  Thank you for the confirmation -> we should update the docs then. @benc-db do you know what versions of the python driver this started?

benc-db:
> Oh so it's on the driver. Thank you for the confirmation -> we should update the docs then. @benc-db do you know what versions of the python driver this started?

I wouldn't expect this to be a driver thing, since the driver mostly just passes the session_configuration without modification.  And if its a DBR change, it would have to be very new since the docs suggest it still works https://docs.databricks.com/aws/en/sql/language-manual/parameters/ansi_mode

amychen1776:
Oh interesting. @benc-db how should we update the docs here?  What should be the recommendated config

benc-db:
@amychen1776 I don't actually know; if @Siete-F can narrow down if changing DBR or changing dbt can get the documented setting to work, then we'd know what part of their env is causing the recommendation not to work, which would inform what we document.

amychen1776:
Given the lack of responses for this PR - I'm going to go ahead and close this out to clean up our backlog. Let's have the conversation on the dbt-databricks repo first if this come up again.

PR #7270, opened 2025-04-30
Fix example for tag based node selection

marcelobour:
It seems that when using examples from other docs, this particular line wasn’t updated to fit the other tag examples.


mirnawong1 (approved):
hey @marcelobour oh that's a great catch! yep this is totally spot on - thanks and merging for you!

PR #7322, opened 2025-05-15
Fix partial_parse flag

adrianoesch:
I tried following the documentation but then got the error 
> 12:28:20  Encountered an error:
Runtime Error
  Do not specify both 'config' in profiles.yml and 'flags' in dbt_project.yml. Using 'config' in profiles.yml is deprecated.

## Checklist
- [x] Needs technical review
- [ ] Change base branch
- [ ] Merge on xyz date

### DIFF
--- a/website/docs/reference/global-configs/parsing.md
+++ b/website/docs/reference/global-configs/parsing.md
@@ -6,13 +6,13 @@ sidebar: "Parsing"
 
 ### Partial Parsing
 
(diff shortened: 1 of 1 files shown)


mirnawong1:
Hey @adrianoesch thanks for raising this. Can I double check what dbt version you're on? I believe the config flag in the profiles.yml is supported for v 1.7 and older (see [here](https://docs.getdbt.com/docs/dbt-versions/core-upgrade/upgrading-to-v1.8#quick-hits))

So it might be best to add version blocks to the content so the correct content displays for the appropriate versions, which we can do

matthewshaver (approved):
Thank you @adrianoesch Good catch!

PR #7325, opened 2025-05-15
Remove unwanted semicolon from SL quickstart

joellabes:
This fails during dbt run because the semicolon breaks the create table statement

---

- https://docs-getdbt-com-git-joellabes-patch-4-dbt-labs.vercel.app/guides/mf-time-spine


mirnawong1 (approved):

mirnawong1:
Great catch thanks @joellabes !

PR #7343, opened 2025-05-22
add callout to deprecations page

graciegoheen:
Adding a callout to clarify what "deprecated" means on deprecations page - see [internal slack thread](https://dbt-labs.slack.com/archives/C08JLMFRAP6/p1747953186263939?thread_ts=1747952665.893659&cid=C08JLMFRAP6)

### DIFF
--- a/website/docs/reference/deprecations.md
+++ b/website/docs/reference/deprecations.md
@@ -6,7 +6,7 @@ title: "Deprecations"
 
 Deprecated functionality still works in the v1.10 release, but it is no longer supported and will be removed in a future version.  
 
-This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've customized the behavior). 
+This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've configured [warnings as errors](/reference/global-configs/warnings)). 
 
 When the functionality is eventually removed, it will cause errors in your dbt runs after you upgrade if the deprecations are not addressed.
 
--- a/website/docs/reference/deprecations.md
+++ b/website/docs/reference/deprecations.md
@@ -8,7 +8,7 @@ title: "Deprecations"
 
 :::
 
-As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. It is important to resolve any deprecation warnings in your project before the changes are made.
+As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. Although it’s just a warning for now, it is important to resolve any deprecation warnings in your project to enable you to work with more safety, feedback, and confidence going forward.
 
 ## Identify deprecation warnings
 
--- a/website/docs/reference/deprecations.md
+++ b/website/docs/reference/deprecations.md
@@ -4,7 +4,12 @@ title: "Deprecations"
 
 :::note
 
-"Deprecated functionality" means that the functionality still works today, but only for historical purposes and it will be removed in a future version.
+Deprecated functionality still works in the v1.10 release, but it is no longer supported and will be removed in a future version.  
+
+This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've customized the behavior). 
+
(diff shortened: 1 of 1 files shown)


website/docs/reference/deprecations.md
```diff
@@ -2,6 +2,12 @@
 title: "Deprecations"
 ---
 
+:::note
```
graciegoheen, line 5: open to other formatting / wording here!

QMalcolm (approved):

website/docs/reference/deprecations.md
graciegoheen: trying to echo the sentiment from the [deprecations warning github post](https://github.com/dbt-labs/dbt-core/discussions/11493):

> We know it will take existing users time to migrate their projects, and it is not our goal to break anyone overnight. The goal is to enable dbt users to work with more safety, feedback, and confidence going forward.

graciegoheen:
I think it would also be great to add a callout for:

> ⚠️ If you are using the —-warn-error flag (or --warn-error-options '{"error": "all"}') to [promote all warnings to errors](https://docs.getdbt.com/reference/global-configs/warnings), this will include new deprecation warnings coming to Core. If you don’t want these to be promoted to errors, the --warn-error-options flag gives you more granular control over exactly which types of warnings are treated as errors - you can set "warn": ["Deprecations"] (https://github.com/dbt-labs/dbt-common/issues/265 as of 1.10) to continue treating the deprecation warnings as warnings.

([like what we have in the 1.10 upgrade guide](https://docs.getdbt.com/docs/dbt-versions/core-upgrade/upgrading-to-v1.10#deprecation-warnings:~:text=If%20you%20are,warnings%20as%20warnings.))

@matthewshaver do you have ideas for format for this?

website/docs/reference/deprecations.md
```diff
@@ -2,6 +2,12 @@
...
+
+"Deprecated functionality" means that the functionality still works today, but only for historical purposes and it will be removed in a future version.
+
+:::
+
 As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. It is important to resolve any deprecation warnings in your project before the changes are made.
```
graciegoheen:
```suggestion
As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. Although it’s just a warning for now, it is important to resolve any deprecation warnings in your project to enable you to work with more safety, feedback, and confidence going forward.
```

website/docs/reference/deprecations.md
```diff
@@ -2,7 +2,13 @@
 title: "Deprecations"
 ---
 
-As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. It is important to resolve any deprecation warnings in your project before the changes are made.
+:::note
+
+"Deprecated functionality" means that the functionality still works today, but only for historical purposes and it will be removed in a future version.
```
matthewshaver:
```suggestion
Deprecated functionality still works in the v1.10 release, but it is no longer supported and will be removed in a future version.  

This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've customized the behavior). 

When the functionality is eventually removed, it will cause errors in your dbt runs after you upgrade if the deprecations are not addressed.

```

matthewshaver (approved):

website/docs/reference/deprecations.md
```diff
@@ -2,7 +2,18 @@
 title: "Deprecations"
 ---
 
-As dbt runs, it generates different categories of [events](/reference/events-logging), one of which is _deprecations_. Deprecations are a special type of warning that lets you know that there are problems in parts of your project that will result in breaking changes in a future version of dbt. It is important to resolve any deprecation warnings in your project before the changes are made.
+:::note
+
+Deprecated functionality still works in the v1.10 release, but it is no longer supported and will be removed in a future version.  
+
+This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've customized the behavior).
```
matthewshaver:
```suggestion
This means the deprecated features only present a warning but don't prevent runs and other commands (unless you've configured [warnings as errors](/reference/global-configs/warnings)). 
```

PR #7347, opened 2025-05-23
fix defaults for behavior change flags

graciegoheen:
Fix docs on defaults for behavior change flags

from -> https://github.com/dbt-labs/dbt-core/issues/11609 and https://github.com/dbt-labs/dbt-core/issues/11610


runleonarun (approved):
These look good!

PR #7356, opened 2025-05-27
Update snapshot-configs.md

aglibert393192:
adding relation everywhere, because it doesn't seem to work without it, and that cost me like 6 hours of trying to understand where my understanding of dbt was wrong


matthewshaver (approved):
Thank you @aglibert393192 This is a good call out. We put them in the definition sheets but not the actual examples.

PR #7403, opened 2025-06-02
Redesign - Foundational Styles

john-rock:
This PR adds the foundational styles (fonts, colors, typography, etc.).

There are 3 new css files (`colors`, `typography`, and `tokens`) that were transferred from the WWW repo. We can use these as needed to build out the redesign. Thanks @breezyfasano for the heavy lift over there.

Also, custom.css has been slightly updated to start pulling in these values and updating the existing docusaurus `ifm` classes. This will most likely need to be continually updated as we build but this should be a good start.

Additionally, there are a few component css files that have their values updated from a find-and-replace against old colors. Again, these along with other component css files will need to be updated as we build out.

## Preview
https://docs-getdbt-com-git-redesign-foundational-styles-dbt-labs.vercel.app/


breezyfasano (approved):
this is looking awesome!! it's really dope how good the site looks already 👀  lgtm!

PR #7407, opened 2025-06-03
Fix typo in static_analysis config

jaypeedevlin:

Fixes an error in the config example


mirnawong1 (approved):
great catch, thanks @jaypeedevlin !!!

PR #7423, opened 2025-06-05
Add freshness examples

luna-bianca:
Added freshness config samples 

## Checklist
- [X] The changes in this PR meet the [docs style guide/fundamentals](https://github.com/dbt-labs/docs.getdbt.com/blob/current/contributing/content-style-guide.md) required for all content.


website/docs/docs/deploy/state-aware-setup.md
```diff
@@ -163,16 +166,56 @@ models:
...
+        updates_on: all 
+```
 </File>
 
+</TabItem>
+<TabItem value="sql" label="Config block">
+
+<File name="models/<filename>.sql">
+  
+```sql
```
matthewshaver: Looks like this might be missing some formatting 
- A bracket missing after `build_after`
- Commas after count, period, update_on
- Indentations

```
{{
    config(
        freshness={
            "build_after": {
                "count": 4,
                "period": "hour",
                "updates_on": "all"
            }
        }
    )
}}
```

matthewshaver: The config block does go into a .sql file, but the blocks themselves are formatted in Jinja
```suggestion
```jinja
```

website/docs/docs/deploy/state-aware-setup.md
```diff
@@ -138,11 +138,14 @@ These configurations are powerful because you can define a sensible default at t
 
 Let's use an example to illustrate how to customize our project so a model and its parent model are rebuilt only if they haven't been refreshed in the past 4 hours &mdash; even if a job runs more frequently than that.
 
-A Jaffle shop wants has recently expanded globally and wanted to make savings. To reduce spend, they found out about <Constant name="cloud" />'s state-aware orchestration and want to rebuild models only when needed. Maggie &mdash; the analytics engineer &mdash; wants to configure her dbt `jaffle_shop` project to only rebuild certain models if they haven't been refreshed in the last 4 hours, even if a job runs more often than that. 
+A Jaffle shop has recently expanded globally and wanted to make savings. To reduce spend, they found out about <Constant name="cloud" />'s state-aware orchestration and want to rebuild models only when needed. Maggie &mdash; the analytics engineer &mdash; wants to configure her dbt `jaffle_shop` project to only rebuild certain models if they haven't been refreshed in the last 4 hours, even if a job runs more often than that.
```
matthewshaver, line 141: Great catch!

website/docs/reference/model-configs.md
```diff
@@ -423,3 +423,30 @@ models:
...
 
 </File>
+
+<VersionBlock firstVersion="1.10">
+
+### Configuring source freshness
+
+The model `freshness` config rebuilds models only when new source or upstream data is available. This is useful for models that depend on other models but only need to be updated periodically. For more information, see [freshness](/reference/resource-configs/freshness).
+
+See the following example of a `model.yml` file using the `freshness` config:
```
matthewshaver: We're pretty flexible on the filename convention used in examples, but just to avoid confusion while scanning, it might be helpful to add a little prefix to the model name to distinguish it more from the directory name
```suggestion
See the following example of a `my_model.yml` file using the `freshness` config:
```

website/docs/reference/model-configs.md
```diff
@@ -423,3 +423,30 @@ models:
...
+
+<File name="models/<filename>.yml">
```
matthewshaver: and then we can use that name here instead of a placeholder
```suggestion
<File name="models/my_model.yml">
```

website/docs/docs/deploy/state-aware-setup.md
luna-bianca: Hi @matthewshaver , I tried changing this to jinja but it looks like this when rendered:
Screenshot 2025-06-05 at 11 26 41 PM

And this is how it looks like when sql is used:
Screenshot 2025-06-05 at 11 26 27 PM

I also checked other SQL file samples and they also use "sql". Please confirm if this should still be changed.

website/docs/reference/model-configs.md
```diff
@@ -423,3 +423,30 @@ models:
 ```
 
 </File>
+
+<VersionBlock firstVersion="1.10">
+
+### Configuring source freshness
```
luna-bianca, line 429: Should this be:
```suggestion
### Configuring model freshness
```

luna-bianca (commented):
Hi @matthewshaver , I have some questions below, please check. Thanks!

website/docs/reference/model-configs.md
matthewshaver: Great question! That can remain as "source freshness" as we're not measuring whether the model itself is fresh, but rather the source data it depends on.

website/docs/docs/deploy/state-aware-setup.md
matthewshaver: Excellent question - we do use Jinja frequenly inside the .sql model files. I know it's a little off putting that it doesn't render the same way (with nice colors and everything) but the config block is Jinja

matthewshaver (approved):

PR #7426, opened 2025-06-05
Redesign - Link Section

john-rock:
Adds the new link section to the homepage based on the wireframe
image

## Preview Link
https://docs-getdbt-com-git-redesign-link-section-dbt-labs.vercel.app/

## Additional Information
- Will follow up with docs team to confirm which links they want in this section. Using the content from the wireframe for now.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -2087,6 +2087,42 @@ section>h2:not(.resource-section) {
   }
 }
 
+.home-link-grid {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 24px;
+  margin-top: 4rem;
+}
+
+.home-link-grid-item {
+  display: flex;
+  flex-direction: column;
+  gap: 12px;
+}
+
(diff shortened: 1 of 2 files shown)


website/src/css/custom.css
```diff
@@ -2087,6 +2087,42 @@ section>h2:not(.resource-section) {
...
+
+.home-link-grid-item a:hover {
+  text-decoration: underline;
+}
+
+@media (max-width: 996px) {
+  .home-link-grid {
+    grid-template-columns: 1fr;
+  }
+}
```
JKarlavige: What do you think of having them go to 2 columns at 996px, and 1 column at a smaller screen size, maybe 768px? There's a good amount of space once they drop down to 1 column initially
image

john-rock: Yeah that's a good call. Just pushed an update for 2 columns at 996px and 1 up to 768px. Looks much better.
image

JKarlavige (approved):
lgtm! 🚀

PR #7433, opened 2025-06-06
Redesign - End of Page CTA

john-rock:
Adds the end of page CTA component.
image

## Preview Links
https://docs-getdbt-com-git-redesign-end-of-page-cta-dbt-labs.vercel.app/


website/static/img/bottom-cta-bg.jpg
JKarlavige, line 1: For this bg, i just uploaded it to the public assets repo and is available here: https://public.cdn.getdbt.com/images/training/end-of-page-cta-bg.jpg

I optimized it, so it's lower quality but only 232kb. What are your thoughts on using that, instead of this one, to try and cut down bandwidth from docs?

website/src/pages/index.js
```diff
@@ -267,6 +267,20 @@ function Home() {
               </div>
             </div>
           </section>
+
+          <section className='bottom-cta'>
+            <div className='container'>
+              <div className='cta-section-text'>
+                <span className="eyebrow">Get started</span>
+                <h2 className="heading-2">Start building with dbt.</h2>
+                <p>Streamline your data transformation process, reduce manual errors, and increase productivity with dbt. Sign up today an take your data transformation workflow to the next level.</p>
```
JKarlavige: I believe this is a typo: "Sign up today an take your data". "an" should be "and" it looks like

JKarlavige (changes requested):
This new homepage is looking so good with each new section. 

Just see one typo in the paragraph for this section for this, all else looks great!

website/src/pages/index.js
john-rock: Ope

JKarlavige (approved):
lgtm! 🚀

PR #7469, opened 2025-06-12
add overflow adjustment

john-rock:
- Fixes overflow on search popup

## Testing
I couldn't trace to exact cause of why the overflow was now showing after not previously making adjustments to the search popup. It previously had overflow-x hidden and I updated that to encompass both overflows to be hidden. Feel free to test that adjustment but I don't see any negative side effects so far.

https://docs-getdbt-com-git-fix-overflow-dbt-labs.vercel.app/


JKarlavige (approved):
🚀

PR #7527, opened 2025-06-23
docs: fix arg for run_query macro

sushi30:
based on https://github.com/dbt-labs/dbt-core/blob/70ad9319d2b791a3b89ca167deb23404ae0e9f05/tests/functional/artifacts/data/state/v1/manifest.json#L1929


matthewshaver (approved):
Good catch! Thanks @sushi30

PR #7550, opened 2025-06-25
Remove monorepo FAQ for VSCE v0.2.8

drewbanin:

Small update removing an FAQ about the dbt VS Code extension. The caveat listed is no longer a problem in v0.2.8 of the extension, so I think we should remove it!


runleonarun (approved):
🚢 it! Set to auto-merge when tests pass

PR #7555, opened 2025-06-25
Fix - Break logo cache

john-rock:
This PR adds params to bust the cache on the dbt logo paths used within `docusaurus.config.js` in an effort to prevent the previous logos from showing


JKarlavige (approved):
cache = busted

lgtm 🚀

PR #7556, opened 2025-06-25
Fix nav border

JKarlavige:
[Notion task](https://www.notion.so/dbtlabs/Longer-version-name-offsets-header-line-20fbb38ebda780a78670dae71b0c03d4?source=copy_link)

This fixes an issue where the navigation pushes the logo container border to the left, causing it to become unaligned with the sidebar border. I adjusted the media query to hide the border earlier.

This PR also adjusts the nav search, to switch to the search icon earlier to provide more room for the nav items as the screen still gets smaller.

There's still overlap on the logo at the smallest desktop screen size before the mobile nav kicks in. We would have to adjust the mobile nav to kick in earlier to solve for this. But we can handle that in a separate task if requested to do so.

## Preview

To test, select the `dbt Fusion engine (v2.0 Beta)` version in the nav, which is the longest version name. Reduce the screen size, and verify the border disappears before being pushed to the left.

### DIFF
--- a/website/src/components/discourse/styles.module.css
+++ b/website/src/components/discourse/styles.module.css
@@ -57,8 +57,10 @@
   color: #fff;
 }
 .discourseCta:hover {
-  background: var(--color-light-teal);
+  background: var(--ifm-color-primary);
   color: #fff;
+  text-decoration: underline;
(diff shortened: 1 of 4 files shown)


john-rock (approved):
:shipit:

PR #7560, opened 2025-06-26
Remove grant of OPERATE on the warehouse

chriskl:
I can't see any reason why the metadata service needs to be able to edit the warehouse's configuration?

### DIFF
--- a/website/docs/docs/explore/external-metadata-ingestion.md
+++ b/website/docs/docs/explore/external-metadata-ingestion.md
@@ -60,7 +60,7 @@ CREATE OR REPLACE ROLE dbt_metadata_role;
 2. Grant access to a warehouse to run queries to view metadata:
 
 ```sql
-GRANT OPERATE, USAGE ON WAREHOUSE "<your-warehouse>" TO ROLE dbt_metadata_role;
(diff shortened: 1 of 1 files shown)


matthewshaver:
Thanks for opening this @chriskl It's a good point.

`Operate` is required only to start the warehouse. If the warehouse is already running during ingestion or has auto-resume enabled, this permission is not required. We'll add a note to the setup that states as much so folks who don't need it aren't adding it

website/docs/docs/explore/external-metadata-ingestion.md
```diff
@@ -60,7 +60,7 @@ CREATE OR REPLACE ROLE dbt_metadata_role;
 2. Grant access to a warehouse to run queries to view metadata:
 
 ```sql
-GRANT OPERATE, USAGE ON WAREHOUSE "<your-warehouse>" TO ROLE dbt_metadata_role;
+GRANT USAGE ON WAREHOUSE "<your-warehouse>" TO ROLE dbt_metadata_role;
 ```
```
matthewshaver, line 65:
```suggestion

If your warehouse needs to be restarted for metadata ingestions (doesn't have auto-resume enabled), you may need to grant `OPERATE` permissions to the role as well. 
```

matthewshaver (approved):

PR #7678, opened 2025-07-23
Add missing mention of Linux where it mentions Mac/Windows.

nevdelap:
Further down the page it tells you how to do the Linux installation. I was momentarily disappointed when I thought it couldn't be used on Linux. :smile:


matthewshaver (approved):
Good catch Thanks @nevdelap

PR #7702, opened 2025-07-25
Refactor Examples section

### LINKED ISSUE #7683: Refactor examples for `--resource-type` and `--exclude-resource-type` flags
### Contributions

https://docs.getdbt.com/reference/global-configs/resource-type#example

Two things:
1. The "Example" heading is singular, but it should be pluralized, because there are multiple examples
2. There's a lot of bullet-points which makes it harder to find the example you are interested in, and it won't show up in the right-hand summary menu

So we should:
- Pluralize the heading to "Examples"
- Make one sub-heading per example
- Make it easy to see the difference between "include" examples and "exclude" examples

### Additional information

I'd be happy to raise a PR for this refactor once https://github.com/dbt-labs/docs.getdbt.com/pull/7681 is merged.

luna-bianca:
Closes: https://github.com/dbt-labs/docs.getdbt.com/issues/7683

Restructured Examples section and added a subheading title for each example so they're visible in the navigation pane

## Checklist

---

- https://docs-getdbt-com-git-issue-7683-dbt-labs.vercel.app/reference/global-configs/resource-type

### DIFF
--- a/website/docs/reference/global-configs/resource-type.md
+++ b/website/docs/reference/global-configs/resource-type.md
@@ -37,6 +37,8 @@ Instead of targeting specific resources, use the `--resource-type` or `--exclude
 
 See the following sample commands for including or excluding resource types. Note that the `--exclude-resource-type` flag is only available in dbt version 1.8 and higher.
 
+<Expandable alt_header="Include resource types">
+
 ### Include multiple resource types
 
 Use the following command to include multiple resource types such as data tests and models in your build process:
@@ -102,6 +104,9 @@ dbt test --resource-type test
 
 </VersionBlock>
 
+</Expandable>
+<Expandable alt_header="Exclude resource types">
+
 ### Exclude multiple resource types
 
 Use the following command to exclude multiple resource types such as data tests and models from your build process:
@@ -141,3 +146,4 @@ dbt test --exclude-resource-type unit_test
 </File>
 
 </VersionBlock>
+</Expandable>
--- a/website/docs/reference/global-configs/resource-type.md
+++ b/website/docs/reference/global-configs/resource-type.md
@@ -52,7 +52,7 @@ dbt build --resource-type snapshot
 
 ### Include all saved queries
 
-In this example, run the following command to include _all_ saved queries with the `--resource-type` flag:
+Use the following command to include _all_ saved queries with the `--resource-type` flag:
 
 <File name='Usage'>
 
@@ -64,7 +64,7 @@ dbt build --resource-type saved_query
 
 ### Include all data tests
 
-In this example, use the following command to include _all_ data tests in your build process:
+Use the following command to include _all_ data tests in your build process:
 
 <File name='Usage'>
 
@@ -76,7 +76,7 @@ dbt build --resource-type test
 
 ### Include multiple resource types
 
-In this example, use the following command to include multiple resource types such as data tests and models in your build process:
+Use the following command to include multiple resource types such as data tests and models in your build process:
 
 <File name='Usage'>
 
@@ -88,7 +88,7 @@ dbt build --resource-type test model
 
 ### Exclude multiple resource types
 
-In this example, use the following command to exclude multiple resource types such as data tests and models from your build process:
+Use the following command to exclude multiple resource types such as data tests and models from your build process:
 
 <File name='Usage'>
 
@@ -100,7 +100,7 @@ dbt build --exclude-resource-type test model
 
 ### Exclude all unit tests
 
-In this example, use the following command to exclude _all_ unit tests from your dbt build process.
+Use the following command to exclude _all_ unit tests from your dbt build process.
 
 <File name='Usage'>
 
@@ -114,7 +114,7 @@ dbt build --exclude-resource-type unit_test
 
 ### Include all data tests during testing
 
-In this example, use the following command to include _all_ data tests when running tests:
+Use the following command to include _all_ data tests when running tests:
 
 <File name='Usage'>
 
@@ -126,7 +126,7 @@ dbt test --resource-type test
 
 ### Exclude all unit tests during testing
 
-In this example, use the following command to exclude _all_ unit tests when running tests:
+Use the following command to exclude _all_ unit tests when running tests:
 
 <File name='Usage'>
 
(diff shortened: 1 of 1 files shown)


dbeatty10 (approved):
Thank you for opening this @luna-bianca !

Here's a couple non-blocking reflections that would be great to address if possible:
- Phrases like "include _all_ snapshots in your dbt build process" don't properly convey that this example actually shows "_only_ include snapshots in your dbt build process"
    - My commentary: `--resource-type` essentially operates like a (non-security related) [allow list](https://en.wikipedia.org/wiki/Whitelist) whereas `--exclude-resource-type` operates like a [deny list](https://en.wikipedia.org/wiki/Blacklist_(computing))
- Since there's a lot of examples (good!) it's also hard to find the example you are looking for (not good)
    - Maybe we group all the exclude examples together and all the include examples together? Or at least lead with the "multiple" examples since it is actually the general case (rather than just being a special case)?

nataliefiann (approved):
Hiya @luna-bianca 

Thanks for creating this PR! Loving the expandables! 

Kind Regards
Natalie

dbeatty10 (approved):
I still think there's some important nuance missing in our title headings and explanations.

Namely, the `--resource-type` flag will _only_ include the resource types listed. Put another way, it will exclude all resource types except for those listed.

Conversely, the `--exclude-resource-type` flag will include _all_ resource types except the ones listed.

It's a bit nerdy 🤓, but here's an option for docs we could add that covers these insights:

### Positive vs. Negative Resource‑Type Filters
`--resource-type` is a **positive filter**: dbt will run **only** the resource types you list, implicitly skipping every other type.  
`--exclude-resource-type` is a **negative filter**: dbt starts with the full catalogue of resource types and then omits the ones you name, so everything *except* those runs.

You can supply **both** flags; dbt first applies the positive filter and then removes the negatives, the same as the *[relative complement](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement)* in set theory. Because the list of dbt resource types is *mutually‑exclusive and collectively‑exhaustive* ([MECE](https://en.wikipedia.org/wiki/MECE_principle)), any `--resource-type` selection can be re‑expressed with `--exclude-resource-type`, and vice‑versa.

website/docs/reference/global-configs/resource-type.md
```diff
@@ -31,91 +30,132 @@ The available resource types are:
...
+## Positive vs negative filters
+
+- `--resource-type` is a positive filter &mdash; dbt only runs the resource types selected in the command, implicitly skipping every other type.
+- `--exclude-resource-type` is a negative filter &mdash; dbt starts with the full catalog of resource types and then omits the types selected in the command. dbt runs everything _except_ those resource types. 
+
+You can use both flags in a command; dbt first applies the positive filter (`--resource-type`) and then removes the types listed in the negative filter (`--exclude-resource-type`). For example:
 
-Instead of targeting specific resources, use the `--resource-type` or `--exclude-resource-type` flags to target all resources of a certain type: `dbt build --resource-type RESOURCE_TYPE` replacing `RESOURCE_TYPE` with the resource type you want to include.
+```text
+dbt build --resource-type model test snapshot --exclude-resource-type snapshot
```
luna-bianca, line 41: Added a sample here, let me know if this is correct. I tested this and it worked.

website/docs/reference/global-configs/resource-type.md
```diff
@@ -31,91 +30,132 @@ The available resource types are:
...
+```
 
-- For example, use the following command to include _all_ snapshots from your dbt build process:
+Note that the list of dbt resource types is mutually exclusive and collectively exhaustive (MECE). This means that any `--resource-type` selection can also be achieved by excluding the other resource types using `--exclude-resource-type`, and vice versa.
```
luna-bianca, line 44: Can you please confirm if this revision is technically correct?

luna-bianca (commented):
Hi @dbeatty10 , thanks for the detailed writeup! I added a new section for this [here](https://docs-getdbt-com-git-issue-7683-dbt-labs.vercel.app/reference/global-configs/resource-type#positive-vs-negative-filters). I added and revised some things as well, please see the comments below.

website/docs/reference/global-configs/resource-type.md
dbeatty10: Yep, looks technically correct @luna-bianca 👍 -- see below for details.

Starting with a project with all the resource types listed [here](https://docs.getdbt.com/reference/global-configs/resource-type) (and assuming `jq` is installed), see below for some tests I did with various combinations using dbt-core 1.10:

See that we can get the same output with either `--resource-type` or `--exclude-resource-type`:

```
$ dbt list -q --output-keys resource_type --output json --resource-type analysis exposure metric model saved_query | jq -r '.resource_type' | sort | uniq 
analysis
exposure
metric
model
saved_query

$ dbt-core list -q --output-keys resource_type --output json --exclude-resource-type seed semantic_model snapshot source test unit_test --resource-type all | jq -r '.resource_type' | sort | uniq
analysis
exposure
metric
model
saved_query
```

See the list of all resource types in this project (which should be all of them):

```
$ dbt list -q --output-keys resource_type --output json --resource-type all | jq -r '.resource_type' | sort | uniq 
analysis
exposure
metric
model
saved_query
seed
semantic_model
snapshot
source
test
unit_test
```

Select just one resource type:

```
$ dbt-core list -q --output-keys resource_type --output json --resource-type analysis | jq -r '.resource_type' | sort | uniq
analysis
```

Exclude just that resource type to get the relative complement:

```
$ dbt-core list -q --output-keys resource_type --output json --exclude-resource-type analysis | jq -r '.resource_type' | sort | uniq
exposure
metric
model
saved_query
seed
semantic_model
snapshot
source
test
unit_test
```

Combine inclusion and exclusion flags in the same command:

```
$ dbt-core list -q --output-keys resource_type --output json --resource-type analysis metric model --exclude-resource-type analysis | jq -r '.resource_type' | sort | uniq 
metric
model

$ dbt-core list -q --output-keys resource_type --output json --resource-type analysis metric model --exclude-resource-type metric model | jq -r '.resource_type' | sort | uniq
analysis
```

nataliefiann (approved):

PR #7882, opened 2025-09-08
IDE Extension nav CTA

JKarlavige:

[Notion task](https://www.notion.so/dbtlabs/Add-CTA-to-docs-nav-268bb38ebda781779a33c5e03ea09469?source=copy_link)

- Adds "Install dbt IDE extension" CTA to primary navigation
- Adjusts nav styles to:
  - Hide full searchbar styles until very wide desktop
  - Better fit and wrap all content until mobile nav styles kick in

## Preview
https://docs-getdbt-com-git-ide-nav-cta-dbt-labs.vercel.app/

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -207,6 +207,7 @@ var siteSettings = {
           label: "Install VS Code extension",
           position: "right",
           to: "/docs/install-dbt-extension",
+          id: "nav-install-vs-code-extension",
           className: "nav-install-dbt-extension",
         },
       ],
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -203,6 +203,12 @@ var siteSettings = {
             },
           ],
         },
+        {
+          label: "Install dbt IDE extension",
+          position: "right",
+          to: "/docs/install-dbt-extension",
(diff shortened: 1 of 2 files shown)


breezyfasano (changes requested):
Great job squeezing all of this in, oof. Just have one point of feedback:
From breakpoint size `997px` to `1155px`, the navbar and logo are misaligned horizontally.
image

 I believe the `min-width: auto;` here is the main culprit
 
image

JKarlavige:
> Great job squeezing all of this in, oof. Just have one point of feedback: From breakpoint size `997px` to `1155px`, the navbar and logo are misaligned horizontally. 
> I believe the `min-width: auto;` here is the main culprit

This was actually intentional to prevent overlap between the nav items and the logo. The same overlap occurs on the live site when shrinking down the browser width. I tried to fit as much as I could into keeping it on 1 line, but run out of room between 996px and 1154px and that's where i have them stacking briefly.

It's definitely not ideal, but we straight up don't have enough room in the nav to fit everything everyone wants in there 😅 
image

breezyfasano (approved):
Ahhh I got you, that makes a lot of sense! Okay, lgtm in that case 👍

PR #7932, opened 2025-09-22
Adjust EOP cta

breezyfasano:
- Fixing the End of Page CTA: CTA on "Request your demo" is linking to `/signup`, but should be `/contact`


mirnawong1 (approved):

mirnawong1:
thanks so much for that speedy update @breezyfasano !!

PR #7976, opened 2025-09-30
Fix change_tracking and update catalog_name format

rahulgupta8:
Updated change_tracking value to a boolean and corrected catalog_name format.

Updating some of the sample config values for snowflake iceberg integration.
- change_tracking: When set to 'True' causes a compilation error and its not easy to track it from the dbt logs why this is being caused. It turns out the value needs to be just a boolean and thus should be set to `True` without the quotes.
- In the model config block, catalog_name is set to catalog_horizon, but when it is trying to compile dbt is searching for a variable catalog_horizon and not the string value `catalog_horizon`, so the correct input to this config attribute is a string value of `catalog_horizon` with the quotes.

## Checklist

### DIFF
--- a/website/docs/docs/mesh/iceberg/snowflake-iceberg-support.md
+++ b/website/docs/docs/mesh/iceberg/snowflake-iceberg-support.md
@@ -205,7 +205,7 @@ catalogs:
         table_format: iceberg
         catalog_type: built_in
         adapter_properties:
-          change_tracking: 'True'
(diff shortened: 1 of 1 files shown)


runleonarun:
Thanks so much for opening this @rahulgupta8 We're just verifying and will merge this fix ASAP.

runleonarun (approved):
I verified in the Snowflakes docs so approving!

PR #7999, opened 2025-10-05
Fix typos in webhook JSON payload examples

ubaidrmn:

- The JSON payload examples for webhooks in the [docs](https://docs.getdbt.com/docs/deploy/webhooks#examples-of-json-payloads) use the key `"webhooksID"`. What DBT actually sends in the payload is `"webhookId"`. This maybe a typo, or maybe the key was changed and the documentation wasn't updated.

- Also updated the value for `"runReason"`field to what DBT sends.

This is a screenshot of a webhook payload I captured using webhooks.site for verifying the keys:

image

### DIFF
--- a/website/docs/docs/deploy/webhooks.md
+++ b/website/docs/docs/deploy/webhooks.md
@@ -97,7 +97,7 @@ An example of a webhook payload for a run that's started:
 ```json
 {
   "accountId": 1,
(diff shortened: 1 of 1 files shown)


mirnawong1 (approved):
hey @ubaidrmn , great catch! thanks so much for opening up a pr to fix this up!

PR #8007, opened 2025-10-08
update flags in flag table

mirnawong1:
Lee and Emma raised that some search items don't appear when searching in algolia. this is because content inside versionblocks aren't picked up by the internal search. This pr removes version blocks to the two higlighted issues and adds a 'version callout' so the content is picked up in search.

Additionally, Emma raised that a user wanted to know the names of the environment variables they should avoid (to avoid unintended behaviors).  added flags/env variables relevant for users and not deprecated

internal slack: https://dbt-labs.slack.com/archives/C02NCQ9483C/p1759835283540029?thread_ts=1759835279.988309&cid=C02NCQ9483C

source: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/cli/params.py

---

- https://docs-getdbt-com-git-fix-searched-items-dbt-labs.vercel.app/reference/global-configs/about-global-configs
- https://docs-getdbt-com-git-fix-searched-items-dbt-labs.vercel.app/reference/resource-configs/databricks-configs


website/docs/reference/resource-configs/databricks-configs.md
```diff
@@ -38,11 +37,8 @@ dbt-databricks v1.9 adds support for the `table_format: iceberg` config. Try it
 In dbt-databricks v1.10, there are several new model configurations options gated behind the `use_materialization_v2` flag.
 For details, see the [documentation of Databricks behavior flags](/reference/global-configs/databricks-changes).
 
-</VersionBlock>
-
-<VersionBlock firstVersion="1.9">
-
 ### Python submission methods
+_Available in versions 1.9 or higher_
```
luna-bianca, line 41: Non-blocker comment but I wanted to ask: Instead of adding this note in several places, should we just remove the mention of version 1.9 altogether (VersionBlock or note)? The versions in our docs site start from 1.9 and users won't have access to docs 1.8 and below. 

I see that there is a "_Available in versions 1.10 or higher_" in line 140 -- maybe we can keep that one?

luna-bianca (approved):
Hi @mirnawong1 , I have a query about the addition _Available in versions 1.9 or higher_, please check. Thanks!

website/docs/reference/resource-configs/databricks-configs.md
mirnawong1: that's a great question @luna-bianca ! i would usually do this but looking at a [quick data search](https://app.hex.tech/fishtown/thread/0199dd92-e186-700d-aaf9-b5d0005087c7), we do have a good chunk of users/accounts that are on older versions still and i think it might be important to let them know of this in case they're interested.  

let me know if you have strong opinions on this though!

PR #8124, opened 2025-11-03
Remove requirement for `unique_key` on `merge` strategy with bigquery

fornwall:
The bigquery adapter seems to work fine without a `unique_key` when using the `merge` strategy.


mirnawong1 (approved):
hey @fornwall yes great catch! also discussed here: https://getdbt.slack.com/archives/CBSQTAPLG/p1762788156202979

PR #8220, opened 2025-11-25
add updates to dbt show

### LINKED ISSUE #4256: Updates for `dbt show`
### Contributions

https://docs.getdbt.com/reference/commands/show

Original content from @dataders  copy-pasted from [here](https://gist.github.com/dataders/0977e4948220d56ecc1fea941e5d6fab) and then lightly edited.

---

### New `dbt show` feature in v1.5
`dbt show` shipped in `1.5.0` with a `--limit` flag that, when provided, would limit the number of results that dbt would grab to display.

### One little problem
As described in https://github.com/dbt-labs/dbt-core/issues/8496, most adapters load the entire model into memory while running `dbt show` which can be prohibitively slow for large data sets.

It does not modify the original query, which means that even if you provide `--limit 5`, the command will not complete until the entire underying query is complete which can take a long time if the query's result set is large. This is especially evident becaues `dbt show` is now also used for dbt Cloud IDE's "preview" button.

### A fix for the problem

The fix was https://github.com/dbt-labs/dbt-core/pull/8641, which extends the implementation of the `--limit` flag to wrap the underlying query into a CTE and append a `LIMIT {limit}` clause.

This fix was backported to 1.5 and 1.6, so that it applies to all versions of dbt that support `dbt show`.

### Changes for adapter maintainers

No change is needed if the adapter's data platform supports ANSI SQL `LIMIT` clauses. Any changes can be made in a dispatched version of `get_limit_subquery_sql()`, see [`default__get_limit_subquery()`](https://github.com/dbt-labs/dbt-core/blob/53845d0277be2b0ab347ac07b84bc86363157c54/core/dbt/include/global_project/macros/adapters/show.sql#L16-L22). This was merged into `1.7.0`, but also backported to `1.5` and `1.6`, so if a change is needed, it will likely also need to be backported and patched accordingly.

`dbt.tests.adapter.dbt_show.test_dbt_show` also contains new tests to ensure the new behavior functions properly:
- `BaseShowSqlHeader`
- `BaseShowLimit`

### Additional information

_No response_

### LINKED ISSUE #8216: Guarantees of `dbt show`
### Contributions

[docs site](https://docs.getdbt.com/reference/commands/show)
[docs repo](https://github.com/dbt-labs/docs.getdbt.com/blob/3fe547552f21f5c7d865269ce9a33c865768a992/website/docs/reference/commands/show.md?plain=1)

On the [page about the show command](https://docs.getdbt.com/reference/commands/show), we state that
> The results of the preview query are not materialized in the data warehouse, or stored in any dbt file ([permalink](https://github.com/dbt-labs/docs.getdbt.com/blob/3fe547552f21f5c7d865269ce9a33c865768a992/website/docs/reference/commands/show.md?plain=1#L16))

This _is true_ when using `dbt show -s "my_model_name"`, but this is not necessarily true when using the `--inline` command. This was identified by a community created [core issue](https://github.com/dbt-labs/dbt-core/issues/11981). [In my response to that issue](https://github.com/dbt-labs/dbt-core/issues/11981#issuecomment-3560902725), I go into: why inline works that way, that we cannot lock it down, and what a work around could be. The TLDR is:
* `--inline` works on arbitrary SQL
* It is impossible for us to guarantee that the user supplied arbitrary SQL won't make changes in the data warehouse
* If the user needs a guarantee that their `--inline` show doesn't make modifications, they need to use a profile with a role that only has "read only" permissions
    * this either means a user have a second profile that they use when using dbt show, i.e. `dbt show --inline ... --profile "my-read-only-profile"`
    * or if the user should only have read only perms full stop, then that should be reflected in the role they have
    * **note:** permissions a role has are typically configured in the data warehouse, we don't have control over what permissions a role has

### Additional information

_No response_

mirnawong1:
adds updates to dbt show command wrt the way it works and how it works against the data warehouse (e.g. `--inline`)

Resolves #8216
Resolves #4256

https://docs-getdbt-com-git-update-show-cmd-dbt-labs.vercel.app/reference/commands/show

---

- https://docs-getdbt-com-git-update-show-cmd-dbt-labs.vercel.app/reference/commands/show

### DIFF
--- a/website/docs/reference/commands/show.md
+++ b/website/docs/reference/commands/show.md
@@ -13,14 +13,17 @@ Use `dbt show` to:
 
 ## How it works
 
-By default, `dbt show` will display the first 5 rows from the query result. This can be customized by passing the flag `--limit n`, where `n` is the number of rows to display.
+By default, `dbt show` will display the first 5 rows from the query result. This can be customized by passing the `limit` or the `inline` flags , where `n` is the number of rows to display.
 
-- The `--limit` flag modifies the underlying SQL and not just the number of rows displayed. This is so that only `n` rows are retrieved from the data warehouse. 
+If previewing a model, dbt will always compile and run the compiled query from source. It will not select from the already-materialized database relation, even if you've just run the model. (We may support that in the future; if you're interested, upvote or comment on [dbt-core#7391](https://github.com/dbt-labs/dbt-core/issues/7391).)
+
+#### `limit` flag
+- The `--limit` flag modifies the underlying SQL and not just the number of rows displayed. By using the `--limit n` flag, it means `n` is the number of rows to display and retrieved from the data warehouse. 
 - This means dbt wraps your model's query in a subquery or CTE and applies a SQL `limit n` clause so that your data warehouse only processes and returns that number of rows, making it significantly faster for large datasets.
-- The results of the preview query are only included in dbt's logs and displayed in the terminal and aren't materialized in the data warehouse or stored in any dbt file, except if you use `dbt show --inline`.
-  - The `--inline` flags enables you to run ad-hoc SQL, which means dbt can't ensure the query doesn't modify the data warehouse. To ensure no changes are made, use a profile or role with read-only permissions, which are managed directly in your data warehouse. For example: `dbt show --inline "select * from my_table" --profile my-read-only-profile`.
-- If previewing a model, dbt will always compile and run the compiled query from source. It will not select from the already-materialized database relation, even if you've just run the model. (We may support that in the future; if you're interested, upvote or comment on [dbt-core#7391](https://github.com/dbt-labs/dbt-core/issues/7391).)
 
+#### `inline` flag
+- The results of the preview query are only included in dbt's logs and displayed in the terminal and aren't materialized in the data warehouse or stored in any dbt file, except if you use `dbt show --inline`.
+- The `--inline` flags enables you to run ad-hoc SQL, which means dbt can't ensure the query doesn't modify the data warehouse. To ensure no changes are made, use a profile or role with read-only permissions, which are managed directly in your data warehouse. For example: `dbt show --inline "select * from my_table" --profile my-read-only-profile`.
 
 ## Example
 
--- a/website/docs/reference/commands/show.md
+++ b/website/docs/reference/commands/show.md
@@ -11,11 +11,18 @@ Use `dbt show` to:
 - Run that query against the data warehouse
 - Preview the results in the terminal
 
+## How it works
+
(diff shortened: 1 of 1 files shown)


website/docs/reference/commands/show.md
```diff
@@ -11,11 +11,18 @@ Use `dbt show` to:
 - Run that query against the data warehouse
 - Preview the results in the terminal
 
+## How it works
+
 By default, `dbt show` will display the first 5 rows from the query result. This can be customized by passing the flag `--limit n`, where `n` is the number of rows to display.
 
-The results of the preview query are not materialized in the data warehouse, or stored in any dbt file. They are only included in dbt's logs and displayed in the terminal. Note also that, if previewing a model, dbt will always compile and run the compiled query from source. It will not select from the already-materialized database relation, even if you've just run the model. (We may support that in the future; if you're interested, upvote or comment on [dbt-core#7391](https://github.com/dbt-labs/dbt-core/issues/7391).)
+- The `--limit` flag modifies the underlying SQL and not just the number of rows displayed. This is so that only `n` rows are retrieved from the data warehouse.
```
matthewshaver: It might be beneficial to make the flags H3 headers
### The `limit` flag
### The `inline` flag

Then explain them and give individual examples. This will be helpful if we need to link to them from elsewhere in the docs

website/docs/reference/commands/show.md
```diff
@@ -11,11 +11,18 @@ Use `dbt show` to:
...
+- This means dbt wraps your model's query in a subquery or CTE and applies a SQL `limit n` clause so that your data warehouse only processes and returns that number of rows, making it significantly faster for large datasets.
+- The results of the preview query are only included in dbt's logs and displayed in the terminal and aren't materialized in the data warehouse or stored in any dbt file, except if you use `dbt show --inline`.
+  - The `--inline` flags enables you to run ad-hoc SQL, which means dbt can't ensure the query doesn't modify the data warehouse. To ensure no changes are made, use a profile or role with read-only permissions, which are managed directly in your data warehouse. For example: `dbt show --inline "select * from my_table" --profile my-read-only-profile`.
+- If previewing a model, dbt will always compile and run the compiled query from source. It will not select from the already-materialized database relation, even if you've just run the model. (We may support that in the future; if you're interested, upvote or comment on [dbt-core#7391](https://github.com/dbt-labs/dbt-core/issues/7391).)
```
matthewshaver:
```suggestion
- If previewing a model, dbt will always compile and run the compiled query from source. It will not select from the already-materialized database relation, even if you've just run the model (We may support that in the future; if you're interested, upvote or comment on [dbt-core#7391](https://github.com/dbt-labs/dbt-core/issues/7391)).
```

matthewshaver (approved):

QMalcolm:
Thank you for doing this ❤️

PR #8307, opened 2025-12-12
Update full_refresh.md

mirnawong1:
clarify full_config per feedback

---

- https://docs-getdbt-com-git-mirnawong1-patch-31-dbt-labs.vercel.app/reference/resource-configs/full_refresh

### DIFF
--- a/website/docs/reference/resource-configs/full_refresh.md
+++ b/website/docs/reference/resource-configs/full_refresh.md
@@ -87,7 +87,17 @@ You can set the `full_refresh` config in the `dbt_project.yml` file or in a reso
 
 ## Recommendation
 - Set `full_refresh: false` for models of especially large datasets, which you would _never_ want dbt to fully drop and recreate.
-- You cannot override an existing `full_refresh` config. To change its behavior in certain circumstances, remove the config logic or update it using variables so the behavior can be overridden when needed.
+- You cannot override an existing `full_refresh` config. To change its behavior in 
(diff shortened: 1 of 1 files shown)


website/docs/reference/resource-configs/full_refresh.md
```diff
@@ -87,7 +87,7 @@ You can set the `full_refresh` config in the `dbt_project.yml` file or in a reso
 
 ## Recommendation
 - Set `full_refresh: false` for models of especially large datasets, which you would _never_ want dbt to fully drop and recreate.
-- For existing `full_refresh` configs, you can't override their behavior. To change their behavior in certain situations, remove the config or update the config logic using variables.
+- You cannot override an existing `full_refresh` config. To change its behavior in certain circumstances, remove the config logic or update it using variables so the behavior can be overridden when needed.
```
runleonarun: I think an example would go along way here. Here's what I mean but I haven't tested this:

```jinja
{{ config(
    materialized = 'incremental',
    full_refresh = var("force_full_refresh", false)
) }}

```
By default, full_refresh stays false. But you can explicitly override the behavior using the variable:

```
dbt run --vars '{"force_full_refresh": true}'

```

runleonarun (approved):
Approved with the suggestion of adding an example!

website/docs/reference/resource-configs/full_refresh.md
mirnawong1: tested and add example

PR #8379, opened 2026-01-06
Fix - Code block highlighting

john-rock:

This PR includes a small set of targeted changes aimed at eliminating intermittent unstyled code rendering. While I wasn’t able to reliably reproduce the issue locally or across browsers, these updates address likely root causes with minimal risk and surface area.

### Summary of changes

1. **Upgrade `prism-react-renderer`**  
   - Bumped from `^2.3.1` → `^2.4.1` to pick up minor fixes and improvements related to token rendering.

2. **Tighten CSS scoping for code elements**  
   - Updated the `code` selector to `:not(pre) > code` to prevent inline code styles from leaking into fenced code blocks  
   - Changed `a code` to `a > code` for more explicit targeting  
   - Ensured token lines always render as block-level elements to prevent occasional single-line “flash” during render

3. **Remove homepage banner animation**  
   - Removed an animation that caused an object module flash on load  
   - While not directly tied to the code rendering issue, this animation is no longer needed and removing it reduces visual instability on page load

## Preview Links
https://docs-getdbt-com-git-fix-code-block-syntax-dbt-labs.vercel.app/reference/resource-properties/freshness

https://docs-getdbt-com-git-fix-code-block-syntax-dbt-labs.vercel.app/best-practices/how-we-mesh/mesh-6-coordinate-versions

## Testing
- Verify code blocks render correctly on initial page load
- Verify code blocks don't flash during client-side navigation
- Verify inline code styling still works correctly
- Verify homepage loads without announcement  banner `[Object module]` flash

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -35,7 +35,7 @@
         "mobx": "^6.3.9",
         "node-polyfill-webpack-plugin": "^3.0.0",
         "papaparse": "^5.3.2",
-        "prism-react-renderer": "^2.3.1",
+        "prism-react-renderer": "^2.4.1",
         "query-string": "^8.1.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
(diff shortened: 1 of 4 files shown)


JKarlavige (approved):
:shipit:

PR #8394, opened 2026-01-08
Update profiles.yml

matthewshaver:

Update the `profiles.yml` docs:
- Give more context on the file
- Where to place it and why
- Add env_var
- Enhance the example

## Checklist


website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+- **Target definitions** &mdash; Define different environments (dev, staging, prod) within a single profile.
+- **Default target** &mdash; Set which environment to use by default.
+- **Execution parameters** &mdash; Thread count, timeouts, and retry settings.
+- **Credential separation** &mdash; Keep sensitive information out of version control.
 
-<File name='profiles.yml'>
+The `profile` field in [`dbt_project.yml`](/reference/dbt_project.yml) references a profile name defined in `profiles.yml`.
+
+## Location of profiles.yml
+
```
mirnawong1, line 20: so great to add this!

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+dbt searches for `profiles.yml` location in a specific order:
```
mirnawong1: for me, this sentence implies all three are needed and the priority stacking.
i think you mean to sayit uses the first one it finds?

```suggestion
dbt searches for `profiles.yml` location in the following order and uses the first file it finds:
```

website/snippets/_profiles.yml.md
mirnawong1, line 20: sugg making it super clear, i've had new users get confused about this in the past
```suggestion
You only need to use one `profiles.yml` file.
```

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+2. Project root directory &mdash; Project-specific credentials.
+3. `~/.dbt/profiles.yml` (Recommended location) &mdash; Shared across all projects.
+
+`~/.dbt/profiles.yml` is the recommended location for the following reasons:
+
+- **Security** &mdash; Keeps credentials out of project directories and version control.
+- **Reusability** &mdash; A single file for all dbt projects on the machine.
+- **Separation** &mdash; Connection details don't travel with project code.
+
+### When to use project root
```
mirnawong1: non blocker but sugg to downgrade this to expandable _under_ the project root directory bullet to reduce the right toc

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+
+Place your `profiles.yml` file in the project root directory for:
+
+- Self-contained demo or tutorial projects.
+- Docker containers with baked-in credentials.
+- CI/CD pipelines with environment-specific configs.
+
+## Create and configure the `profiles.yml` file
+
+The easiest way to create and configure a `profiles.yml` file is to execute `dbt init` after you've installed dbt on your machine. This takes you through the process of configuring an adapter and places the file into the recommended `~/.dbt/` location.
```
mirnawong1: do we want to mention that `dbt init` overwrites existing profiles only if confirmed?

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+
+You can also manually create the file and add it to the proper location. To configure an adapter manually, copy and paste the fields from the [adapter setup instructions](/docs/about-dbt-install) along with the appropriate values for each. 
+
+### Example configuration
```
mirnawong1, line 51: love that we have a detailed example here! 

i'm a little confused with the fields though. because there are no universal fields across adapters (although there are accepted values). i think we should avoid example placeholders like `schema_field` because otherwise the user might think this is what they need to add or might ask themselves 'schema_field or schema, which is correct?'.  

we should clearly signal “these keys vary by adapter” inthe doc.

maybe we should select a warehouse and use it in the example. e.g:

This example uses Snowflake as an adapter. Required fields and field names in the `profiles.yml` vary by adapter. For a complete and accurate field names, refer to the adapter-specific setup instructions for the complete list.

```yaml
my_project_profile:
  target: dev
  outputs:
    dev:
      type: snowflake  # Example adapter
      account: account
      user: user
      password: password
      database: database
      schema: schema
      warehouse: warehouse
      threads: 4
```

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+<File name='~/.dbt/profiles.yml'>
 
 ```yml
+my_project_profile:  # Profile name (matches dbt_project.yml)
+  target: dev  # Default target to use
+  outputs:
+    dev: # Development environment
+      type: <adapter_type> # Required: snowflake, bigquery, databricks, redshift, postgres, etc
+      # Connection identifiers (adapter-specific)
+      <account_or_host>: '<value>'
```
mirnawong1: also i'm not a fan of user `<>` for the placeholders as we had users assume they had to add this in their code (not everyone knows it implies an example field). 


some of the placeholders have single quotes around them - i strongly think we should remove them as the reader might think the quotes are required


if you feel strongly that we should use the angular brackets, then can we have a sentence around it so users know? e.g.

> Angle brackets (`< >`) indicate placeholders and need to be replaced with your real values. You don't need to include the brackets or quotes in your actual config.

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+      user: '{{ env_var("ADAPTER_USER") }}'
+      password: '{{ env_var("ADAPTER_PASSWORD") }}'
+      database: '{{ env_var("ADAPTER_DATABASE") }}'
+      schema: '{{ env_var("ADAPTER_SCHEMA") }}'
+      warehouse: '{{ env_var("ADAPTER_WAREHOUSE") }}'
+      role: '{{ env_var("ADAPTER_ROLE") }}'
+      threads: 4
 ```
 
 </File>
```
mirnawong1, line 116: we generally lean towards using double quotes. sugg making it clear that both quotes are accepted but lean into double quotes in your above example so its consistent across the docs
```suggestion
If you're using `env_var` in your `profiles.yml` file, make sure to quote the entire Jinja string. dbt accepts either single or double quotes: user: 
- ✅`"{{ env_var("ADAPTER_ACCOUNT") }}"` 
- ✅`'{{ env_var("ADAPTER_ACCOUNT") }}'`
- ❌ {{ env_var("ADAPTER_ACCOUNT") }}

```

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,106 @@
...
+
+- Self-contained demo or tutorial projects.
+- Docker containers with baked-in credentials.
+- CI/CD pipelines with environment-specific configs.
+
+## Create and configure the `profiles.yml` file
+
+The easiest way to create and configure a `profiles.yml` file is to execute `dbt init` after you've installed dbt on your machine. This takes you through the process of configuring an adapter and places the file into the recommended `~/.dbt/` location.
+
+You can also manually create the file and add it to the proper location. To configure an adapter manually, copy and paste the fields from the [adapter setup instructions](/docs/about-dbt-install) along with the appropriate values for each.
```
mirnawong1: the adapter setup instructions link takes you to the about instal page and not the adapter setup, making it more harder for the user to know where to go. 

becuse we currently have core + adapter and fusion+adapter docs, we shoudl link to those sections and be clearer about them.

website/snippets/_profiles.yml.md
matthewshaver: The reason I took this approach was to avoid singling out any one adapter or giving any indication that fields like that are universal. The individual pages contain the specific configs they can use and we call that out in this article. I think we should leave it generic here to avoid showing preference or causing confusion

website/snippets/_profiles.yml.md
matthewshaver: I can change it to all caps like we use elsewhere but we wanted to indicate that these are not static fields, they change depending on the adapter, and then point users to the individual adapter docs to get more info

website/snippets/_profiles.yml.md
matthewshaver: I'll add the double quotes here so it's consistent with the env_var page we link to

website/snippets/_profiles.yml.md
mirnawong1: i def understand that, but we currently use r[edshift](https://docs.getdbt.com/reference/dbt-jinja-functions/profiles-yml-context) and other adapter across teh docs so it's not really an anti pattern. 

the main point from my end is, if we leave it with fields like `<database_field>` or `<schema_field>` - then i'd be really worried users would take it literal and cause them unnecessary debugging/support ticket creation. 

if you really want to keep it generic, can we clearly mark placeholders as placeholders or use more accepted values?

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,113 @@
...
+2. Project root directory &mdash; Project-specific credentials.
+3. `~/.dbt/profiles.yml` (Recommended location) &mdash; Shared across all projects.
+
+`~/.dbt/profiles.yml` is the recommended location for the following reasons:
+
+- **Security** &mdash; Keeps credentials out of project directories and version control.
+- **Reusability** &mdash; A single file for all dbt projects on the machine.
+- **Separation** &mdash; Connection details don't travel with project code.
+
+#### When should I use project root?
```
matthewshaver, line 35: Expandable looked a little out of place and was easily missable, so changed to an H4 header to avoid side TOC

website/snippets/_profiles.yml.md
matthewshaver: I gotcha - I'll split the difference and add the specific example to the bottom and keep the generic up top while altering it to remove the brackets!

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,119 @@
...
+
+- Self-contained demo or tutorial projects.
+- Docker containers with baked-in credentials.
+- CI/CD pipelines with environment-specific configs.
+
+## Create and configure the `profiles.yml` file
+
+The easiest way to create and configure a `profiles.yml` file is to execute `dbt init` after you've installed dbt on your machine. This takes you through the process of configuring an adapter and places the file into the recommended `~/.dbt/` location. 
+
+If an existing profiles.yml file exists, running dbt init will prompt the user to amend or overwrite it. If the user selects the existing adapter for configuration, dbt will automatically populate the existing values.
```
mirnawong1: recommend keeping it first person so its consistent with the rest of the section
```suggestion
If your project has an existing `profiles.yml` file, running `dbt init` will prompt you to amend or overwrite it. If you select the existing adapter for configuration, dbt will automatically populate the existing values.
```

website/snippets/_profiles.yml.md
```diff
@@ -1,40 +1,119 @@
...
+<File name='~/.dbt/profiles.yml'>
 
 ```yml
+my_project_profile:  # Profile name (matches dbt_project.yml)
+  target: dev  # Default target to use
+  outputs:
+    dev: # Development environment
+      type: adapter_type # Required: snowflake, bigquery, databricks, redshift, postgres, etc
+      # Connection identifiers (placeholder examples, see adapter-specific pages for supported configs)
+      account: 'abc123'
```
mirnawong1: this looks awesome! i love the combo - my only personal nit is that i still don't think the placeholder value should include single quotes _just_ so the user doesn't think it's needed or confuse them, esp since the bottom example has no quotes. but just my sugg!

mirnawong1 (approved):

PR #8414, opened 2026-01-12
update pages user feedback

mirnawong1:
updating the following pages due to user feedback from the thumbs up/down feedback box:

- /docs/about-dbt-install.md : It skips from "Chose your development experience" to "After installing your local development experience" with no links or instructions for how to install the local development experience after choosing, or no content at all (or
- /reference/resource-configs/hive-configs.md :  add /reference/resource-configs/hive-configs to the sidebar.js in the reference section

---

- https://docs-getdbt-com-git-fix-clarify-install-dbt-labs.vercel.app/docs/about-dbt-install


website/docs/docs/about-dbt-install.md
```diff
@@ -9,13 +9,21 @@ pagination_prev: null
 
 <Constant name="dbt" /> enables data teams to transform data using analytics engineering best practices. Choose your local development experience from these tools:
 
-* Local command line interface (CLI) tools leveraging the speed and scale of the <Constant name="fusion_engine" /> or using <Constant name="core" />.
-* VS Code or Cursor with the dbt extension
+#### Local command line interface (CLI)
```
luna-bianca: Is there a reason these sections use #### headings before the ### Getting started section?
Wondering if this is intentional (for styling/TOC reasons) or if we should align the hierarchy of these headings

mirnawong1: That's a good question and it's mostly for stylistic reasons since I thought having a bullet looks too busy so I wanted to separate them with sections, but please let me know if there's a better way to format, maybe just bold?

luna-bianca: A bulleted list + bold formatting could work since we have the sentence "Choose your local development experience from these tools:" to introduce the list!

mirnawong1: ok i originally had it like that but changed. i'll change it back

luna-bianca (approved):

PR #8563, opened 2026-02-08
Correct reference to related model in `dbt test --select` documentation

MichaelG-create:
In the doc. on `dbt test --select` I changed a reference to another model name in on sentence (possible typo).
The sentence was refering twice to the same model.

So:
'For example, when you run dbt test --select model_b, dbt includes tests defined on model_b as well as tests on related models (like model_b) that reference model_b.'

would become:
'For example, when you run dbt test --select model_b, dbt includes tests defined on model_b as well as tests on related models (like model_a) that reference model_b.'


mirnawong1 (approved):
hey @MichaelG-create great catch and thanks for creating a pr to fix this up!

MichaelG-create:
Glad it helped :)

PR #8655, opened 2026-02-24
clarify database admin permissions

mirnawong1:
database admin permission explaination a bit confusing. this pr tries to clarify what access the permission has wrt connections.

raised internally https://dbt-labs.slack.com/archives/C02NCQ9483C/p1771916033647449 and confirming here https://dbt-labs.slack.com/archives/C04KKUHLT32/p1771929745578389

---

- https://docs-getdbt-com-git-update-database-admin-dbt-labs.vercel.app/docs/cloud/manage-access/enterprise-permissions


ericbaumann:
personally i think we should just add the Account permission write_connections permission to this permission set.

they have project_level connections_write but that is only good for legacy APIs and is not used from cloud-ui as far as I can remember. 

I'll approve and you can ask @nehahystad and the enterprise team about that change.

ericbaumann (approved):

mirnawong1:
hey @ericbaumann thanks so much! do you mean adding the account permission write programmatically?

ericbaumann:
> hey @ericbaumann thanks so much! do you mean adding the account permission write programmatically?

yes, i mean modifying this [array](https://github.com/dbt-labs/dbt-cloud/blob/3968095b583a4cc60348ac7a56136d204ff627ab/sinter/models/permissions/sets/project/database_admin.py#L12-L28) to include the value `AccountPermissions.connections_write`

ericbaumann:
this will make the docs accurate as they exist today. What i am supposing in my comment is that `ProjectPermissions.connections_write` is effectively useless for a cloud-ui user. It only has value if you are hitting the [legacy connections API ](https://docs.getdbt.com/dbt-cloud/api-v3#/operations/Create%20Connection) which is indicated as deprecated in the docs.

PR #8684, opened 2026-02-28
Create rehype plugin that readds markdown codeblock language to codeblocks

joshuataylor:
See https://docusaurus.io/docs/markdown-features/plugins

This plugin adds the language from codeblocks into the final built markdown, for pages like
https://docs.getdbt.com/reference/dbt-jinja-functions/source.md.

E.g:

```sql
select 1
```

Compiles to:

```
select 1
```

This plugin re-adds the sql block.

---

Context: I'm building a plugin for JetBrains IDEs (PyCharm, IntelliJ, DataSpelll etc), and have a documentation provider that shows documentation on hover, e.g for ref:

image

But the compiled markdown files don't contain the types, so it loses its syntax highlighting.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -1,6 +1,7 @@
 import path from "path";
 import math from "remark-math";
 import katex from "rehype-katex";
+import rehypeCodeLanguage from "./plugins/rehypeCodeLanguage.js";
 const { themes } = require('prism-react-renderer')
 
 const { versions, versionedPages, versionedCategories } = require("./dbt-versions");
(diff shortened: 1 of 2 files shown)


mirnawong1:
hey @joshuataylor , this is really cool and thanks for opening this up! i'll share this with the team and we'll review this soon!

joshuataylor:
FYI - I've setup a process which pushes to https://github.com/joshuataylor/dbt-docs-markdown for now, built on this patch. It allows me to get going with something, as I use the dbt docs as a documentation provider.

mirnawong1 (approved):
preview looks good, approving! thanks for opening this up @joshuataylor !

PR #8693, opened 2026-03-02
Remove Mutiny references

john-rock:
Removes mutiny script and references now that our contract has expired.


JKarlavige (approved):
lgtm! :shipit:

PR #8753, opened 2026-03-10
Hiding the cache page for fusion where it is deprecated

Bazsy:
 
The function described on the cache page is deprecated in Fusion according to:
https://docs.getdbt.com/docs/dbt-versions/core-upgrade/upgrading-to-fusion?version=1.12#deprecated-flags

In addition fusion gives a warning if this is used: 
"warning: dbt1700: --cache-selected is no longer supported"


mirnawong1 (approved):
hey @Bazsy good catch and thanks for opening this up!

PR #8767, opened 2026-03-11
Fix example for unit test with `format: sql`

jtcohen6:
The column name for input fixture `rows` needs to match the `expect` rows (expected output)

---

- https://docs-getdbt-com-git-jerco-fix-unit-test-sql-example-dbt-labs.vercel.app/docs/build/unit-tests


mirnawong1 (approved):
thanks so much for opening this up @jtcohen6 ! great to see you here ✨

PR #8856, opened 2026-03-26
Update date filters to use session_start instead of page_view_start

roberto-rosero:
In the config section of sql model is configured like event_time --> session_start right? It must be the real filter, isnt it?


matthewshaver:
@roberto-rosero dbt auto-filters upstream ref() models based on their own configured `event_time`. The `page_views` upstream model has event_time='page_view_start', so dbt correctly generates WHERE page_view_start >= ... when it queries that upstream table. The filter column in the compiled SQL reflects the upstream model's event_time config, not the downstream model's.

PR #8884, opened 2026-03-31
Support for admin api for remote dbt mcp

DevonFulcher:

We recently added support for the admin API tools to remote MCP. I think this is the only place where this restriction is mentioned.


mirnawong1 (approved):
good shout! thank you @DevonFulcher !

PR #8929, opened 2026-04-09
Clarify override scoping and fix typos in unit-test-overrides

### LINKED ISSUE #8735: Overrides in unit tests only work for direct references, not indirect ones
### Contributions

https://docs.getdbt.com/reference/resource-properties/unit-test-overrides

Explain that unit test overrides only apply to macros, vars, env vars directly used within the model being unit tested. The override will replace what the macro (or var or env var) would normally return. If you provide an override that is only used indirectly, it will **not** be applied.

### Additional information

See here for context:
- https://github.com/dbt-labs/dbt-core/issues/12165

EoinTrial:
Closes #8735.

The Macros section states "You can override the output of any macro in your unit test definition" without qualification, which implies overrides work for any macro anywhere in the call stack.

This PR adds a clarifying sentence explaining that overrides only apply to macros, variables, and environment variables called directly within the model being tested. Indirect references (inside a macro that the model calls) are not overridden.

Also fixes three typos in the same file:
- `defition` → `definition`
- `explicity` → `explicitly` (×2, including inside a code comment)
- `overidden` → `overridden`

Checklist:
- [x] I have read the contribution docs
- [ ] The content in this PR requires a dbt release note (not applicable)
- [x] Applied proper versioning rules if applicable (not applicable either, this is a correction to existing content, not new versioned functionality)


runleonarun (approved):
This looks excellent @EoinTrial! Approving and will merge this PR soon

PR #8955, opened 2026-04-13
Update pointers to artifact architecture doc

joellabes:
The docs point to a file at a point in time, which has since been moved to the architecture directory. 

The link didn't 404 since it was pointing to a specific commit, but it was stale. This updates it. 

Potentially open to pinning to a specific commit if this is a docs team standard practice, but it diverging over time feels bad 🤷

---

- https://docs-getdbt-com-git-joellabes-patch-4-dbt-labs.vercel.app/reference/global-configs/behavior-changes


mirnawong1 (approved):
looks like a good solution, thanks @joellabes and this is fair!

PR #8984, opened 2026-04-16
Remove redundant scheduler='ENABLE' from example

igorbelianski-cyber:
## Checklist

### DIFF
--- a/website/docs/reference/resource-configs/snowflake-configs.md
+++ b/website/docs/reference/resource-configs/snowflake-configs.md
@@ -391,7 +391,6 @@ To enable Snowflake-managed scheduling with a target lag:
 {{ config(
     materialized='dynamic_table',
     snowflake_warehouse='MY_WH',
-    scheduler='ENABLE',
     target_lag='5 minutes',
 ) }}
 


igorbelianski-cyber:
@luna-bianca  one line fix PTAL

website/docs/reference/resource-configs/snowflake-configs.md
```diff
@@ -391,7 +391,6 @@ To enable Snowflake-managed scheduling with a target lag:
 {{ config(
     materialized='dynamic_table',
     snowflake_warehouse='MY_WH',
-    scheduler='ENABLE',
```
luna-bianca, line 394: Hey @igorbelianski-cyber, thanks for opening this PR! Is it necessary to remove `scheduler='ENABLE'` here? In the docs, we currently state that when `scheduler: ENABLE` is set, `target_lag` is required. Keeping it in the sample can help reinforce that relationship, rather than relying on the default behavior. 

Also, since this example is under the `Scheduler` section, it might make sense to explicitly include the parameter in the example. Thoughts?

igorbelianski-cyber: I think removing it gives clear message that this is not needed in this context similar to the example above (we didn't add SCHEDULER = DISABLE there). 

I think most customers 99% will never specify SCHEDULER, just  use target lag (present or absent to control behavior) 
We added scheduler = ENABLE|DISABLE support more for completeness ( so it is not obviously missing) as opposed to enable some meaningful scenario.

luna-bianca: Thanks for the clarification, @igorbelianski-cyber!

luna-bianca (approved):

PR #9003, opened 2026-04-20
remove old internal dbt package

JKarlavige:

Removes old dependency pointing to the archived [react-dbt-dag repo](https://github.com/dbt-labs/react-dbt-dag), along with the associated components/files. This was the only thing we were using the `NPM_TOKEN` env variable for. Verify eslint git action still passes.


john-rock (approved):
:shipit:

PR #9246, opened 2026-05-26
fix: restore inline code colors in card bodies in dark mode

luna-bianca:
## Problem

`<code>` tags inside `Card` component `body` props are unreadable in dark mode. The only affected page is `/docs/build/organize-your-outputs`.

**Root cause:** This rule in `custom.css` sets `color: white` on every child of linked cards in dark mode:

```css
html[data-theme="dark"] main .row .col:first-of-type a article * {
  color: white;
}
```

That wildcard matches `<code>` elements injected via `dangerouslySetInnerHTML` inside `.cardBody`, overriding the `background-color: rgba(246, 245, 255, 1)` and `color: rgba(61, 24, 154, 1)` set by `:not(pre) > code`. The result is white text on a near-white background — invisible.

## Fix

One rule added to `custom.css` immediately after the wildcard, with identical specificity but targeting `code` specifically, so it wins:

```css
html[data-theme="dark"] main .row .col:first-of-type a article code {
  background-color: rgba(246, 245, 255, 1);
  color: rgba(61, 24, 154, 1);
}
```

## Files changed

- `website/src/css/custom.css` (+5 lines)

## Testing

Verified on Vercel preview in dark mode at `/docs/build/organize-your-outputs` — `schema`, `database`, and `alias` code chips now render correctly and match inline backtick code elsewhere on the site.


JKarlavige (approved):
Nice fix @luna-bianca! Confirmed the issue on the live site, and confirmed the code text is readable in dark mode with this edit! :shipit:

PR #9301, opened 2026-06-03
fix: resolve platform function doc URLs correctly

mirnawong1:
The function-tracking workflow built platform doc URLs by naive string concatenation, which breaks on root-relative hrefs. Snowflake now emits root-relative hrefs (e.g. `/sql-reference/functions/finetune-show`), producing doubled paths like `/en/sql-reference//sql-reference/functions/finetune-show` that 404. These URLs flow directly into the auto-filed gap issues (see [dbt-core#15115](https://github.com/dbt-labs/dbt-core/issues/15115)).

## Changes
- Replace string concatenation with WHATWG `URL` resolution (`new URL(href, pageUrl)`) in `website/scripts/platforms.config.js`, applied across all six platforms (Snowflake, Databricks, Redshift, BigQuery, Trino, DuckDB).
- Normalize Snowflakes missing `/en` locale so links hit the canonical page instead of 301-redirecting.

## Verification
Ran the real parser against the live Snowflake all-functions page (941 functions):
- 0 broken double-path URLs
- 0 missing `/en/`
- `finetune-show` resolves to `https://docs.snowflake.com/en/sql-reference/functions/finetune-show`

The committed `snowflake.json` is currently clean (predates Snowflakes markup change), so no data backfill is needed — the next workflow run regenerates correct URLs.


mengdilin-dev (approved):

mirnawong1:
thanks so much @mengdilin-dev !

PR #9334, opened 2026-06-05
PRODDOCS-867: sample-flag

nataliefiann:
Resolves [JIRA:PRODDOCS-867](https://dbtlabs.atlassian.net/browse/PRODDOCS-867)

This PR updates the [About sample flag doc](https://docs.getdbt.com/docs/build/sample-flag?version=2.0&name=Fusion):

- Explains which field `--sample` uses (`event_time`) and what happens if it's missing
- Adds how filtering works (subquery, UTC time window)
- Fixes the `stg_orders` example command
- Rewords the static date range example
- Improves the `.render()` opt-out example with a join scenario
- Moves the "what happens when you run `--sample`" sentence before Examples

## Checklist

---

- https://docs-getdbt-com-git-nfiann-sampleflag-dbt-labs.vercel.app/docs/build/sample-flag


website/docs/docs/build/sample-flag.md
runleonarun: This is not only the IDE, right? It's all CLI environments?
```suggestion
Let's say you want to run your `stg_customers` model and build the table in your development schema with a relative time spec sample size of three days. Your CLI command would look something like this:
```

website/docs/docs/build/sample-flag.md
```diff
@@ -43,33 +46,39 @@ Let's say you want to run your `stg_customers` model and build the table in your
...
+
+)
 
+select *
+from orders
+left join customers using (customer_id)
 ```
 
-dbt will then execute the model SQL against the target data warehouse and build the tables with data from the sample sizes.
+In this example, `orders` is sampled using the `event_time` configured on the `jaffle_shop.orders` source. `stg_customers` is not sampled because `.render()` skips the time filter for that `ref()`.
```
runleonarun: >`orders` is sampled using the `event_time` configured on the `jaffle_shop.orders

where is this ^ ? I don't see the event_time in the example? Does this happen in the command above it when you build the table using `dbt run --select path/to/stg_orders --sample="6 hours"` [here](https://github.com/dbt-labs/docs.getdbt.com/pull/9334/changes#diff-162043e58be3974108fc547312cb4852ae145c204cb9b2ae10f9695bd19dc936R52)
 If so, that example uses stg_orders and here we say "orders" should they be the same?

website/docs/docs/build/sample-flag.md
```diff
@@ -43,33 +46,39 @@ Let's say you want to run your `stg_customers` model and build the table in your
...
 
-Next, let's say you want to validate data for your entire business from a sample size further in the past - your busiest week in July, from the first until closing time on the eighth. You can run the following: 
+Next, to sample a fixed historical time range instead of a window relative to when the command runs, specify a start and end date or timestamp. For example, to validate data for your entire business during your busiest week in July, from the first until closing time on the eighth:
 
 ```
 dbt run --sample="{'start': '2024-07-01', 'end': '2024-07-08 18:00:00'}"
 ```
 
-To prevent a `ref` from being sampled, append `.render()` to it:
+To prevent a `ref()` or `source()` from being sampled, append `.render()` to it. For example, you might keep a small dimension table unfiltered while still sampling a large fact table:
```
runleonarun: The following is not a fct table, but maybe it should be? If so, we should also change the fct in the previous CLI example:

> dbt run --select path/to/fct_orders --sample="6 hours"

website/docs/docs/build/sample-flag.md
```diff
@@ -43,33 +46,39 @@ Let's say you want to run your `stg_customers` model and build the table in your
...
+customers as (
 
     select * from {{ ref('stg_customers').render() }}
 
 ),
 
-...
+orders as (
+
+    select * from {{ source('jaffle_shop', 'orders') }}
```
runleonarun: make this fct table?

website/docs/docs/build/sample-flag.md
runleonarun: Is this correct? are we adding this jinja to the model's sql file?
```suggestion
To prevent a `ref()` or `source()` from being sampled, append `.render()` in the model's `.sql` file. For example, you might keep a small dimension table unfiltered while still sampling a large fact table:
```

runleonarun (commented):
Hey @nataliefiann I had some questions about this that might help make it clearer for our users too.

PR #9346, opened 2026-06-08
fix(ci): prevent shell injection in wizard CLI drift issue creation

mirnawong1:
## What

The `drift-check` job in `wizard-cli-drift.yml` was failing with exit code 126 in the **Open drift issue** step.

The issue body was built with an unquoted heredoc (`cat <<BODY`) and the captured diff was spliced in via `${{ steps.drift.outputs.diff }}`. That diff is the wizard `--help` output — full of backticks, `$(...)`, and flag syntax like `-c, --config <key=value>` — so the shell tried to execute it, producing `command not found` / `stdin is not a terminal` errors and finally `gh: Argument list too long`.

## Changes

- Move `${{ }}` expressions into `env:` so untrusted content is never spliced into the script source (GitHub-recommended script-injection mitigation).
- Build the issue body into `issue_body.md` using `printf '%s\n'` (no expansion) and create the issue with `--body-file`. Also resolves the "Argument list too long" error.
- Apply the same env-var treatment to the comment step.

## Testing

Workflow YAML validates. After merge, can be re-run manually via `workflow_dispatch`.


runleonarun (approved):
Approving because I'm not sure how to test.

PR #9377, opened 2026-06-10
docs(databricks): add metric_view materialization for Unity Catalog metric views

sd-db:
## What this PR does

Adds documentation for the new `metric_view` materialization in dbt-databricks (databricks/dbt-databricks#1285) to the Databricks configs reference.

Set `materialized='metric_view'` on a model and write the body as a Unity Catalog metric view YAML definition (`version`, `source`, `dimensions`, `measures`, optional `filter`). dbt creates it via `CREATE OR REPLACE VIEW ... WITH METRICS LANGUAGE YAML`, and supports `databricks_tags`, `tblproperties`, and `grants`. With `view_update_via_alter=true`, YAML-definition changes are applied via `ALTER VIEW ... AS` and tag/tblproperty changes via `ALTER ... SET` instead of a full replace.


luna-bianca:
Hey @sd-db , it looks like a merge got committed with unresolved conflict markers still in the file. `databricks-configs.md` currently contains `<<<<<<<`, `=======`, and `>>>>>>>` markers around the #### query section, which is making this hard to review.

The conflict bundled the new Metric views section together with an older copy of the query / on_configuration_change / streaming-tables text, while current has a reworded version of that same text. Could you re-resolve so that:

- The current side's updated query / on_configuration_change / streaming-tables wording is kept (don't revert it to the older phrasing)
- Only the new ## Metric views section is re-added
- All <<<<<<< / ======= / >>>>>>> markers are removed

Happy to re-review once the markers are gone. Thank you!

sd-db:
> Hey @sd-db , it looks like a merge got committed with unresolved conflict markers still in the file. 
> Happy to re-review once the markers are gone. Thank you!

Fixed this + added some more things as recommended by @Tonayya . Thx !

matthewshaver (approved):
THanks @sd-db !

PR #9380, opened 2026-06-10
docs(databricks): support cron, interval, and on-update schedules for MVs and streaming tables

sd-db:
## What this PR does

Documents that materialized views and streaming tables now support three refresh-schedule modes via the `schedule` config, as of dbt-databricks v1.12. Previously only cron schedules were documented.

This documents the feature added in databricks/dbt-databricks#1434.

## Changes

- Adds `SCHEDULE EVERY` and `TRIGGER ON UPDATE` rows to the MV/ST configuration support table (v1.12+).
- Rewrites the `#### schedule` section to describe the three mutually exclusive modes:
  - `cron` — refresh on a cron schedule (`time_zone_value` valid only here).
  - `every` (v1.12+) — refresh on a fixed interval (`SCHEDULE EVERY`), units `HOURS`/`DAYS`/`WEEKS`.
  - `on_update` (v1.12+) — refresh on upstream change (`TRIGGER ON UPDATE`), with optional `at_most_every` rate limit (min 60 seconds).
- Clarifies that dbt issues a manual refresh on every run for `cron`, but not for `every`/`on_update` (Databricks auto-manages those).

## Which audience does this affect?

dbt-databricks users on v1.12+ who configure refresh schedules on materialized views or streaming tables.


mirnawong1 (approved):
hey @sd-db thanks so much for adding this! some small tweaks and approving!

PR #9397, opened 2026-06-11
Basic pagination for job-based endpoints

nataliefiann:
This PR updates the [Query the Discovery API](https://docs.getdbt.com/docs/dbt-apis/discovery-querying?version=2.0), [Models object schema](https://docs.getdbt.com/docs/dbt-apis/discovery-schema-job-models?version=2.0), and [Use cases and examples for the Discovery API](https://docs.getdbt.com/docs/dbt-apis/discovery-use-cases-and-examples?version=2.0) pages:

- `discovery-querying.md` — main guide (behavior, setup steps 1–7, examples
- `discovery-use-cases-and-examples.md` — short paginated example + link
 - `schema-discovery-job-models.md` — schema page example + link
 
 Eng PR: https://github.com/dbt-labs/codex-api/pull/1456
 Slack thread: https://dbt-labs.slack.com/archives/CBHMT978R/p1779206113181719

## Checklist

---

- https://docs-getdbt-com-git-nfiann-job-based-endpoints-dbt-labs.vercel.app/docs/dbt-apis/discovery-querying
- https://docs-getdbt-com-git-nfiann-job-based-endpoints-dbt-labs.vercel.app/docs/dbt-apis/discovery-use-cases-and-examples


nataliefiann:
## Test plan

Tested job-based Discovery API cursor pagination against `at049` (US1 MT).

### Setup
- Created a **Metadata Only** service token in **Account settings** → **Service Tokens**
- Discovery API URL: `https://at049.metadata.us1.dbt.com/graphql`
- Job ID `70437463740595` from job URL
- Run ID `70437506051490` from run URL

### First request (`first: 10`, `after: null`)
- Ran the paginated `job { models(...) }` query via `curl` with `Authorization: Token <token>`
- Response returned 6 models, each with `uniqueId` and `paginationCursor`
- Confirmed opaque base64 cursors (not plain `uniqueId`)

### Pagination (`first: 3`, `after: null`)
- Re-ran with `first: 3` to force multiple pages on a small job
- Page 1 returned 3 models: `dim_properties`, `fct_leases`, `stg_leases`

### Result
- Pagination works on this instance
- `first` / `after` / `paginationCursor` behave as documented
- Job with 6 models and `first: 10` returns a single page (fewer rows than `first` = last page)

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,6 +199,113 @@ pageInfo {
 totalCount # Total number of records across all pages
 ```
 
+#### Job-based queries
+
+Job-based list endpoints also support cursor pagination. Use this pattern when you query resources under the [`job`](/docs/dbt-apis/discovery-schema-job) object (or the equivalent legacy top-level queries such as `models(jobId:, runId:)`).
```
matthewshaver:
```suggestion
Job-based list endpoints also support cursor-based pagination. Use this pattern when you query resources under the [`job`](/docs/dbt-apis/discovery-schema-job) object (or the equivalent legacy top-level queries such as `models(jobId:, runId:)`).
```

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,6 +199,113 @@ pageInfo {
...
+- `tests`
+- `macros`
+- `metrics`
+- `exposures`
+
+Pass `first` to set the page size (capped at 100). Each returned item includes a `paginationCursor` field. Pass that value as `after` on the next request to fetch the following page.
+
+Unlike environment queries, these job-based endpoints do not return a `PageInfo` object or `hasNextPage` field. When a page contains fewer rows than `first`, you have reached the last page.
+
+If you omit `first` and `after`, the API still returns all matching rows today. In a future update, unpaginated requests will return at most 100 rows. That change will roll out after advance notice. Until then, pass `first` and `after` explicitly when you query large jobs.
```
matthewshaver: Maybe this should be a callout. We can alter it as the dates are defined. Makes it easier to identify when it's time to remove as well

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,6 +199,113 @@ pageInfo {
...
+        }
+      ]
+    }
+  }
+}
+```
+
+**Next page**
+
+Set `after` to the `paginationCursor` from the **last** row of the previous page. The cursor is an opaque encoded string, not the `uniqueId`. Keep `first` the same unless you want a different page size.
```
matthewshaver:
```suggestion
Set `after` to the `paginationCursor` from the _last_ row of the previous page. The cursor is an opaque encoded string, not the `uniqueId`. Keep `first` the same unless you want a different page size.
```

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,6 +199,113 @@ pageInfo {
 totalCount # Total number of records across all pages
 ```
 
+#### Job-based queries
```
matthewshaver: Love the attention to detail here, but this H4 is almost three times as long as the H3 it sits under or any other section on this page. Is there any way to condense it or to link out to the API docs for the users to test examples instead of covering it step-by-step here? Not a problem if you feel this level of detail is required to accurately convey the feature

matthewshaver (approved):

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,34 +199,72 @@ pageInfo {
 totalCount # Total number of records across all pages
 ```
 
-The `PageInfo` and `totalCount` pattern described above applies to the `environment` endpoints, which return results as a connection of `edges` and `nodes`.
+The previously described `PageInfo` and `totalCount` pattern applies to the `environment` endpoints, which return results as a connection of `edges` and `nodes`.
 
 #### Job-based endpoint pagination
 
-The job-based endpoints (the list fields on the `job` object, such as `models`, `sources`, `seeds`, `snapshots`, `tests`, `macros`, `metrics`, and `exposures`) use a simpler form of cursor-based pagination. These endpoints return a flat list rather than an `edges`/`nodes` connection, so they don't expose a `PageInfo` object. Instead:
+Job-based list endpoints also support cursor-based pagination. Use this pattern when you query resources under the [`job`](/docs/dbt-apis/discovery-schema-job) object (or the equivalent legacy top-level queries such as `models(jobId:, runId:)`). These endpoints return a flat list rather than an `edges`/`nodes` connection, so they do not expose a `PageInfo` object.
```
jzhu13: let's not mention this portion

"(or the equivalent legacy top-level queries such as `models(jobId:, runId:)`)"

There's no traffic to those and they're marked as deprecated because they don't support bigint (required in a multicell world)

website/docs/docs/dbt-apis/discovery-querying.md
```diff
@@ -199,34 +199,72 @@ pageInfo {
...
+- `metrics`
+- `exposures`
 
-There's no `hasNextPage` field on these endpoints. You've reached the end of the result set when a page returns fewer elements than `first` (or fewer than the default page size).
+Pass `first` to set the page size (capped at 100). Each returned item includes a `paginationCursor` field—an opaque, per-element value you select in your query. Pass the last item's `paginationCursor` as `after` on the next request to fetch the following page.
 
-Below is an example that fetches a page of models from a job and selects the `paginationCursor` for each model so you can request the next page:
+Unlike environment queries, these job-based endpoints do not return a `PageInfo` object or `hasNextPage` field. You have reached the last page when a page returns fewer rows than `first` (or fewer than the default page size).
+
+We recommend always specifying `first` to keep response sizes manageable. If you omit `first` and `after`, the API returns all matching rows today. For large jobs, pass `first` and `after` explicitly so your integration does not depend on unpaginated responses.
```
jzhu13: "If you omit `first` and `after`, the API returns all matching rows today"

Even with the note regarding an upcoming change below, I think we shouldn't say "the API returns all matching rows today". Would it be possible to change it to something like "historically, the API returned all rows"? I don't want to encourage any newer use cases to rely on that behavior.

website/docs/docs/dbt-apis/discovery-querying.md
runleonarun:
```suggestion
Job-based list endpoints also support cursor-based pagination. Use this pattern when you query resources under the [`job`](/docs/dbt-apis/discovery-schema-job) object. These endpoints return a flat list rather than an `edges`/`nodes` connection, so they do not expose a `PageInfo` object.
```

website/docs/docs/dbt-apis/discovery-querying.md
runleonarun:
```suggestion
We recommend always specifying `first` to keep response sizes manageable. Historically, omitting `first` and `after` would cause the API to return all matching rows. For large jobs, best practice is to pass `first` and `after` explicitly so your integration does not depend on unpaginated responses.
```

PR #9478, opened 2026-06-23
Document legacy Team plan ineligibility for dbt State

mirnawong1:
users were confused and asking whether legacy Team plan customers were eligible for dbt State. clarifying docs to clarify this and avoid any confusion

## Changes

- **About dbt State** — Added a `:::note` callout in the "Signing up for dbt State" section clarifying that legacy Team plan accounts are not eligible.
- **Set up dbt State** — Added the same eligibility note to the "Before you begin" prerequisites list.


mirnawong1:
Superseded by a clean PR with only the dbt State changes.

PR #9482, opened 2026-06-23
docs: sync MT release notes for June 17, 2026

nataliefiann:
Automated sync of weekly single-tenant release notes into the monthly multi-tenant/VPC release notes.

Week synced: **June 17, 2026**

Please review for accuracy and add any MT-only items manually.

(Test PR from Test 2 — close without merging.)

Made with [Cursor](https://cursor.com)

PR #9483, opened 2026-06-23
Fix publish workflow permissions for Test 3

nataliefiann:
## Summary
- Fix `gh pr ready` failure (`Resource not accessible by integration`) by using REST API to mark draft PRs ready
- Send Slack via Zapier **before** mark-ready so the ping still fires if that step fails
- Auto-create `release-notes` label in sync workflow if missing (fixes Test 2 label error)

## Test plan
- [ ] Merge this PR
- [ ] Re-run **Publish release notes PR** with draft #9482 open
- [ ] Confirm Slack message in `#tmp-test-agents`

Made with [Cursor](https://cursor.com)


.github/workflows/sync-release-notes.yml
```diff
@@ -92,6 +92,10 @@ jobs:
             exit 0
           fi
 
+          gh label create "release-notes" \
+            --description "Automated ST to MT release notes sync PRs" \
+            --color "43CA69" 2>/dev/null || true
+
```
johnchappelledbt: I'm a little unclear on this one. Do you need to create a new label on each run?

johnchappelledbt: I guess, what I mean is, you could create the label once on the repo manually. This would try to create it each time, and ignore the failure if it already exists. Could skip this call to create entirely, unless I'm misunderstanding the intent of this.

.github/workflows/publish-release-notes.yml
```diff
@@ -65,3 +60,14 @@ jobs:
...
+      - name: Mark PR ready for review
+        id: mark_ready
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh api \
+            --method PATCH \
+            "repos/${{ github.repository }}/pulls/${{ steps.pr.outputs.number }}" \
+            --input - <<< '{"draft":false}'
```
johnchappelledbt, line 73: It is possible that the GH api doesn't support moving a PR into ready status, in this way. When I asked Codex, it suggested 

```sh

gh pr ready "${{ steps.pr.outputs.number }}"

```

instead

.github/workflows/sync-release-notes.yml
nataliefiann: Thanks for taking a look at this @johnchappelledbt 

Good point — the label already exists now, so we don't need to create it on every run. I'll remove that step from the workflow.

.github/workflows/publish-release-notes.yml
nataliefiann: Hiya @johnchappelledbt 

I tried the gh pr ready in an earlier [test](https://github.com/dbt-labs/docs.getdbt.com/actions/runs/28033879503) and it failed with `Resource not accessible by integration from the default GITHUB_TOKEN`. That's why I switched to the REST API.

johnchappelledbt (approved):

PR #9505, opened 2026-06-26
dbt lint updates

matthewshaver:

dbt lint doc updates

- Document dbt format 
- Document limitations around lint and macros
- Add Render variant limit limitation

## Checklist

---

- https://docs-getdbt-com-git-lint-updates-dbt-labs.vercel.app/reference/commands/lint


website/docs/reference/commands/lint.md
```diff
@@ -68,13 +68,28 @@ The following dialects are currently supported with `dbt lint`:
...
+`[FILE]` is optional. When omitted, `dbt format` formats all SQL files in your project.
+
 ## Beta limitations
 
 Keep these limitations in mind:
 
-### Layout rules
+### Macro linting
+
+`dbt lint`, by default, doesn't lint the SQL that macro calls produce. This behavior is similar to SQLFluff's `ignore_templated_areas`, but you can't configure it in dbt today. As a result, `dbt lint` doesn't run introspection queries during Jinja render time, and it suppresses lint diagnostics from SQL that macros produce.
```
eliasdefaria: I would reframe this section. 

The last section you mentioned is the most accurate part "it suppresses lint diagnostics from SQL that macros produce" 

I don't think this should be in beta limitations, but rather should be something we call out as an "opinionated decision" or ours or something. Like an FAQ? I'm sure you have prior art for something like this. 

Essentially we do lint the SQL produced, we just don't issue diagnostics if it came from a specific type of macro. Specifically those that would issue introspection queries if `execute` were true. This is because without setting `execute` to true, these macros typically produce invalid SQL. 

This models the behavior of `ignore_templated_areas` in SQLFluff, however we do not allow this to be configured as we don't believe it's the right model for dbt projects.

website/docs/reference/commands/lint.md
```diff
@@ -68,13 +68,28 @@ The following dialects are currently supported with `dbt lint`:
...
+
+### Render variant limit
 
-`dbt lint` doesn't currently issue layout (`LT*`) warnings. Layout enforcement will ship as a separate subcommand, `dbt fmt` (coming soon), which will format code to your layout policy rather than warn. This keeps lint output focused on correctness and style issues that require human intervention, while layout becomes a formatting step.
```
eliasdefaria, line 77: Same as above with this section, this is not a beta limitation but rather an explicit opinionated choice. We are open to feedback of course, but we believe this is the right way to lint dbt projects.

wizardxz (approved):

website/docs/reference/commands/lint.md
```diff
@@ -89,6 +96,24 @@ The following rules report violations but can't be auto-fixed by `--fix`. They r
...
+
+This is a deliberate choice. Linting every possible render variant is expensive and surfaces violations in SQL your project may never execute. dbt Labs believes linting the SQL your project produces using its parse-time inputs is the right model for dbt projects. If you have feedback on this approach, open an issue in the [dbt-core GitHub repository](https://github.com/dbt-labs/dbt-core/issues) with the `Linter` label.
+
+</DetailsToggle>
+
+<DetailsToggle alt_header="Why doesn't dbt lint report violations from some macros?">
+
+`dbt lint` lints the SQL that all macros produce. However, it intentionally suppresses diagnostics for macros that would issue introspection queries if `execute` were set to `true`. Without `execute` enabled, these macros typically produce invalid SQL. Flagging violations against that output generates noise rather than signal.
+
+This behavior is similar to SQLFluff's `ignore_templated_areas` setting. However, you can't configure it today. dbt Labs believes suppressing diagnostics for introspection macros is the correct default for dbt projects, and exposing a toggle would create more confusion than it resolves.
```
eliasdefaria: I would tone back this part a bit but the rest looks good. Maybe just and with However, you can't configure it today

matthewshaver:
```suggestion
This behavior is similar to SQLFluff's `ignore_templated_areas` setting. However, you can't configure it today.
```

runleonarun (approved):
Looks good

PR #9510, opened 2026-06-26
Fix unlisted guides appearing in guides page search

runleonarun:
## Summary

- The `buildQuickstartIndexPage` plugin loaded **all** files from `docs/guides/` into the search index without checking the `unlisted` frontmatter flag
- Guides with `unlisted: true` were still appearing in the `/guides` page search even though they were correctly hidden from the sidebar and nav
- Added a check to skip any guide where `unlisted: true` before adding it to the data passed to the search component

## Test plan

- [ ] Confirm that a guide with `unlisted: true` (e.g. `tenancy-migration`) does not appear in the guides page search
- [ ] Confirm that all other guides still appear as expected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JKarlavige (approved):
Nice fix! Confirmed the "Migrate between dbt platform tenancy types" appears in the guide search on the live site, and no longer appears in the guide search within this preview :shipit:

PR #9577, opened 2026-07-09
dbt State doc updates

luna-bianca:

Addresses the following:
- [Slack thread 1](https://dbt-labs.slack.com/archives/C092RKDSGRL/p1783530304081519?thread_ts=1783409736.853799&cid=C092RKDSGRL)
- [Slack thread 2](https://dbt-labs.slack.com/archives/C092RKDSGRL/p1783420277510679)

Updates three dbt State docs:

- **`dbt-state-about.md`**: Expanded the clone bullet to clarify that dbt State searches all environments and jobs for the freshest matching object, including schemas where a model was built before it ever ran in production.
- **`faqs/State/views-rebuilt.md`**: Added a new section on BigQuery external sources (such as Google Sheets), which always rebuild because BigQuery doesn't expose modification timestamps. Also updated the intro to note that some rebuilds happen due to freshness determination, not just SQL changes, and updated "These two patterns" to "The following patterns" to reflect the added third case.
- **`dbt-state-migration.md`**: Added BigQuery external sources as a third cause of unexpected rebuilds in the "Known differences" section.

## Preview links

- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/docs/deploy/dbt-state-about
- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/docs/deploy/dbt-state-migration
- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/faqs/State/views-rebuilt

## Checklist

---

- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/docs/deploy/dbt-state-about
- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/docs/deploy/dbt-state-migration
- https://docs-getdbt-com-git-state-doc-updates-dbt-labs.vercel.app/faqs/State/views-rebuilt


website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -151,6 +151,7 @@ State-aware orchestration and dbt State differ in a few ways:
 - **More models rebuilding than expected**: If you notice more rebuilds after you migrate, the most common causes are:
   - **Views with `select *`**: dbt State can't determine which columns `select *` resolves to without querying the upstream schema, so it always rebuilds these views rather than risk reusing a stale result.
   - **Non-determinism in Jinja-templated SQL**: Macros like `dbt_utils.get_relations_by_pattern` with `dbt_utils.union_relations` can return relations in a different order on each run, which produces different compiled SQL. dbt State detects a new hash and rebuilds the model. If that model has downstream dependencies, those models rebuild, too.
+  - **Models with external sources on BigQuery**: Models that use external sources (such as Google Sheets) always rebuild because BigQuery doesn't expose modification timestamps for external sources, so dbt State can't determine freshness.
```
reubenmc: Can we add a callout that they can used a loaded at query or loaded at field and point to any timestamp field if they don't want the source to always be considered fresh?

website/docs/faqs/State/views-rebuilt.md
```diff
@@ -5,12 +5,13 @@ sidebar_label: 'Why is my model being rebuilt instead of reused?'
 id: views-rebuilt
 ---
 
-dbt State decides whether to reuse a model by comparing its compiled SQL to the hash from the previous run. If anything changes that compiled SQL between runs, even when the logic hasn't meaningfully changed, dbt State rebuilds the model.
+dbt State decides whether to reuse a model by comparing its compiled SQL to the hash from the previous run. If anything changes that compiled SQL between runs, even when the logic hasn't meaningfully changed, dbt State rebuilds the model. In some cases, dbt State also rebuilds when it can't determine source freshness, regardless of SQL changes.
```
reubenmc: Can we add the why here:
Something like: The basis for state's decisions is that it is better to build (the previous behavior without state) than skip building when it's unclear if the data has changed.

reubenmc (approved):
A few quick suggestions, but looks great!

website/docs/faqs/State/views-rebuilt.md
luna-bianca: @reubenmc what do you think of:

```suggestion
dbt State decides whether to reuse a model by comparing its compiled SQL to the hash from the previous run. If anything changes that compiled SQL between runs, even when the logic hasn't meaningfully changed, dbt State rebuilds the model. In some cases, dbt State also rebuilds when it can't determine source freshness, regardless of SQL changes. This behavior is intentional: when it's unclear if the data has changed, dbt State prefers rebuilding the model (the same behavior as running without dbt State) over skipping a build that might be necessary.
```

website/docs/docs/deploy/dbt-state-migration.md
luna-bianca: Added here: https://github.com/dbt-labs/docs.getdbt.com/pull/9577/commits/8698db8bd20c0d55213bae121345d5821b76bb99

website/docs/docs/deploy/dbt-state-about.md
```diff
@@ -33,7 +33,7 @@ dbt State delivers efficiency gains across both production and development envir
...
-- **Reuse node from different schema (clone)** — If the same object exists with matching logic and fresh data, dbt State clones it. The node is marked as **Reused** at a fraction of the compute cost.
+- **Reuse node from different schema (clone)** — dbt State looks across all environments and jobs for a matching object with identical logic and fresh data. This includes schemas where a model was built before it ever ran in production. When multiple candidates exist, dbt State clones from the one with the freshest data, regardless of which environment it came from. For example, if a CI schema has fresher data than production, dbt State clones from there. The node is marked as **Reused** at a fraction of the compute cost.
```
joellabes:
```suggestion
- **Reuse node from different schema (clone)** — dbt State looks across all environments and jobs for a matching object with identical logic and fresh data. This includes schemas where a model was built before it ever ran in production. When multiple candidates exist, dbt State clones from the one with the freshest data, regardless of which environment it came from. For example, if a CI schema has fresher data than production and identical logic, dbt State clones from there. The node is marked as **Reused** at a fraction of the compute cost.
```

You say this in the first sentence, but I want to drive it home for people who are skimming

website/docs/faqs/State/views-rebuilt.md
joellabes: >If anything changes that compiled SQL between runs, even when the logic hasn't meaningfully changed, dbt State rebuilds the model

This sounds counter to the idea of hashing the AST instead of the raw code. Any logic change will be picked up, but compiled SQL changes that don't impact logic will be ignored. (This handwaves over a lot, but is more accurate than not)

I'd say something like this: 

> dbt State decides whether to reuse a model by parsing the rendered SQL into a syntax tree and comparing the hash. If the hash has changed (implying the model's logic has changed), dbt State rebuilds the model.
> 
> dbt State prioritises safety and precision; if it can't guarantee skipping a node is safe, then it will re-execute to be sure. A few patterns that commonly cause overeager rebuilds are listed below, along with recommendations to increase reuse rate.

website/docs/faqs/State/views-rebuilt.md
joellabes: It's avoiding an error - if you have this view 

```
select * from some_table
```

and some_table gains a column, then when the view is queried it will cause an error.

joellabes (changes requested):

website/docs/faqs/State/views-rebuilt.md
luna-bianca: Thanks @joellabes ! Applied some editorial changes here:

```suggestion
dbt State decides whether to reuse a model by parsing the rendered SQL into a syntax tree and comparing the hash. If the hash has changed (implying the model's logic has changed), dbt State rebuilds the model.

dbt State prioritizes safety and precision; if it can't guarantee skipping a node is safe, then it rebuilds the node to be sure. A few patterns that commonly cause overeager rebuilds are listed on this page, along with recommendations to increase reuse rate.
```

website/docs/faqs/State/views-rebuilt.md
luna-bianca: @joellabes Clarified here:

```suggestion
dbt State reuses a model when its compiled SQL matches the stored hash. For views with `select *`, dbt State can't determine which columns the query selects without querying the upstream schema, so it can't confirm the SQL is unchanged. It always rebuilds these views to avoid errors &mdash; if the upstream table gains a column, querying the view can fail. When dbt State rebuilds a view, it also re-runs any tests defined on the model.
```

website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -151,6 +151,7 @@ State-aware orchestration and dbt State differ in a few ways:
...
   - **Views with `select *`**: dbt State can't determine which columns `select *` resolves to without querying the upstream schema, so it always rebuilds these views rather than risk reusing a stale result.
   - **Non-determinism in Jinja-templated SQL**: Macros like `dbt_utils.get_relations_by_pattern` with `dbt_utils.union_relations` can return relations in a different order on each run, which produces different compiled SQL. dbt State detects a new hash and rebuilds the model. If that model has downstream dependencies, those models rebuild, too.
+  - **Models with external sources on BigQuery**: Models that use external sources (such as Google Sheets) always rebuild because BigQuery doesn't expose modification timestamps for external sources, so dbt State can't determine freshness. To avoid this, configure [`loaded_at_field`](/reference/resource-properties/freshness#loaded_at_field) or [`loaded_at_query`](/reference/resource-properties/freshness#loaded_at_query) in your source definition to point to a timestamp field &mdash; this lets dbt State query a timestamp field directly to determine freshness, instead of relying on warehouse metadata.
```
mirnawong1: sugg breaking this up so the 'avoid' portion is easier to find
```suggestion
  - **Models with external sources on BigQuery**: Models that use external sources (such as Google Sheets) always rebuild because BigQuery doesn't expose modification timestamps for external sources, so dbt State can't determine freshness. 
  
  To avoid this, configure [`loaded_at_field`](/reference/resource-properties/freshness#loaded_at_field) or [`loaded_at_query`](/reference/resource-properties/freshness#loaded_at_query) in your source definition to point to a timestamp field &mdash; this lets dbt State query a timestamp field directly to determine freshness, instead of relying on warehouse metadata.
```

mirnawong1 (approved):
looks great! left one small sugg and approved!

joellabes (approved):

PR #9579, opened 2026-07-09
docs(cost-insights): document Snowflake query-attribution limitations

MitchellBarker:

Cost Insights attributes Snowflake cost at the query grain via Snowflake's [`QUERY_ATTRIBUTION_HISTORY`](https://docs.snowflake.com/en/sql-reference/account-usage/query_attribution_history#usage-notes) view. Per Snowflake's own usage notes, that view **excludes** two material classes of spend:

- **Short-running queries** (roughly 100 ms or less)
- **Queries run on Adaptive Warehouses** (not present in the view at all)

Because of this, Cost Insights totals are typically **lower** than a customer's Snowflake billing dashboards. These gaps are known internally (raised in a customer feedback session as a "$39K vs $44K" trust concern, and re-confirmed recently while debugging an Adaptive-warehouse cost sawtooth) but were **surfaced nowhere customer-facing**. Worse, the page asserted "Query attribution data is always available for Snowflake," which directly contradicts Snowflake's usage notes.

This PR:
- **Removes** the inaccurate "always available for Snowflake" sentence.
- **Adds** a `:::info Snowflake attribution limitations` callout to the Snowflake cost-calculation section (short queries, Adaptive Warehouses, non-execution costs, data availability), matching the existing `:::info` precedent in the Redshift section.
- **Extends** the "Why might my actual warehouse costs differ from displayed costs?" FAQ with the Snowflake-specific exclusions.

### Design decisions
- Removed the "always available" claim rather than softening it, since leaving it alongside a limitations callout would be self-contradictory.
- Kept the "~100 ms" figure approximate, matching Snowflake's own hedged wording ("<= ~100ms").

## Checklist

> [!NOTE]
> **Open item before merge:** the Adaptive Warehouse bullet reflects behavior as of today. If an in-flight fix sources Adaptive-warehouse queries from `QUERY_METERING_HISTORY` (tracked under META-7919), that bullet may need to be removed or reworded. Please confirm with the META-7919 owner / PMM before publishing.

### DIFF
--- a/website/docs/docs/explore/cost-insights.md
+++ b/website/docs/docs/explore/cost-insights.md
@@ -61,7 +61,7 @@ The following sections explain how costs are calculated for each supported wareh
 
 <Expandable alt_header="Snowflake">
 
-dbt computes Snowflake query costs using Snowflake's query attribution data and your credit price (`price_per_credit`). Query attribution data is always available for Snowflake. dbt pulls the `per_credit` price directly from Snowflake when available; otherwise, dbt uses the configured or default value in the <Constant name="dbt_platform" />. For more information about configuring or viewing these values, see [Configure Cost Insights settings](/docs/explore/set-up-cost-insights#configure-cost-insights-settings-optional).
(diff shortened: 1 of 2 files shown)


MitchellBarker:
Superseded by #9580 — recreated on a shorter branch (`mb/ci-snowflake-limits`) so the Vercel preview alias fits the 63-char DNS label limit; the old branch name broke the update-pr-description check.

PR #9729, opened 2026-07-30
Scroll docs sidebar to the active page on load

### LINKED ISSUE #9728: Left sidebar doesn't scroll the active page into view on load
## What's the issue?

When you land on a deeply nested docs page — via search, an external link, or a shared URL — the left sidebar renders scrolled to the top instead of to the page you're actually on. You have to manually scroll the sidebar a long way down to see where you are in the information architecture, or to navigate to a sibling page.

Docusaurus already *expands* the ancestor categories of the active page, so the correct branch of the tree is open. It just never scrolls the sidebar's scroll container to bring the active item into view.

## Steps to reproduce

1. Open <https://docs.getdbt.com/docs/build/iceberg/catalogs-yml> in a fresh tab.
2. Look at the left sidebar.

**Expected:** the sidebar shows the `Build dbt projects` → `Enhance your models` → `Apache Iceberg` → `Using catalogs.yml` branch, with the current page visible.

**Actual:** the sidebar is scrolled to the top, showing `About` / `Get started`. The active item is far below the fold.

Measured on that page at a 1440x900 viewport:

| Property | Value |
| --- | --- |
| Sidebar scroll container | `nav.menu.thin-scrollbar` (from `@theme/DocSidebar/Desktop/Content`) |
| `scrollHeight` | 2653px |
| `clientHeight` | 595px |
| `scrollTop` on load | **0** |
| Active link (`a.menu__link[aria-current="page"]`) offset | 1529px below the top of the container |

So the active item sits ~2.5 viewports below what's rendered.

This gets worse the deeper a page sits in the tree, and our sidebar is both deep and long, so it affects a large share of our pages.

## Upstream context

This is a known, unfixed gap in Docusaurus rather than a misconfiguration on our side. There is no `themeConfig.docs.sidebar` option for it:

- facebook/docusaurus#7980 — "Scroll to selection on sidebar when visiting with a direct link"
- facebook/docusaurus#10615 — "Sidebar not scrolling to open page"
- facebook/docusaurus#823 — "Sidebar scroll position" (original long-running thread)
- facebook/docusaurus#6058 and facebook/docusaurus#9904 — the same class of problem for the right-hand table of contents

Because there's no config knob, fixing it requires a theme component override (a Docusaurus "swizzle").

## Proposed fix

Add a **wrap**-style swizzle at `website/src/theme/DocSidebar/Desktop/Content/index.js` that renders the original Docusaurus `Content` component and, once on mount, scrolls the sidebar's scroll container so the active link is vertically centered.

Notes on the approach:

- Set `scrollTop` on the scroll container directly rather than calling `element.scrollIntoView()`. `scrollIntoView()` walks *every* scrollable ancestor and would also scroll the main document, moving the reader away from the top of the article.
- Run the scroll after paint (`requestAnimationFrame`). Our `website/src/theme/DocSidebarItem/index.js` override returns `null` for items that don't apply to the selected dbt version, so the sidebar's final height isn't known at first commit.
- Only fire on initial mount, not on every route change. Within-site navigation should leave the sidebar scroll where the reader put it — re-centering on each click would fight the user.
- Use a wrap rather than an eject so we don't fork upstream's component and can keep inheriting its fixes.
- No-op when the active link is already visible, and skip on mobile (where the sidebar is a separate drawer component).

dataders:

Closes #9728.

Landing on a deeply nested docs page — from search, a shared link, or an external referrer — rendered the left sidebar scrolled to the top, so readers had to scroll a long way down just to see where they were in the tree.

Docusaurus already *expands* the ancestor categories of the active page. It just never scrolls the sidebar's scroll container to bring the active item into view, and there's no `themeConfig.docs.sidebar` option to turn it on. It's a known upstream gap: facebook/docusaurus#7980, facebook/docusaurus#10615, facebook/docusaurus#823.

On `/docs/build/iceberg/catalogs-yml` at 1440x900, the sidebar scroll container (`nav.menu`) is 2653px of content in a 775px window, and the active link sits 1529px down — about two viewports below what's rendered.

### The change

One new file: `website/src/theme/DocSidebar/index.js`, a **wrap** swizzle of `@theme/DocSidebar` that centers the active link in the sidebar once on mount.

Notes on the approach, since a few details are load-bearing:

- **Wrap, not eject.** Upstream marks `DocSidebar` as `eject: unsafe` / `wrap: safe`. This copies no upstream code and adds no markup, so if Docusaurus restructures the sidebar the effect degrades to a no-op rather than breaking the layout.
- **Sets `scrollTop` directly** instead of calling `link.scrollIntoView()`. `scrollIntoView()` walks *every* scrollable ancestor and would also scroll the main document, moving the reader off the top of the article.
- **Runs synchronously in `useEffect`, not in `requestAnimationFrame`.** rAF never fires while a tab is hidden, so an rAF version silently did nothing for pages opened in a background tab (cmd+click). The DOM is already committed when the effect runs, so the measurement is valid without waiting for a frame.
- **Finds the scroll container by walking up from the active link**, bounded by `.theme-doc-sidebar-container`, rather than hardcoding a class. The container is Infima's `nav.menu`, but the bounded walk means we can never accidentally scroll the document, and it survives class renames.
- **Matches `a.menu__link[aria-current="page"]`.** Ancestor categories also carry `menu__link--active`, so matching on that alone selects a parent instead of the current page.
- **Mount-only.** Navigating within the site preserves whatever scroll position the reader left the sidebar in; re-centering on every route change would fight the user.
- **No-ops when the active link is already visible**, so shallow pages are untouched.
- Desktop only. The mobile sidebar is a separate component in the navbar's secondary menu, and is already at the top when opened.

### Verification

Tested locally against the dev server at 1440x900:

| Scenario | Result |
| --- | --- |
| Fresh load of `/docs/build/iceberg/catalogs-yml` | `scrollTop` 0 → 1160, active link fully visible, `window.scrollY` still 0 |
| Same, in a **hidden** tab | Same result (this is what the rAF version got wrong) |
| Fresh load of `/docs/introduction` (shallow) | No-op, `scrollTop` stays 0 |
| Sidebar click to a sibling page | Reader's scroll position preserved, no re-center |

Before / after on `/docs/build/iceberg/catalogs-yml`: the sidebar previously opened on `About` / `Get started`; it now opens with `Apache Iceberg` → `Using catalogs.yml` in view.

`npx eslint` clean.

## Checklist
- [ ] ~Applied the proper versioning rules~ — n/a, no content changes.
- [ ] ~Requires a dbt release note~ — n/a, site behavior only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JKarlavige (approved):
lgtm! nice improvement 🚀

mirnawong1:
works beautifully! thanks so much @dataders !

PR #9755, opened 2026-08-06
Add on_schema_change behavior table for incremental models

### LINKED ISSUE #2902: Explain how `on_schema_change` handles new or removed columns for an incremental model
### Contributions

https://docs.getdbt.com/docs/build/incremental-models#what-if-the-columns-of-my-incremental-model-change

@Fleid noted [here](https://github.com/dbt-labs/dbt-adapters/issues/601) that the following is a more complete and accurate description of how the value of `on_schema_change` behaves for incremental models when:
1. new columns are added, and
2. when columns are removed

|`on_schema_change`|In case of new columns...|In case of removed columns...|
|-|-|-|
|`ignore`|columns ignored|run failed|
|`fail`|run failed|run failed|
|`append_new_columns`|columns added|columns ignored|
|`sync_all_columns`|columns added|columns removed|

### Additional information

_No response_

krishn1301:

Resolves #2902.

The `on_schema_change` section lists what each value does, but the descriptions are spread across a bullet list, a note, and a "Default behavior" subsection, and they don't cover all three kinds of schema change side by side. This adds a summary table so you can see, for one value, what happens when a column is added, removed, or retyped.

The table reflects the corrections from @luna-bianca's testing and @itsnamangoyal's review:

- Only `ignore` skips detection of a data type change, because `default__process_schema_changes` returns before `check_for_schema_changes` runs. `append_new_columns` detects the change and doesn't apply it, since `default__sync_column_schemas` only calls `alter_column_type` under `sync_all_columns`.
- `sync_all_columns` issues an `ALTER` whose outcome is platform-dependent, so the table says the `ALTER` is attempted rather than promising the type changes.
- With `append_new_columns` and a removed column, existing values are kept under `merge` and lost under `delete+insert`.
- Added a note that the table assumes rows are matched on a `unique_key`, which is not the default on Databricks or Athena, and a note that string widening can happen before `on_schema_change` is evaluated.

No new pages, no versioned content, no release note needed.

## Checklist


website/docs/docs/build/incremental-models.md
```diff
@@ -213,6 +213,15 @@ The possible values for `on_schema_change` are:
 - `append_new_columns`: Append new columns to the existing table. Note that this setting does *not* remove columns from the existing table that are not present in the new data.
 - `sync_all_columns`: Adds any new columns to the existing table, and removes any columns that are now missing. Note that this is *inclusive* of data type changes. On BigQuery, changing column types requires a full <Term id="table" /> scan; be mindful of the trade-offs when implementing.
 
+The following table summarizes what each value does when you add a column to your model, remove a column from your model, or change a column's data type:
```
luna-bianca, line 216:
```suggestion
The following table summarizes what each value does when you add a column, remove a column, or change a column's data type:
```

luna-bianca (commented):
Thanks for opening this PR, @krishn1301 !

I did some testing to verify the behavior:

| `on_schema_change` | Column added | Column removed | Column data type changed |                               
  |---|---|---|---|                      
  | `ignore` | Run succeeds, column not added | Run fails | Run succeeds, type unchanged; the incoming value is implicitly cast to the target column's existing type (fails if the cast is not possible)  |                                                                                                     
  | `fail` | Run fails | Run fails | Run fails |                                                                  
  | `append_new_columns` | Run succeeds, column added (`null` for existing rows) | Run succeeds, column stays; new rows get `null` | Run succeeds, type unchanged; the incoming value is implicitly cast to the target column's existing type (fails if the cast is not possible)  |                            
  | `sync_all_columns` | Run succeeds, column added (`null` for existing rows) | Run succeeds, column dropped |Macro issues an `ALTER` but outcome is platform-dependent: fails on DuckDB, fails on Snowflake for incompatible conversions (e.g. VARCHAR→NUMBER), succeeds for compatible ones (e.g. widening VARCHAR) |

I'll check with our engineers if my findings are accurate since there are behaviors that are not reflected by the contents of the PR:

- `sync_all_columns` + data type changes: the PR says "the data type is altered on the target table," but in testing this is platform-dependent. The macro does issue an ALTER, but it fails on DuckDB (transaction conflict) and fails on Snowflake for incompatible conversions (VARCHAR→NUMBER), while succeeding for compatible ones (widening a VARCHAR).
- `append_new_columns` + column removed: the PR says "existing values are kept," which seems to be true for insert-only models. With a `unique_key`, matched rows are updated and the removed column is set to null.

itsnamangoyal:
Thanks for the contribution @krishn1301. I checked each cell against the dbt-adapters source. A few fixes needed before merge (adding to @luna-bianca's review above):

1. `append_new_columns` + data type changed

The description says this "doesn't detect a data type change at all." That's wrong. Only `ignore` skips detection completely. `append_new_columns` does detect the type change — it just doesn't apply it. The table cell already says this correctly ("Not applied, type stays the same"). Only the paragraph above the table needs fixing.

2. `sync_all_columns` + data type changed

This isn't a flat "the data type is altered." It depends on the cast:
- Incompatible cast (e.g. VARCHAR → NUMBER): the run fails.
- Compatible cast (e.g. widening a NUMBER's precision): it succeeds.

Suggested wording: "An ALTER is attempted. It succeeds or fails depending on whether the platform allows that type conversion."

Separate note, string columns only: widening a VARCHAR's length happens automatically, before `on_schema_change` is even checked. So that specific case would work under ignore or fail too — it's not really `sync_all_columns` doing it. This doesn't apply to numbers or other types; those only change if `on_schema_change` is set to `sync_all_columns`.

3. `append_new_columns` + column removed (with unique_key)

The PR says old values are kept. Bianca's review says they get set to null. Both are right — just for different setups:
- Default on Snowflake and BigQuery (`merge` strategy): old values are kept.
- Default on Postgres and Redshift (`delete+insert` strategy): old values are lost, become null.

Suggested wording: "Kept if using the `merge` strategy (default on Snowflake, BigQuery). Lost (becomes null) with `delete+insert` (default on Postgres, Redshift)."

4. One more caveat worth adding:

This table assumes setting `unique_key` always makes rows match up during an incremental run. That's true on Snowflake and BigQuery, but not everywhere — on Databricks the default is plain `append` (no matching, unless you set incremental_strategy: `merge` yourself), and Athena defaults to `insert_overwrite`. Worth a short note that this table covers the common merge-based case, not every adapter's default.

krishn1301:
Thanks both — you were right on all four, and I've pushed the corrections.

I re-read `on_schema_change.sql` after @itsnamangoyal's note and the detection point is clear: `default__process_schema_changes` returns immediately for `ignore`, so `check_for_schema_changes` never runs. For every other value it does run, and it computes `new_target_types` via `diff_column_data_types`. So `append_new_columns` detects the type change; it just never reaches `alter_column_type`, because `default__sync_column_schemas` only loops over `new_target_types` under `sync_all_columns`. My PR description had that wrong and is now fixed.

What changed in the table:

- `sync_all_columns` + type changed: now "An `ALTER` is attempted. It succeeds or fails depending on whether your data platform allows that type conversion." That matches @luna-bianca's DuckDB and Snowflake results without naming specific platforms in a cell.
- `append_new_columns` + type changed: now "Detected, but not applied."
- `append_new_columns` + column removed: now splits on strategy — kept under `merge` (default on Snowflake, BigQuery), lost under `delete+insert` (default on Postgres, Redshift). That reconciles the two results.
- Added a note that the table assumes rows are matched on a `unique_key`, and that Databricks defaults to `append` and Athena to `insert_overwrite`.

I also added @itsnamangoyal's string-widening caveat as a separate sentence, since it applies regardless of `on_schema_change` and would otherwise look like a contradiction of the `ignore` row.

One open question: I kept the platform names in prose rather than in table cells to stop the cells growing further. If you'd rather the strategy split lived in the cell itself, or in a footnote, happy to move it.

krishn1301:
Hi @luna-bianca @itsnamangoyal — gentle nudge on this one. All four corrections from your reviews went in on Aug 13 (summary in my comment above), and I've just checked that `incremental-models.md` on `current` still has no per-value behaviour table, so it isn't overtaken by anything upstream.

One question from that comment is still open: do you want the `merge` / `delete+insert` split kept in prose, or moved into the table cell or a footnote? I'm happy to make either change. Otherwise this should be ready for another look whenever you have time.

luna-bianca:
Thanks for following up, @krishn1301 ! We'll take a look at this ASAP!

Issue #9339, opened 2026-06-05
Markdown no longer shows the proper compiled page


joshuataylor:
I use the compiled Markdown to provide documentation for Datamancer, a plugin for JetBrains IDEs (PyCharm, IntelliJ, etc) which provides dbt support.

## Bug

1. Go to https://docs.getdbt.com/docs/build/documentation?version=1.11&name=Core 
2. Click "Copy as Markdown" (or go to https://docs.getdbt.com/docs/build/documentation.md)
3. We see:

```
import CopilotBeta from '/snippets/_dbt-copilot-avail.md';

Good documentation for your dbt models will help downstream consumers discover and understand the datasets you curate for them.
dbt provides a way to generate documentation for your dbt project and render it as a website.


<CopilotBeta resource='documentation' />
```

and:
```
### FAQs
<FAQ path="Project/example-projects" alt_header="Are there any example dbt documentation sites?"/>
```
etc.

## Expected result

The full page.

## When did this break?

Unsure. I swear it use to work.

mirnawong1:
hey @joshuataylor ! thanks for raising this and i've shared this internally with our eng team! it's copying the markdown, including partials we use for reusable content. 

did you say you use to see the full markdown?

joshuataylor:
Yes! I also managed to figure this out a while ago, and forgot.

I've got a PR which fixes it, coming now :-).

PR #9765, opened 2026-08-07
Generate compiled per-page markdown from rendered HTML

joshuataylor:
Follow-up to the discussion in #9339 (thanks @mirnawong1!).

I've been pointing tooling at the per-page `.md` files and kept getting output that was missing half its content, so I dug in and traced it to the generator wiring, I've been using this for the last few months to build https://github.com/joshuataylor/dbt-docs-markdown .

## The underlying issue

The `.md` you get for any doc page today (and what the "Copy page" button hands out) is the raw MDX source with only the frontmatter stripped. Nothing inlines the partials or renders the components, so:

- `import CopilotBeta from '/snippets/_dbt-copilot-avail.md';` is left verbatim and the partial body never appears.
- `<FAQ path=.../>`, `<File>`, `<VersionBlock>`, `<Term>` pass through as literal tags -- everything that lives inside them is missing. That's why `### FAQs` sections show up empty.
- `?version=` is ignored, since `<VersionBlock>` is never resolved.

It's `plugins/buildRawMarkdownData`, it writes the static `.md` straight from source, while the HTML-deriving generator (`@signalwire/docusaurus-plugin-llms-txt`, already a dependency -- its README: `processes your final HTML output ... not your source`) had `enableMarkdownFiles: false` so the two wouldn't write the same output paths.

## proposed fix

Two commits:

1. Generate compiled per-page markdown from rendered HTML. Change `enableMarkdownFiles` to `true` so the llms-txt plugin owns the on-disk `.md` (partials inlined, components resolved). `buildRawMarkdownData` keeps its `loadContent`/`contentLoaded` global data -- the in-browser "Copy page" button reads it via `usePluginData` and still works exactly as before (its client-side `<VersionBlock>` filtering included) -- but its `postBuild` file writer, the source of the collision, is removed.

2. Rewrite generated markdown links to be document-relative. The llms-txt plugin emits cross-doc links as absolute `https://docs.getdbt.com/<path>.md` URLs, and its `relativePaths` option only produces root-relative paths. A small `postbuild` npm script rewrites them to document-relative paths when the target `.md` exists in the build, and leaves them absolute otherwise so they stay valid against the live site.

One non-obvious bit worth calling out: the rewriter can't be a sibling Docusaurus plugin ordered "after" llms-txt. Docusaurus runs every plugin's [postBuild](https://docusaurus.io/docs/api/plugin-methods/lifecycle-apis#postBuild) concurrently (`Promise.all` in `executePluginsPostBuild`, `@docusaurus/core` `commands/build/buildLocale.ts`), so a plugin-based rewriter races the file writer and silently rewrites nothing -- I found this out the hard way :-)........... Running it as an npm `postbuild` step (the repo already uses `prebuild`, so this matches convention) guarantees the files exist. 

## Before / after

`build/reference/commands/lint.md`, before (raw source, components unrendered):

```markdown
# About dbt lint command <Lifecycle status="beta" />
...
`dbt lint` is part of the <Constant name="fusion_engine" />. It is not the same as
`dbt sqlfluff lint` on the <Constant name="platform_cli" />. For SQLFluff on the
platform CLI, see [Configure the dbt CLI](/docs/platform/configure-dbt-cli).
```

After (compiled, links relative):

```markdown
# About dbt lint command [Beta](...)
...
`dbt lint` is part of the dbt Fusion engine. It is not the same as
`dbt sqlfluff lint` on the dbt CLI. For SQLFluff on the platform CLI, see
[Configure the dbt CLI](../../docs/platform/configure-dbt-cli.md).
```

---

Please never get me to write JS again, thanks :cry:.

Happy for this to be reworked however you see fit -- if you'd rather keep the raw-source files under a different path alongside the compiled ones, or split the link rewriter out, take whatever's useful - I just want a solution and ain't tied to any particular fix.


joshuataylor:
Happy to test once vercel runs, i'm curious if it works there.

runleonarun:
Hey @joshuataylor I approved those workflows and the deployment to Vercel.

JKarlavige (approved):
lgtm, thanks for getting this update in @joshuataylor! Spot-checked a number of pages on the live site and compared them to the preview, and verified the full content is showing in the preview md files vs. having the components in the live md files. :shipit:

PR #9783, opened 2026-08-11
Remove duplicated entries in dbt-versions

joshuataylor:

Remove duplicate entries in website/dbt-versions.js. Probably due to a wonky rebase etc.


mirnawong1 (approved):
great catches @joshuataylor and removing the dups! cc @matthewshaver as a heads up since it touches your pages

PR #9795, opened 2026-08-12
dbt State model concurrency

luna-bianca:
Closes https://dbtlabs.atlassian.net/browse/PRODDOCS-1256

Documents a known gap in dbt State: unlike state-aware orchestration, dbt State doesn't detect when the same model is already being built by another concurrent job. Both jobs run independently, resulting in duplicate compute (though no data corruption). Adds a new "No concurrent build detection" entry to the Known differences section of the migration guide, including the workaround of configuring job schedules or avoiding overlapping manually triggered runs.

Source: https://dbt-labs.slack.com/archives/C0B0L9R47HB/p1786483730719469?thread_ts=1781736611.990089&cid=C0B0L9R47HB

## Checklist

- [ ] The content in this PR requires a dbt release note, so I added one to the [release notes page](https://docs.getdbt.com/docs/dbt-versions/release-notes).).

---

- https://docs-getdbt-com-git-dbt-state-model-concurrency-dbt-labs.vercel.app/docs/deploy/dbt-state-migration
- https://docs-getdbt-com-git-dbt-state-model-concurrency-dbt-labs.vercel.app/docs/deploy/state-aware-about


website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -161,6 +161,9 @@ State-aware orchestration and dbt State differ in a few ways:
 - **`build_after` vs `lag_tolerance`**: Both configs reduce how often a model runs when upstream data is frequently fresh, but they work differently:
   - `freshness.build_after` (for example, `{count: 4, period: hour}`) skips the model unless the configured interval has elapsed _and_ upstream sources have new data since the last run. A SQL change alone does not trigger a rebuild; both conditions must be met.
   - `state.lag_tolerance` (for example, `4h`) skips the model unless upstream data is newer than the model's last run by at least the configured interval. Unlike `build_after`, a detected SQL change triggers a rebuild.
+- **No concurrent build detection**: In state-aware orchestration, if a model was being built by one job, any other job that included the same model would skip it. dbt State doesn't have this mechanism; if two jobs include the same model and run at the same time (for example, an hourly and a daily job that overlap), both jobs build the model independently. This doesn't cause data corruption because the warehouse handles concurrent writes, but it does result in duplicate compute.
```
joellabes: >This doesn't cause data corruption because the warehouse handles concurrent writes, but it does result in duplicate compute. 

This _can_ cause data corruption in snapshots and incremental models if they execute simultaneously on two jobs.

joellabes (changes requested):

website/docs/docs/deploy/dbt-state-migration.md
luna-bianca: Fixed [here](https://github.com/dbt-labs/docs.getdbt.com/pull/9795/commits/8e1dbb944692c072a1b6a6884da4c1a441d7efc5)

website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -161,6 +161,9 @@ State-aware orchestration and dbt State differ in a few ways:
 - **`build_after` vs `lag_tolerance`**: Both configs reduce how often a model runs when upstream data is frequently fresh, but they work differently:
   - `freshness.build_after` (for example, `{count: 4, period: hour}`) skips the model unless the configured interval has elapsed _and_ upstream sources have new data since the last run. A SQL change alone does not trigger a rebuild; both conditions must be met.
   - `state.lag_tolerance` (for example, `4h`) skips the model unless upstream data is newer than the model's last run by at least the configured interval. Unlike `build_after`, a detected SQL change triggers a rebuild.
+- **No concurrent build detection**: In state-aware orchestration, if a model was being built by one job, any other job that included the same model would skip it. dbt State doesn't have this mechanism; if two jobs include the same model and run at the same time (for example, an hourly and a daily job that overlap), both jobs build the model independently. This results in duplicate compute and, for snapshots and incremental models, can cause data corruption. 
+
```
reubenmc, line 165: I don't think this isn't correct. This isn't really about duplicate compute. SAO would also have one model build, and then see if the other one needed building. In 99.9% of the time, State will do the exact same thing. Two hourly jobs can run the same model, job 1 will build it, and job 2 will see it's already been built and reuse it. This is only about the edge case where job 1 and job 2 hit the model at the exact same time so job 1 hasn't yet pushed back to state to say that the model has been updated.

luna-bianca: Addressed here: https://github.com/dbt-labs/docs.getdbt.com/pull/9795/commits/42a9a33614b3f15c8172cbbd20ced56af549a6f0

website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -161,6 +161,9 @@ State-aware orchestration and dbt State differ in a few ways:
 - **`build_after` vs `lag_tolerance`**: Both configs reduce how often a model runs when upstream data is frequently fresh, but they work differently:
   - `freshness.build_after` (for example, `{count: 4, period: hour}`) skips the model unless the configured interval has elapsed _and_ upstream sources have new data since the last run. A SQL change alone does not trigger a rebuild; both conditions must be met.
   - `state.lag_tolerance` (for example, `4h`) skips the model unless upstream data is newer than the model's last run by at least the configured interval. Unlike `build_after`, a detected SQL change triggers a rebuild.
+- **No concurrent build detection**: In most cases, dbt State handles overlapping jobs the same way as state-aware orchestration: Job 1 builds the model, and Job 2 sees that the model has already been built and reuses it. However, if both jobs reach the same model at the same time, before Job 1 records in state that the model is complete, both jobs can build the model independently. For snapshots and incremental models, this can lead to data corruption.
```
joellabes: Models are excluded from execution at the start of the job. So as long as Job 1 has finished building the snapshot before Job 2 starts, then it will be completely skipped/reused. If Job 1 finishes building the snapshot after Job 2 starts, but before Job 2 gets to that snapshot, then it will waste some compute re-determining that there's no new data, but won't corrupt data. 

The risk scenario is if Job 2 starts building the snapshot before Job 1 has finished. Then both invocations will detect new/modified records inside of their standalone transactions, and both changesets will be committed.

joellabes (dismissed):
This is accurate as-is but could be more precise. Happy for it to be merged without a second review loop from me once this is incorporated

website/docs/docs/deploy/dbt-state-migration.md
luna-bianca: Hey @joellabes , let me know if this correctly captures the behavior:

```suggestion
- **No concurrent build detection**: In most cases, dbt State handles overlapping nodes the same way as state-aware orchestration: if a node is built by Job 1 before Job 2 starts, Job 2 sees the model as already built and reuses it. If Job 1 finishes building a node after Job 2 starts &mdash; but before Job 2 builds that node &mdash; Job 2 still builds it and uses compute to re-determine there's no new data. However, if Job 2 starts building the same snapshot or incremental model before Job 1 executing the same node, both jobs can detect the same changes in their separate transactions and commit them, which can lead to duplicate records or other data corruption.
```

joellabes: I've edited to clarify - it's at the node level, not whether the whole job has completed. This is good to go now

website/docs/docs/deploy/dbt-state-migration.md
```diff
@@ -161,6 +161,9 @@ State-aware orchestration and dbt State differ in a few ways:
...
+- **No concurrent build detection**: In most cases, dbt State handles overlapping nodes the same way as state-aware orchestration: if a node is built by Job 1 before Job 2 starts, Job 2 sees the model as already built and reuses it. If Job 1 finishes building a node after Job 2 starts &mdash; but before Job 2 builds that node &mdash; Job 2 still builds it and uses compute to re-determine there's no new data. However, if Job 2 starts building the same snapshot or incremental model before Job 1 executing the same node, both jobs can detect the same changes in their separate transactions and commit them, which can lead to duplicate records or other data corruption.
+
+  To prevent duplicate builds, ensure jobs that share the same models don't run at the same time &mdash; whether by configuring job schedules through the [<Constant name="dbt_platform" /> job scheduler](/docs/deploy/job-scheduler) or an external orchestration tool (for example, Airflow or Dagster), or by avoiding manually triggered runs that overlap.
```
luna-bianca:
```suggestion
  To prevent duplicate builds, ensure jobs that share the same nodes don't run at the same time &mdash; whether by configuring job schedules through the [<Constant name="dbt_platform" /> job scheduler](/docs/deploy/job-scheduler) or an external orchestration tool (for example, Airflow or Dagster), or by avoiding manually triggered runs that overlap.
```

matthewshaver (approved):

PR #9803, opened 2026-08-13
Update version picker: v1 selectable as group with checkmark

luna-bianca:
Updates the version picker so that 'v1' behaves as a first-class selectable option, similar to 'v2'.

## Changes

- **'v1' is now a selectable group header** — clicking 'v1' selects it with a checkmark, just like 'v2'. Previously, clicking 'v1' would silently auto-select 'dbt Core v1.12'.
- **'dbt Core v1.12' and 'dbt Core v1.11' remain as indented sub-options** for selecting a specific minor version.
- **URL shows `?version=1`** when 'v1' is selected (not `?version=1.12`). Specific version selections still show `?version=1.12` or `?version=1.11`.
- **Content filtering**: selecting 'v1' shows all v1 content and hides v2-only content.
- **Auto-updates when a new v1.x is released** — the 'v1' group entry in `dbt-versions.js` derives its version dynamically from the first entry in the v1 subProducts list, so adding 'dbt Core v1.13' at the top automatically updates what 'v1' points to.


JKarlavige (approved):
lgtm!

PR #9809, opened 2026-08-14
Render VersionBlock content during SSR so versioned tables appear in generated markdown

mirnawong1:

Follow-up to #9765 and #9769.

Versioned content is missing from the generated per-page `.md` files. For example, [docs/build/conversion.md](https://docs.getdbt.com/docs/build/conversion.md) has the "The specification for conversion metrics is as follows:" line but no parameter table or spec after it — the two `<VersionBlock>` tables and the YAML spec blocks are all gone. Roughly every page that uses `<VersionBlock>` is affected.

### What's happening

Since #9765, the `.md` files are generated from the rendered HTML (via `@signalwire/docusaurus-plugin-llms-txt`). Anything not present in the static HTML never makes it into the `.md`.

`VersionBlock` gated its children behind a client-only `useEffect`:

```js
const [loading, setLoading] = useState(true);
useEffect(() => { version && setLoading(false); }, [version]);
...
return loading ? null : <>{children}</>;
```

`useEffect` never runs during SSR, so `loading` stays `true` and every `VersionBlock` renders `null` in the static HTML — so all versioned content is dropped from the generated markdown. Same class of issue as the `FilterableTable` placeholder fixed in #9769.

### The fix

Drop the `loading` null-gate. `version` already falls back to the default version in `VersionContext`, so the range checks resolve correctly during SSR and the default version's content renders into the static HTML (and the `.md`). The first client render uses the same default, so it matches the server output; the client still swaps to the selected version after mount, exactly as before.

### tl;dr

Versioned tables and code blocks now appear in the generated `.md`. No behavior change for users with JS — the block still updates to the selected version on mount.

> Would love a Vercel preview build to confirm the generated `.md` output, same as #9769.

## Checklist

- [x] The changes in this PR meet the docs style guide/fundamentals required for all content. (Component/theme change, no prose edits.)

---

- https://docs-getdbt-com-git-fix-versionblock-ssr-markdown-dbt-labs.vercel.app/docs/build/about-metricflow
- https://docs-getdbt-com-git-fix-versionblock-ssr-markdown-dbt-labs.vercel.app/reference/global-configs/about-global-configs


JKarlavige (approved):
lgtm! confirmed the content missing from the live markdown file appears within the preview 🚀

mirnawong1:
awesome thanks so much!

PR #9895, opened 2026-08-27
[dbt State] compare_unrendered_code

luna-bianca:
Docs for the new `compare_unrendered_code` dbt State config, which prevents unnecessary rebuilds for nodes that use non-deterministic macros or environment variables.

Woohoo promoted from dbt-labs/docs-internal#2260


runleonarun (approved):
Looks good! Reviewed in internal repo.

PR #9905, opened 2026-08-31
docs(databricks): add missing incremental_predicates to delete+insert example

### LINKED ISSUE #1652: Documentation on delete-insert is not consistent
To be specific, the example is missing the `incremental_predicates`

Image

Image

sd-db:
## Summary
- Fixes the inconsistent `delete+insert` docs on Databricks configs: the source example now includes `incremental_predicates`, matching the prose and the generated run SQL.
- Aligns the example with the existing `replace_where` pattern (`user_id >= 10000`), so the run tabs no longer reuse the model's `is_incremental()` `date_day` filter as if it were an `incremental_predicates` clause (that column is not in the temp view).
- Resolves [databricks/dbt-databricks#1652](https://github.com/databricks/dbt-databricks/issues/1652).

## Test plan
- [ ] Preview the Databricks configs page → **The `delete+insert` strategy** section
- [ ] Confirm Source / Run (DBR 17.1+) / Run (DBR < 17.1) tabs all show the same `incremental_predicates` (`user_id >= 10000`)
- [ ] Confirm the `is_incremental()` `date_day` filter still appears only in the model query / temp view, not as the insert/delete predicate

### DIFF
--- a/website/docs/reference/resource-configs/databricks-configs.md
+++ b/website/docs/reference/resource-configs/databricks-configs.md
@@ -678,7 +678,8 @@ You can optionally use `incremental_predicates` to further filter which records 
     materialized='incremental',
     file_format='delta',
     incremental_strategy='delete+insert',
-    unique_key='user_id'
+    unique_key='user_id',
+    incremental_predicates='user_id >= 10000' # Never delete/insert users with ids < 10000
 ) }}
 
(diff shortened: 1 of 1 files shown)


luna-bianca (approved):
Thanks for opening this PR, @sd-db !

PR #9908, opened 2026-08-31
Pin internal actions to floating major tags (IS-6458)

jeremyhutt11:
Swaps 1 SHA-pinned `dbt-labs/internal-actions` refs to their floating major tags. Policy is major tag for internal actions, full SHA for external ones ([README](https://github.com/dbt-labs/internal-actions#pinning-policy)). Each ref already resolved to the major it's pinned at, so no behaviour change. IS-6458.

- `changed-files-v1`

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Get changed files
         id: changed-files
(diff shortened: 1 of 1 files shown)


ddk-dbt (approved):

runleonarun (approved):
Let's see what this does!

PR #9973, opened 2026-09-09
dbt state lag tolerance docs

luna-bianca:
dbt state lag tolerance docs

Woohoo promoted from dbt-labs/docs-internal#2465

## What changed

Adds docs for the **lag tolerance recommendations** feature on the dbt State page, which surfaces models that could safely tolerate more lag and projects compute savings.

### Changes per file

- **`release-notes.md`** — Adds a "New" entry announcing the lag tolerance recommendations section on the dbt State page.

- **`dbt-state-interface.md`** — Adds a new `## Lag tolerance recommendations` section documenting the recommendations table (columns: model name, project, current lag, recommended lag, % time saved, projected 30d time savings), search/filter options, and how to apply a recommendation by updating the `lag_tolerance` config.

- **`dbt-state-setup.md`** — Imports and renders the `_lag-tolerance-recommendations-tip` snippet after the `lag_tolerance` config details section. Also adds a "Monitor dbt State activity" related doc link.

- **`lag-tolerance.md`** — Imports and renders the `_lag-tolerance-recommendations-tip` snippet after the compiled SQL change caveat. Also adds a "Monitor dbt State activity" related doc link.

- **`dbt-state-about.md`** — Adds a "Monitor dbt State activity" link to the Related docs section.

- **`snippets/_lag-tolerance-recommendations-tip.md`** *(new)* — New reusable snippet with a tip pointing readers to the lag tolerance recommendations on the dbt State page.

---

- https://docs-getdbt-com-git-dbt-state-lag-tolerance-public-dbt-labs.vercel.app/docs/dbt-versions/release-notes
- https://docs-getdbt-com-git-dbt-state-lag-tolerance-public-dbt-labs.vercel.app/docs/deploy/dbt-state-about
- https://docs-getdbt-com-git-dbt-state-lag-tolerance-public-dbt-labs.vercel.app/docs/deploy/dbt-state-interface
- https://docs-getdbt-com-git-dbt-state-lag-tolerance-public-dbt-labs.vercel.app/docs/deploy/dbt-state-setup
- https://docs-getdbt-com-git-dbt-state-lag-tolerance-public-dbt-labs.vercel.app/reference/resource-configs/lag-tolerance


website/docs/docs/deploy/dbt-state-interface.md
```diff
@@ -33,6 +33,27 @@ The **State** tab also displays the following charts:
...
+
+<SimpleTable>
+| Column | Description |
+|--------|-------------|
+| **Model name** | The name of the model that could benefit from a higher `lag_tolerance` value. |
+| **Project** | The dbt project the model belongs to. |
+| **Current lag** | The model's current `lag_tolerance` setting. |
+| **Recommended lag** | The `lag_tolerance` value dbt State recommends based on observed upstream data refresh patterns. |
+| **% time saved** | The estimated percentage of build time you'd save by applying the recommended `lag_tolerance`. |
+| **Projected 30d time savings** | The estimated compute time saved over the next 30 days if you apply the recommended `lag_tolerance`. |
```
luna-bianca: @evabgood @sidpatel13 Is the description for the `Projected 30d time savings` correct?

evabgood: @luna-bianca can we update the language to be "estimated runtime saved over the next 30 days", I think that may be a little clearer, but @sidpatel13 would love your input on how this is being calculated in the backend & what makes most sense here

evabgood (approved):

website/docs/docs/deploy/dbt-state-interface.md
sidpatel13: Yeah so here is an example:
Say you turn on the recommended lag tolerance for Model A, it goes from the 45m default to 2 hrs now. Because it doesn't build for 2 hrs now if there is Model B downstream that relies on A, it may also not build, so the estimate might not be accurate as the compute savings might be more. 

I think we should frame it along the lines of "An estimate of the model build time you'd avoid over the next 30 days by applying the recommended lag_tolerance, based on how often the model rebuilt without its inputs having changed over the last 30 days. It counts this model only. If downstream models stop rebuilding as a result, your actual savings will be higher." Now I will add, I don't know how much of this we should add, but I'll leave the judgement up to you.

website/docs/docs/deploy/dbt-state-interface.md
```diff
@@ -33,6 +33,29 @@ The **State** tab also displays the following charts:
...
+
+The **dbt State** page includes a **Lag tolerance recommendations** section that identifies models that could safely tolerate more lag, letting dbt State skip more runs and save additional compute.
+
+The recommendations table displays the following columns:
+
+<SimpleTable>
+| Column | Description |
+|--------|-------------|
+| **Model name** | The name of the model that could benefit from a higher `lag_tolerance` value. |
+| **Project** | The dbt project the model belongs to. |
```
sidpatel13: We should state somehow that this isn't the platform project, but rather the project that gets set in the dbt_project.yaml. Here is the wording I got from claude by giving it my insight, but feel free to change it best you see fit to convey this message somehow:

"The dbt project the model belongs to, as set by name: in the project's dbt_project.yml. This is the dbt project name, which may differ from the name of the project in the dbt platform."

sidpatel13 (changes requested):
I think the big thing is we should tell them how we pick recommendations. Here is the whole methodology (but we don't need to share it all), just maybe find a more succinct way to share something so users can feel confident on how we pick these:

```
How recommendations are calculated

dbt State evaluates each model against the last 30 days of build history. Models with fewer than 10 recorded builds in that window are excluded, because there is not enough history to draw a reliable conclusion from.

For each remaining model, dbt State identifies redundant builds, meaning builds where the model's definition and inputs were unchanged from the previous build. It then tests a fixed set of candidate lag_tolerance values against those builds: 1, 2, 4, 8, 12, 16, and 24 hours, followed by 2, 4, and 7 days. For each candidate value, it totals the build time that would have been avoided.

The recommendation is the smallest candidate value that would have saved more than 30 minutes of build time across the window. Smaller values are preferred because they keep your data as fresh as possible while still removing redundant builds. Larger values always save more time, so recommending the largest qualifying value would push every model toward unnecessary staleness.

A model does not appear in the table when its current lag_tolerance is already equal to or greater than the recommended value, because there is nothing left to act on. Views are also excluded, since creating a view does not compute results and so has no build time to save. Each account shows the 20 models with the highest projected savings.
```

sidpatel13 (approved):

website/docs/docs/dbt-versions/release-notes.md
```diff
@@ -22,6 +22,7 @@ For <Constant name="fusion_engine" /> updates, refer to the [v2 changelog](https
 
 ## September 2026
 
+- **New:** The **dbt State** page now includes a [**Lag tolerance recommendations**](/docs/deploy/dbt-state-interface#lag-tolerance-recommendations) section that identifies models that could safely tolerate more lag. For each model, it shows the current lag tolerance, the recommended value, the estimated percentage of build time you'd save, and projected compute savings over the next 30 days. You can search by model name or filter by project, and apply recommendations by updating the [`lag_tolerance`](/reference/resource-configs/lag-tolerance) config.
```
nataliefiann: I was looking at [Sid’s comment](https://github.com/dbt-labs/docs.getdbt.com/pull/9973#discussion_r3972849652) on the **Projected 30d time savings** column, and he framed this as estimated model build time (this model only) rather than compute savings. Should this release note use the same wording?

website/docs/docs/deploy/dbt-state-interface.md
```diff
@@ -33,6 +33,41 @@ The **State** tab also displays the following charts:
 - **DATT** &mdash; Shows the target tables processed by dbt State, split into **Billable** and **Free**. Daily active target tables (DATTs) are the [billable units](/docs/platform/billing/dbt-state-usage#daily-active-target-tables) for dbt State. During a trial, all DATTs are counted as free.
 - **Asset builds** &mdash; Shows all model builds for the month, including models reused and cloned.
 
+## Lag tolerance recommendations
+
+The **dbt State** page includes a **Lag tolerance recommendations** section that identifies models that could safely tolerate more lag, letting dbt State skip more runs and save additional compute.
```
nataliefiann: Hiya Bianca

There’s no click path for the dbt State page like there is for the line above that describes the State tab. So a user might not know where to access this.

website/docs/docs/deploy/dbt-state-interface.md
```diff
@@ -33,6 +33,41 @@ The **State** tab also displays the following charts:
...
+
+The recommendations table displays the following columns:
+
+<SimpleTable>
+| Column | Description |
+|--------|-------------|
+| **Model name** | The name of the model that could benefit from a higher `lag_tolerance` value. |
+| **Project** | The dbt project the model belongs to, as set by `name:` in `dbt_project.yml`. This may differ from the project name in the <Constant name="dbt_platform" />. |
+| **Current lag** | The model's current `lag_tolerance` setting. |
+| **Recommended lag** | The `lag_tolerance` value dbt State recommends based on observed upstream data refresh patterns. |
```
nataliefiann: “based on observed upstream data refresh patterns" doesn’t match How dbt State calculates recommendations (further down), which describes 30-day build history and builds where definition and inputs hadn’t changed.

website/snippets/_lag-tolerance-recommendations-tip.md
```diff
@@ -0,0 +1 @@
+To help you tune `lag_tolerance` values, the **dbt State** page on the <Constant name="dbt_platform" /> provides [lag tolerance recommendations](/docs/deploy/dbt-state-interface#lag-tolerance-recommendations) based on observed upstream data refresh patterns, so you can see which models would benefit from a higher tolerance.
```
nataliefiann: Same comment as [here](https://github.com/dbt-labs/docs.getdbt.com/pull/9973/changes#r3987998383)

nataliefiann: Also (this is a non-block / take it or leave it) This file is named `_lag-tolerance-recommendations-tip.md`, but it isn’t a `:::tip` callout like other `*-tip` snippets (for example `_data-freshness-sla-tip.md`). It might be worth wrapping it in a tip callout for consistency, or renaming if we want to keep it as plain body text?

website/docs/reference/resource-configs/lag-tolerance.md
```diff
@@ -86,6 +86,10 @@ This config accepts two value types:
   lag_tolerance: "{{ '4h' if target.name == 'prod' else '7d' }}"
   ```
 
+import LagToleranceRecommendationsTip from '/snippets/_lag-tolerance-recommendations-tip.md';
```
nataliefiann: This tip currently sits in the middle of **Definition**, between the Jinja value-type example and **When does `lag_tolerance` apply**. That interrupts the config explanation. Could we move it to after Definition (or wrap it in a `:::tip`) so that section stays continuous?

website/docs/docs/deploy/dbt-state-interface.md
```diff
@@ -33,6 +33,41 @@ The **State** tab also displays the following charts:
...
+
+<SimpleTable>
+| Column | Description |
+|--------|-------------|
+| **Model name** | The name of the model that could benefit from a higher `lag_tolerance` value. |
+| **Project** | The dbt project the model belongs to, as set by `name:` in `dbt_project.yml`. This may differ from the project name in the <Constant name="dbt_platform" />. |
+| **Current lag** | The model's current `lag_tolerance` setting. |
+| **Recommended lag** | The `lag_tolerance` value dbt State recommends based on your model's 30-day build history. |
+| **% time saved** | The estimated percentage of build time you'd save by applying the recommended `lag_tolerance`. |
+| **Projected 30d time savings** | The estimated build time you could save over the next 30 days by applying the recommended `lag_tolerance`, based on redundant builds in the previous 30 days. This estimate includes only this model, so actual savings may be higher if downstream models also do not rebuild.|
```
nataliefiann:
```suggestion
| **Projected 30d time savings** | The estimated build time you could save over the next 30 days by applying the recommended `lag_tolerance`, based on redundant builds in the previous 30 days. This estimate includes only this model, so actual savings may be higher if downstream models also do not rebuild. |
```

nataliefiann (approved):

PR #10024, opened 2026-09-17
Fix last updated dates by stamping them from the GitHub API at build

mirnawong1:
## Problem
Doc pages showed the wrong "last updated on" date — most resolved to a single recent date instead of each page's real last edit.

## Root cause
Docusaurus (`showLastUpdateTime: true`) derives dates from git history. Vercel builds with a **shallow clone (~20 commits)**, so any page older than that boundary resolves to the boundary date. Confirmed in-build: the clone reports `isShallow: true`, and `git fetch --unshallow` is a no-op — Vercel won't serve deeper history to the build. So the date can't be recovered from git at build time.

## Fix
A `prebuild` script fetches each doc's real last-commit date from the **GitHub GraphQL API** (batched, ~25 calls for all docs) and writes it into the native `last_update` frontmatter field. Docusaurus reads frontmatter over git, so both the footer date and the structured-data date become correct with no component changes.

- Edits happen in the build sandbox only — nothing is committed.
- Respects any manually set `last_update`.
- Safe: no token or API error just falls back to git (current behavior); the build never fails on it.

## Required setup
Add a `GITHUB_TOKEN` env var to the Vercel project (Production + Preview). A read-only token is enough (fine-grained: Contents → Read on this repo, or a classic token with `public_repo`). Until it's set, dates fall back to git.

## How to verify
On the preview, open a page last edited long ago (e.g. `/reference/database-permissions/databricks-permissions`) — it should show its real 2023 date, not a recent one.


breezyfasano (approved):
this lgtm! :shipit: Also verified the Report Code functionality works. Very great feature! Nice!

mirnawong1:
awesome thank you so much @breezyfasano !

